PR #69, opened 2017-09-11
Refactor registration

DanNixon:
Removes the assumption about `mantidimaging` being the first entry on the Python path and located in the users home directory.

Replaces the directory walk with a package walk.


mantidimaging-ipython.sh
```diff
@@ -1,11 +1,5 @@
...
-# this will be used in the env script
-python_exec="ipython"
 
-# Call using . syntax, this will change this script's environment,
-# i.e. all the variables defined within the -env script will be visible here
-. $mantidimaging_dir/mantidimaging-env-py27.sh
+mantidimaging_dir=${HOME}/mantidimaging
 
-PYTHONPATH="$mantidimaging_dir:$PYTHONPATH" ${python_location} -i $mantidimaging_dir/mantidimaging/ipython.py "$@"
+${MANTIDIMAGING_PYTHON_BIN}/python $mantidimaging_dir/mantidimaging/ipython.py "$@"
```
martyngigg: Is it worth removing the `MANTIDIMAGING_PYTHON_BIN` variable and just assuming the user has set the path to the correct python executable?

DanNixon: I think so yes. With that assumption the scripts could just be generated by setuptools and these scripts removed then.

martyngigg: That sounds good to me. Would you prefer to do that separately from this?

DanNixon: I think that would move things along quicker.

In the current state of this PR I could get the CI up and running at least and this should only be a small addition.

martyngigg: :+1:

PR #70, opened 2017-09-11
Replace readme generation with Python logging

### LINKED ISSUE #68: Use Python logging framework
A lot of `readme.py` and `helper.py` seems to be devoted to writing a log to a file and console.

A good amount of this could be replaced with a reasonable logging framework (Python's own [`logging`](https://docs.python.org/3/library/logging.html#module-logging) for instance).

DanNixon:
Removes code dedicated to writing log/readme files.

This also has the benefit of being timestamped and the file appended to regularly (as opposed to at the end of execution).

Fixes #68


mantidimaging/core/imopr/cor.py
```diff
@@ -10,12 +12,13 @@ def sanity_checks(config):
...
-    helper.print_start(
-        "Running IMOPR with action COR. This works ONLY with sinograms")
+    log = getLogger(__name__)
+
+    log.info("Running IMOPR with action COR. This works ONLY with sinograms")
 
     tool = importer.timed_import(config)
 
-    print("Calculating projection angles..")
+    log.ingo("Calculating projection angles..")
```
martyngigg: Is this a typo? Should it be `log.info`?

PR #71, opened 2017-09-11
Fix GUI tests on Python 3.5

DanNixon:
On Python 3.5 `mock.create_autospec` was causing the stack visualiser presenter test to error, this appeared to be related to it being derived from a PyQt class.

PR #73, opened 2017-09-11
Change image index via mouse scroll

### LINKED ISSUE #42: Changing the image index with mouse scroll while viewing a volume
This should be handled as an event in the Stack visualiser view!

DanNixon:
Allows the index of the image displayed in the stack visualiser to be changed via mouse scroll when the cursor is over the canvas area (figure and slider).

Requires #71 to be merged first.

Fixes #42

### DIFF
--- a/mantidimaging/gui/stack_visualiser/sv_view.py
+++ b/mantidimaging/gui/stack_visualiser/sv_view.py
@@ -104,9 +104,9 @@ class StackVisualiserView(Qt.QMainWindow):
         :param event: Mouse scroll wheel event
         """
         if event.button == 'up':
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/stack_visualiser/sv_view.py
```diff
@@ -92,6 +93,23 @@ def current_roi(self):
     def current_roi(self, value):
         self._current_roi = value
 
+    def handle_canvas_scroll_wheel(self, event):
```
martyngigg: Should this be going through the presenter so that we could add a test?

PR #76, opened 2017-09-12
Output and load COR files

### LINKED ISSUE #64: Make Centers of Rotation readable from an input file
The ideal way to do this would be to add an additional command line flag in the functional config that takes a filepath as a parameter. The file will then be read and the data inside will be used to specify the sinogram IDs and their respective CORs.

The format is left to the developer's choice. Ideally it should be simple to parse and simple to write by hand

DanNixon:
Allows output and loading back of centres of rotation to/from files. The files in question are simple space separated text files containing pairs of slice index and COR.

The file is generated when an output directory is specified when running automatic COR finding (i.e. `--imopr cor`) and is loaded back by specifying a filename to the `--cor` parameter when performing reconstruction.

Fixes #64

### DIFF
--- a/mantidimaging/core/configs/recon_config.py
+++ b/mantidimaging/core/configs/recon_config.py
@@ -4,7 +4,9 @@ import argparse
 import os
 import sys
 import tempfile
+
 from argparse import RawTextHelpFormatter
+from logging import getLogger
 
 import numpy as np
 
@@ -73,6 +75,8 @@ class ReconstructionConfig(object):
             self.handle_special_arguments()
 
     def handle_special_arguments(self):
+        log = getLogger(__name__)
+
         if self.args.region_of_interest:
             if len(self.args.region_of_interest) != 4:
                 raise ValueError(
@@ -133,9 +137,8 @@ class ReconstructionConfig(object):
                     "Slice Indices (len {1})!".format(len_cors,
                                                       len_cor_slices))
 
-        # TODO
-        print(self.func.cors)
-        print(self.func.cor_slices)
+        log.debug("CORs: {}".format(self.func.cors))
+        log.debug("COR slices: {}".format(self.func.cor_slices))
 
         # if the reconstruction is ran on already cropped images, then no ROI
         # should be provided, however if we have a ROI then the Centers of Rotation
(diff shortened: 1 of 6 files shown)


mantidimaging/core/configs/recon_config.py
```diff
@@ -123,6 +133,10 @@ def handle_special_arguments(self):
                     "Slice Indices (len {1})!".format(len_cors,
                                                       len_cor_slices))
 
+        # TODO
+        print(self.func.cors)
```
martyngigg: Should these print statements be left here?

DanNixon: Nope, that is a mistake.

DanNixon:
Does that error also happen on `master` if you specify the CORs manually? 

If so I'll open a new issue for it as it looks unrelated to the changes here.

martyngigg:
I get a different error on `master`:

```
2017-09-25 09:31:59,352 DEBUG Logging initialised
  File "/home/dmn58364/miniconda3/envs/py27/bin/mantidimaging", line 11, in <module>
    load_entry_point('mantidimaging', 'console_scripts', 'mantidimaging')()
  File "/media/data1/source/github/mantidproject/mantidimaging/mantidimaging/main.py", line 72, in cli_main
    main(None)
  File "/media/data1/source/github/mantidproject/mantidimaging/mantidimaging/main.py", line 25, in main
    config = recon_config.grab_full_config(default_args)
  File "/media/data1/source/github/mantidproject/mantidimaging/mantidimaging/core/configs/recon_config.py", line 48, in grab_full_config
    return ReconstructionConfig(functional_args, args)
  File "/media/data1/source/github/mantidproject/mantidimaging/mantidimaging/core/configs/recon_config.py", line 73, in __init__
    self.handle_special_arguments()
  File "/media/data1/source/github/mantidproject/mantidimaging/mantidimaging/core/configs/recon_config.py", line 111, in handle_special_arguments
    self.func.cors = [float(cor) for cor in self.func.cors]
ValueError: could not convert string to float: /tmp/imaging/cors.txt
```

I guess that's because it doesn't understand the `cors` file though?

DanNixon:
Yes, I meant specifying the value of the COR using `--cors`.

DanNixon:
I've opened #100 to address the output path issue.

PR #78, opened 2017-09-13
Search groups of similar file extensions

### LINKED ISSUE #20: Easier format specifying
Currently what is specified in the `--in-format` command is _exactly_ what the loader looks for. This could be tedious to deal with:

- The images might be saved out as `.tif`, but if you pass `--in-format tiff`, _no_ images will be found, because the glob patterns looks for exact match for `tiff`.
- The same is valid for `fit` and `fits` images.

I don't believe glob has an `OR` statement. Adding a pattern match with a wildcard after to form `fit` or `tiff` might pick up wrong files, if saved out as `image.tiff_bak` or `image.tiff.bak`

Doing a double look up for the files (one with `.tif`, and if nothing found with `.tiff`) could be an easy solution, the hard part is knowing when to associate `tif` with `tiff`, `fit` with `fits`, and so on, without having to manually type it out.

DanNixon:
Searches for files with an ordered list of file extensions based on groups of similar/equivalent extensions (e.g. `.tif` and '.tiff').

If no files are found for the extension that the user specifies the loader will look for files with equivalent extensions.

Fixes #20

### DIFF
--- a/mantidimaging/core/io/utility.py
+++ b/mantidimaging/core/io/utility.py
@@ -1,10 +1,16 @@
 from __future__ import absolute_import, division, print_function
 
 import glob
+import itertools
 import os
 import re
 
 DEFAULT_IO_FILE_FORMAT = 'tif'
+
+SIMILAR_FILE_EXTENSIONS = (
+        ('tif', 'tiff'),
+        ('fit', 'fits')
+    )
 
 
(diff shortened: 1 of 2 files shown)


DanNixon:
@DanNixon do you want to request a review on this PR?

PR #79, opened 2017-09-13
Stack visualiser context menu

### LINKED ISSUE #58: Add right click on stack to show menu with options for histogram, contrast adjustment, etc
Add right click on stack to show menu with options for histogram, contrast adjustment, and any other features that might be added in the future

DanNixon:
Adds a context menu to the canvas on the stack visualiser.

Currently this only contains option to clear active ROI and to show the histogram but can easily be added to in the future.

Fixes #58

### DIFF
--- a/mantidimaging/gui/stack_visualiser/sv_presenter.py
+++ b/mantidimaging/gui/stack_visualiser/sv_presenter.py
@@ -60,6 +60,12 @@ class StackVisualiserPresenter(object):
         """
         self.view.show_histogram_of_current_image(new_window=True)
 
+    def do_clear_roi(self):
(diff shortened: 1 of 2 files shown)


DanNixon:
@martyngigg That would be a nicer option, however the way plotting is handled at the moment wouldn't allow for this as there is no way to know what plot windows are already open.

Eventually I'm planning on treating line plots in the same way as image plots (i.e. dockable in the main window), but this is not a priority at the moment.

martyngigg:
Fair enough. The addition of the context menu is certainly an improvement so we can tidy the options up in the future.

PR #81, opened 2017-09-14
Generate startup scripts with setuptools

### LINKED ISSUE #35: Auto generate imgpy.sh/.bat and imgpy-gui.sh/.bat with setup.py
This will be a necessary installation step. Currently the path is hardcoded to work locally. 

The desired behaviour can be achieved by setting an environment variable. However this could be difficult to enforce and maybe error prone.

I think a better solution would be for the `setup.py` generating script to get the full path to the installation directory and generate the `imgpy.sh` and `imgpy-gui.sh` with the _full path_ hard coded in them.

DanNixon:
Assumes that the target Python distribution is placed at the higher priority in the `PATH`.

Can use `pip install -e .` to setup locally for development.

Once installed, scripts can then be launched using `mantidimaging`, `mantidimaging-gui` and `mantidimaging-ipython`.

Fixes #35

### DIFF
--- a/mantidimaging-gui.sh
+++ b/mantidimaging-gui.sh
@@ -1,3 +1,0 @@
-#!/usr/bin/env bash
-
-${MANTIDIMAGING_PYTHON_BIN}/python -m "mantidimaging.main" --gui -i . "$@"
--- a/mantidimaging-ipython.sh
+++ b/mantidimaging-ipython.sh
@@ -1,5 +1,0 @@
-#!/usr/bin/env bash
-
-mantidimaging_dir=${HOME}/mantidimaging
-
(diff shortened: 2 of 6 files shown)


martyngigg:
@DanNixon There's a merge conflict here now too.

DanNixon:
One thing to note that just tripped me up. You need to run `pip install -e .` from the virtualenv only, for some reason after I had done it in the "main" environment the installation in the virtualenv was not used.

On a related note I'll go through and update the documentation after this is merged.

martyngigg:
I think by default the virtualenv doesn't have a link to the system does it?

Good plan on the documentation.

DanNixon:
> I think by default the virtualenv doesn't have a link to the system does it?

I thought so too, but `which mantidimaging-gui` gave me a path that was in the `bin` directory of Anaconda rather than the virtualenv.

PR #83, opened 2017-09-14
Add option to restrict file finding by filename prefix

### LINKED ISSUE #19: Prefix file loading
Sometimes the images might be in the same directory, there should be a prefix option to specify the prefix for the sample, dark and flat images.

- This should be part of the `core.imgdata.loader.load` function. 
- The `core.imgdata.loader.load_data` function also needs to be changed to read the prefixes from the config.
- The prefix parameters should be added in the `functional_config.py`.

DanNixon:
Provides an additional option (`--in-prefix`) that allows the a filename prefix to be specified when searching for input files.

For example `--in-prefix s1` would match the first two files in this list but not the last two:
- `s1_1.tiff`
- `s1_2.tiff`
- `s2_1.tiff`
- `s2_2.tiff`

Fixes #19

### DIFF
--- a/mantidimaging/core/configs/functional_config.py
+++ b/mantidimaging/core/configs/functional_config.py
@@ -27,6 +27,7 @@ class FunctionalConfig(object):
         self.input_path = None
         self.input_path_flat = None
         self.input_path_dark = None
+        self.in_prefix = ''
         self.in_format = DEFAULT_IO_FILE_FORMAT
         self.construct_sinograms = False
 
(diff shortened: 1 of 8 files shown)


martyngigg:
@DanNixon What's the best way to test this?

martyngigg:
@DanNixon More conflicts from formatting changes I think.

DanNixon:
I've added some more logging to the file finding step.

The easiest way would be to create a directory with several image files with the filenames listed in the description (they need not be valid TIFF images) then use this script to show the results of the file finder: `mantidimaging -i ~/test/ --in-prefix s1 --in-format tiff -o ~/test2/ --verbosity 10` (where `~/test` is the directory with input files).

martyngigg:
Thanks. Looks good.

PR #72, opened 2017-09-11
GUI error messagebox

DanNixon:
Adds the logic to show error dialog boxes on the GUI.

Requires #71 to be merged first.

PR #88, opened 2017-09-19
Handle loading errors in the GUI

### LINKED ISSUE #87: Loading errors crash the GUI
See https://github.com/mantidproject/mantidimaging/pull/72#issuecomment-330256929

DanNixon:
Fixes #87

Adds error handling around the data loading stages in the load dialog (where files are inspected for dimensions) and main window presenter (where the data is actually loaded).

There will trigger the error message box (see #72).

### DIFF
--- a/mantidimaging/core/algorithms/size_calculator.py
+++ b/mantidimaging/core/algorithms/size_calculator.py
@@ -77,4 +77,4 @@ def full_size_MB(shape=None, axis=None, dtype=None):
 
 
 def number_of_images_from_indices(start, end, step):
-    return int((end - start) / step)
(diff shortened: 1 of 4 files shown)


DanNixon:
Tests are added for data loading (once "OK" is clicked on the load dialog).

martyngigg:
Sorry I meant a unit test that the presenter does the right thing for bad input.

DanNixon:
That is what I've added (https://github.com/mantidproject/mantidimaging/pull/88/files#diff-92e1808852f2aae37fd9412c6a8abe4dR31), maybe I worded that badly.

PR #90, opened 2017-09-19
Fix failing tests on Python 2

### LINKED ISSUE #85: Address failing test on Python 2.7
See: http://builds.mantidproject.org/view/Imaging/job/mantidimaging-master/7/execution/node/23/log/

DanNixon:
- Python < 3.3 does not have the `inspect` module, so fall back to `funcsigs`
- Fix unicode string handling with StringIO
- Fix finding package name on Python 2

Fixes #85

### DIFF
--- a/mantidimaging/core/process_list/process_list.py
+++ b/mantidimaging/core/process_list/process_list.py
@@ -1,11 +1,12 @@
 from __future__ import absolute_import, division, print_function
 
-from io import StringIO
-import inspect
 import os
+import sys
 # using pickle instead of dill, because dill is not available on SCARF
 import pickle
+
 from ast import literal_eval
+from six import StringIO
 
 DEFAULT_ARGUMENT_SEPARATOR = ' '
 DEFAULT_FUNCTION_SEPARATOR = ';'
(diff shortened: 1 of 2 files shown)


martyngigg:
I think we should start using `six` here for Python 3 compatible things, e.g https://pythonhosted.org/six/#six.StringIO. It's the used by many other projects to simplify these things.

mantidimaging/core/process_list/process_list.py
```diff
@@ -120,18 +121,23 @@ def store(self, func, *args, **kwargs):
...
         Note: More arguments can be appended when the funciton is called for execution.
 
-        :param func: This works with a function reference, because metadata needs to be read from the funciton.
+        :param func: This works with a function reference, because metadata needs to be read from the function.
         :param args: Arguments that will be forwarded to the function. They must be literal values.
         :param kwargs: Keyword arguments that will be forwarded to the function. They must be literal values.
         """
         self._store_func(func, args, kwargs)
 
     def _store_func(self, func, args, kwargs):
```
martyngigg: I know this is not related to these changes but would it be better to define a type that captures the function/arguments into a single object that is callable rather than having to store the stringified version of the function + signature?

DanNixon: I think the advantage of the stringified version was the ability to read and edit them as plain text. That being said I'm not adversed to looking at other ways of doing this.

martyngigg: Fair point - I don't know enough about the requirements to say for sure.

martyngigg:
@DanNixon Now these are fixed should we look into adding the checks to the pull requests?

PR #94, opened 2017-09-20
UI files tidy

### LINKED ISSUE #93: Top level layouts on windows
Some of the `.ui` files do not have a top level layout and in some cases (e.g. load dialog) are completely statically positioned.

DanNixon:
- Ensures a top level layout is used on all `.ui` files
- Fixes a crash when a filter is run but no stack is loaded
- Address flake8 warnings in `algorithm_dialog.py`

To test, load an image and perform a background correction. Sample, dark and flat images are in the `notebooks/demoimages.tar.lzma` archive.

Fixes #93


martyngigg:
Note that I was able to cause a crash, not related to this PR, with the following steps:

* load a sample
* click filters -> Background Correction
* click ok without filling in the `flat` or `dark` fields

The GUI now hangs.

I don't know if this is a more general problem with how errors are handled during filter execution?

PR #96, opened 2017-09-21
Load dialog data type options

### LINKED ISSUE #91: Allow setting data type on GUI load dialog
Logic for handling data type in size calculation is already present (see `load_dialog.py:__init__()`), however this needs to be exposed to the GUI to be configurable by the user.

DanNixon:
Adds option to select pixel bit depth for size calculation when loading images.

Fixes #91


mantidimaging/core/filters/crop_coords/crop_coords_gui.py
```diff
@@ -5,6 +5,5 @@
 
 # def _gui_register(main_window):
```
martyngigg: Can we remove completely the commented out code?

DanNixon: That is done in #117

martyngigg: :+1:

PR #99, opened 2017-09-22
Tidy Python package init files

DanNixon:
Tidy and harmonise the syntax used in __init__.py files.


martyngigg:
@DanNixon This has just reminded me that maybe we should setup some static analysis on the server. At a minimum a `flake8` check would be good. We can use a [`.flake8`](mantidproject/mantid/blob/master/.flake8) file to control the config.

mantidimaging/core/convert/__init__.py
```diff
@@ -23,4 +23,7 @@
 # Code Documentation is available at: <http://doxygen.mantidproject.org>
 
 __all__ = ['convert']
+
 from . import convert  # noqa: F401, F403
+
+del absolute_import, division, print_function
```
martyngigg: What's the reason we need to delete this here?

mantidimaging/core/parallel/__init__.py
```diff
@@ -1,21 +1,31 @@
 from __future__ import absolute_import, division, print_function
 
+import sys
+
 __doc__ = """
```
martyngigg: Can we not just put this at the top of the module has normal?

mantidimaging/core/convert/__init__.py
DanNixon: It's done to prevent those symbols appearing when iPython auto-completion is used.

martyngigg: I see. I hadn't expanded the diff so I couldn't see them originally so I wondered how they had crept in.

mantidimaging/core/parallel/__init__.py
DanNixon: `__future__` imports are supposed to come at the start of the file.

martyngigg: There are a few things that are allowed to come before them though: https://docs.python.org/2/reference/simple_stmts.html#future.

We generally have them as the first thing after the docstring in all of our mantid code and I think it's cleaner. What do you think?

PR #111, opened 2017-10-03
Add method of populating filter UIs with basic details

DanNixon:
Removes a lot of duplicated code that handles UI in filters.

To review try using a selection of filters that cover all data types.

- Background Correction (sample, flat and dark images in `demoimages`
- Gaussian Filter (sample image in `demoimages`)
- Clip values (sample image in `demoimages`, min = 2500, max = 999999)

### DIFF
--- a/mantidimaging/core/filters/background_correction/background_correction_gui.py
+++ b/mantidimaging/core/filters/background_correction/background_correction_gui.py
@@ -3,13 +3,9 @@ from __future__ import absolute_import, division, print_function
 import os
 from functools import partial
 
-from PyQt5 import Qt
-
 from mantidimaging.core import io
-from mantidimaging.core.algorithms import gui_compile_ui as gcu
 from mantidimaging.core.algorithms import value_scaling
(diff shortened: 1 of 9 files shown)


martyngigg:
I don't think this is to do with these changes but I'm a bit confused at the interplay between the min/max and new min/new max boxes. Is the former supposed to specify the current values and the latter your new requirements?

DanNixon:
@martyngigg Do you mean in the clip values filter?

The new values are what the given pixel is set to if it is outside a bound.

For instance if min = 0.2 and new min = 0.05 and you have a pixel with value 0.1 then it will be replaced with 0.05.

martyngigg:
@DanNixon Ah, I see. Not necessarily as part of this but it feels like we could present these options in a better way. What do you think?

DanNixon:
@martyngigg Yes, the current names are not entirely obvious.

martyngigg:
@DanNixon Shall we deal with this in another issue?

DanNixon:
I'll take a look in this issue, another thing that would be nice is adding tool tips through this generic method.

PR #112, opened 2017-10-03
Update the stack list in algorithm dialogs

### LINKED ISSUE #98: Image stack list is not updated when loading new stacks
When you have a filter dialog open and a new stack is loaded, the input stack selector is not updated to show the newly loaded stack.

DanNixon:
Ensures that the stack list shown in algorithm dialogs is kept up to date when stacks are loaded or closed whilst algorithm/filter dialogs are open.

Fixes #98 

To test:
- Open Filters > Background Corrections
- Load and close several image stacks (sample, flat or dark images in `demoimages`)
- See that the stack selection drop down in the filter dialog remains up to date with the loaded stacks


martyngigg:
@DanNixon More conflicts in this one too.

Issue #107, opened 2017-10-02
Add GUIs for all filters


DanNixon:
Some filters do not have a GUI, add a minimal working GUI for each filter that is missing one.

Requires #111 to be merged first.

Modify data in place:
- [x] Cut off
- [x] Minus log
- [x] Rotate stack
- [x] Stripe removal

Do not modify data in place:
- [x] Crop coords
- [x] Rebin

PR #117, opened 2017-10-05
Filter GUIs - part 1

DanNixon:
Re #107 

Requires ~~#111~~ to be merged first.

- Adds a simple way of adding fields to a filter dialog
- Add GUIs for the following filters:
    - Cut off
    - Minus log
    - Rotate stack
    - Stripe removal
- Ensures that filter dialogs are created the correct size for the widgets they contain

To test:
  - Load a sample image (see `demoimages`)
  - Run some filters on it (all except stripe removal are pretty trivial)

### DIFF
--- a/mantidimaging/core/filters/cut_off/cut_off.py
+++ b/mantidimaging/core/filters/cut_off/cut_off.py
@@ -1,5 +1,7 @@
 from __future__ import (absolute_import, division, print_function)
+
 from mantidimaging import helper as h
+
 import numpy as np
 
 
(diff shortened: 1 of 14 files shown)


martyngigg:
@DanNixon We've got some merge conflicts here.

DanNixon:
@martyngigg I should have added a note mentioning this needs #111 to be merged first.

martyngigg:
@DanNixon I've merged #111 so we should be able to resolve these now.

martyngigg:
@DanNixon Is there any documentation anywhere about what each of the filters is supposed to do?

DanNixon:
Not at the moment.

Most of them are filters/operations the scientists are familiar with therefore the priority at the moment is making them work on the GUI (or at least be in a position for them to say they do not).

Documentation will be fully updated in time for it to be provided to users.

martyngigg:
@DanNixon Ok. I think we should try and do this sooner rather than later. From a reviewers point of view at this level we may want to check that the intention of a filter matches with the implementation.

martyngigg:
I was running the stripe removal filter with the default parameters and the log prints
```
Received parameter value ()
```
Is that an error or should the line be left out?

DanNixon:
@martyngigg That log message is fine, it is just informing that the slice visualiser provided no parameters to the algorithm (filter) dialog. Really it should probably be debug level.

Re documentation, I've made a start on this now.

martyngigg:
@DanNixon Thanks. We've got more merge conflicts to deal with but then this is good to go.

martyngigg:
Some merge conflicts again, @DanNixon. Let me know when they're fixed and I'll get this one through.

DanNixon:
@martyngigg Done

PR #129, opened 2017-10-10
Flake8 fixes in mantidimaging.imopr

DanNixon:
Code review and the builds still passing should be sufficient testing.

### DIFF
--- a/mantidimaging/core/imopr/visualiser.py
+++ b/mantidimaging/core/imopr/visualiser.py
@@ -2,24 +2,26 @@ from __future__ import (absolute_import, division, print_function)
 
 from logging import getLogger
 
+from mantidimaging.core.imopr.utility import handle_indices
+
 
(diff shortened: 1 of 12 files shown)


mantidimaging/core/imopr/visualiser.py
```diff
@@ -2,24 +2,29 @@
...
+
 
 def sanity_checks(config):
     pass
 
 
 def execute(sample, flat, dark, config, indices):
-    from mantidimaging.core.imopr import helper
     getLogger(__name__).info("Running IMOPR with action VISUALISE/SHOW")
-    import matplotlib.pyplot as plt
```
martyngigg: Is it possible this was intentionally imported only within the execution step as it's part of the `core`?

martyngigg:
@DanNixon More conflicts in this one too.

PR #131, opened 2017-10-10
Flake8 fixes in mantidimaging.tools

DanNixon:
Code review and the builds still passing should be sufficient testing.


DanNixon:
@DanNixon there are merge conflicts here, can you resolve them.

PR #136, opened 2017-10-10
Filter GUIs - part 2

### LINKED ISSUE #107: Add GUIs for all filters

DanNixon:
Fixes #107

- Allows filters that modify data by return value only to work in the GUI
- Fix an issue with data type conversion happening in rebin
- Add GUIs for crop coordinates and rebin filters

To test:
- Load an image
- Apply crop coordinates and rebin filters (rebin should be fairly self explanatory, the ROI selector can be used to get coordinates for crop coordinates)
- See that the stacks are resized in the GUI as expected

### DIFF
--- a/mantidimaging/core/filters/crop_coords/crop_coords_gui.py
+++ b/mantidimaging/core/filters/crop_coords/crop_coords_gui.py
@@ -1,9 +1,31 @@
-from mantidimaging.core.utility import gui_compile_ui as gcu
+from __future__ import absolute_import, division, print_function
+
+from functools import partial
+
 from mantidimaging.gui.algorithm_dialog import AlgorithmDialog
+
+from . import crop_coords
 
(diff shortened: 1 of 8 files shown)


martyngigg:
@DanNixon Sorry for leaving #139 for so long. There's a merge conflict to resolve now.

DanNixon:
Yes, there is a method of doing that currently which is used in the ROI normalisation filter. However as the current filter GUI is being replaced (and the implementation of that functionality having to change) I'm not sure using this here is worth the time.

PR #137, opened 2017-10-12
Common progress reporting interface

### LINKED ISSUE #106: Common method of progress reporting
A common way of reporting progress of a task is needed, this will likely replace most of what is left in `helper.py`.

This should provide time and progress estimations (based on an estimated number of steps and notifications after completion of a step, similar to `Progress` in Mantid) and logging of these.

It should be extended to include TQDM, the custom version of it included in the source and eventually the async task dialog for the UI.

DanNixon:
Fixes #106

Adds a progress reporting mechanism that is somewhat based on `Progress` in Mantid.

Anything that uses it has an optional `progress` parameter which may take a pre-configured instance (this is intended for the GUI when progress dialogs will be shown) or if left as the default will print an ASCII progress bar to standard out.

To test:
 - See that nothing is broken, pretty much all of the touched functionality is unit tested and they would have failed if the progress reporting was broken
- Load a large dataset (either on command line or GUI) and see  that the progress bar is shown and the reporting numbers look reasonable (i.e. correct number of images, bar progression looks right for progress). IMAT data from the last cycle should be good enough for this.


mantidimaging/core/utility/progress_reporting/progress.py
```diff
@@ -0,0 +1,184 @@
...
+
+        # Log initial memory usage
+        getLogger(__name__).debug(
+                "Memory usage before execution: %s",
+                get_memory_usage_linux_str())
+
+    def __str__(self):
+        return 'Progress(\n{})'.format('\n'.join(self.progress_history))
+
+    def __enter__(self):
```
martyngigg: Nice use of context management facilities :+1:

PR #150, opened 2017-10-20
New filter window

### LINKED ISSUE #138: New filter GUI
Open the filter dialog from a single menu option and select from a drop down of different filters. This is more in line with what other packages do.

EDIT: original issue was just as above, however the GUI could do with being rewritten to follow MVP style and to better contain its functionality (half of it's logic is in `sv_presenter.py`).

Needs done before #120 

To do:
- [x] Implement new window
- [x] Update filter registration
- [x] Unit tests
- [x] Parameter passing from SV

DanNixon:
Closes #138

Replaces the filter dialog with a new window that follows the MVP design pattern and contains the filter application logic (which was previously contained in the stack visualiser presenter).

To test:
- Load `demoimages/sample`
- Apply the background correction with the flat and dark images in `demoimages` (tests pre and post processing operations)
- Select a ROI containing the sample and apply the Crop Coordinates filter (tests auto parameters from stack)
- Try invalid input, missing data, etc. Stack traces should appear in the log but the UI should not crash.


DanNixon:
@martyngigg Not at the moment. It's something I'm not going to invest a lot of time in until it is asked for.

My priority is making sure bad input won't lose data (an exception in a notification to presenters used to propagate the exception to the point of exiting the UI).

PR #151, opened 2017-10-20
Unit test structure

DanNixon:
Restructure unit tests to keep the tests in a child `test` package to what they are testing.

PR #157, opened 2017-10-30
Add summed image mode to stack visualiser

### LINKED ISSUE #60: Add easier ROI selector
To select the correct region of interest, you have to check against projections from all projection angles. This can be made easier if we summed or averaged all of the images into a single one and showed that. 

The final image will contain the information from all the images from the stack and will clearly show the movement of the object through each projection angle.

DanNixon:
Contains #151 which should be merged first.

Closes #60

Allows stack of images to be summed when viewed on the stack visualiser.

This is important for selecting a ROI that covers the entire sample or contains only air.

To test:
- Load a dataset with at least two images (the sample images in `demoimages` are good as they have distinct features)
- Switch between stack and sum modes (via the context menu on the stack visualiser)


martyngigg:
This seems to work when switching between modes. In summed mode you can still step through the images - does that make sense? Was that requested?

DanNixon:
@martyngigg Added tests.

The scrolling was unintended, this is prevented when in summed mode now.

martyngigg:
@DanNixon Thanks

martyngigg:
GitHub is showing the diff incorrectly because of the rebase after merging #151. The compare view shows only the 4 files changed as expected.

PR #173, opened 2017-11-07
Abstract MVP classes

DanNixon:
Moves common MVP functionality to abstract classes (under `mantidimaging.gui.abstract_mvp`).

To test, open the following windows:
- main window
- filter window
- stack visualiser (i.e. load some data)


DanNixon:
@martyngigg Yep. No idea why I kept the "abstract" name in hindsight.

PR #175, opened 2017-11-07
Add Conda packaging

### LINKED ISSUE #164: Package using conda
Package for distribution via Conda.

This is probably going to be the easiest way to distribute the package going forward.

DanNixon:
Closes #164 

Adds the required files to build a Conda distribution of the `mantidimaging` package.

Can be built using the `buildscripts/conda/build.sh` script (once `conda-build` is installed) (the `--croot` option is to get around the fact `conda-build` does not like encrypted volumes).

If TomoPy is not installed first then nothing will work (see #176), and AFAIK TomoPy can not be automatically installed unless the correct channel is added to `.condarc` first (in which case the user may as well just install it).

To test:
- Create the packages using the `build.sh` script
- On a clean environment install TomoPy then the created package archive
- Run `mantidimaging-gui`, it should open and be usable

### DIFF
--- a/buildscripts/conda/build.sh
+++ b/buildscripts/conda/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+conda-build --croot /tmp/mantidimaging-conda mantidimaging
--- a/buildscripts/conda/mantidimaging/bld.bat
+++ b/buildscripts/conda/mantidimaging/bld.bat
@@ -1,0 +1,2 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
--- a/buildscripts/conda/mantidimaging/build.sh
+++ b/buildscripts/conda/mantidimaging/build.sh
@@ -1,0 +1,1 @@
+$PYTHON setup.py install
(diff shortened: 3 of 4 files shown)


buildscripts/conda/mantidimaging-gui/meta.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: mantidimaging
```
martyngigg: Should the name here be `mantidimaging-gui`?

DanNixon: I assumed not since it is still the same package. The only difference is that GUI related dependencies are included and the app entry point is defined. Both packages actually include the GUI code, but it would only work in this one.

I guess a better way to do this would be to have completely separate packages for the core and GUI.

DanNixon:
For now I'm just going to leave it as a single package with all the GUI dependencies.

Later on it can be split into two separate core and GUI packages.

martyngigg:
Agreed. This seems like a good starting point.

Issue #156, opened 2017-10-30
Remove CLI


DanNixon:
For discussion.

Currently the bloat around the CLI (i.e. `mantidimaging.core.configs`) makes the package a pain to use interactively (try loading data and finding the centre of rotation and see for yourself) and adds a significant maintenance overhead.

The CLI also offers little benefit over IPython (assuming a cleaner API without the mentioned bloat) due to it's complexity (if a command line tool needs documentation beyond `man` or `--help` something is wrong).

As an experienced user the API and IPython should be a viable replacement for the CLI assuming the bloat around configurations is removed. As a new user the GUI should provide the ability to perform the majority of required processing.

If required the IPython launch script could also import a series of helper scripts to make it more link a conventional CLI.

Removing the CLI will remove the need for the `mantidimaging.core.configs` and `mantidimaging.core.configurations` modules and greatly simplify parts of the codebase.

Things that would benefit from command line tools (e.g. process list execution) can go into small, specific commands that will live in a separate module (e.g. `mantidimaging.cli.[cmd]`).

PR #177, opened 2017-11-07
Guard filters that require TomoPy

### LINKED ISSUE #176: TomoPy is required to be installed
Ideally the TomoPy features should just not work when it is not installed, leaving the non-dependant functionality intact.

DanNixon:
Closes #176

For context the fact that a new module was added to deal with optional imports rather than just using the existing functionality in `tomopy_tool.py` is that a large refactoring of the `core.tools` package is planned when work starts on TomoPy integration (depending on how tools are handled in the GUI and the outcome of #156 tools may be handled differently to how they are now).

To test:
- Open the GUI in an environment with TomoPy installed, you should see all filters in the filter window
- Do the same in an environment that is missing TomoPy, you should see 5 filters missing

### DIFF
--- a/mantidimaging/core/utility/optional_imports.py
+++ b/mantidimaging/core/utility/optional_imports.py
@@ -1,0 +1,44 @@
+"""
+A place for availability checking and import logic for optional dependencies to
+live.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+import importlib
+import traceback
+import warnings
+
+from six import StringIO
+from logging import getLogger
+
+
+def safe_import(name):
+    """
+    Try to import an optional library that may not be present.
+
+    Will fail gracefully by returning None if the library is not available.
+    """
+    try:
+        module = importlib.import_module(name)
+
+    except ImportError:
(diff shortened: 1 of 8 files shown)


mantidimaging/core/utility/optional_imports.py
```diff
@@ -0,0 +1,34 @@
+"""
+A place for availability checking and import logic for optional dependencies to
+live.
+
+If something is not essential but enebles more functionality it should be in
```
martyngigg: Typo: *enebles*->*enables*

DanNixon:
@martyngigg The reason the traceback was used was to indicate what tried to include the missing module. But yes, it is not really an error.

In most cases the user is going to have TomoPy installed anyway since so much functionality is built off it at the moment.

martyngigg:
@DanNixon Ok. We could include the module name in the warning message?

DanNixon:
@martyngigg The only way to get what imported it is with a stack trace. This is now logged at DEBUG level as it is usually quite long and the warning emitted.

PR #178, opened 2017-11-08
Add generic stack selector widget

### LINKED ISSUE #171: Generic stack selector widget
Stack selection is going to be done in several places.

Add a common widget that handles keeping an up to date stack list.

DanNixon:
Closes #171 

To test:
- See that stack selection in the filter window is still working correctly
- In particular when the first stack is loaded and the last stack deleted


DanNixon:
That's taken care of in #179

PR #179, opened 2017-11-08
Address issues with filter window

### LINKED ISSUE #168: Remaining issues with filter window
- [x] Previews are not cleared after last data is removed
- [x] Changing filter properties does not trigger a preview refresh
- [x] Filter window remains open when main window is closed
- [x] Image zoom and pan is reset every time a preview is performed

DanNixon:
Closes #168

To test:
- Open the filter window and close the main window, there should be no windows open
- Load some data, open the filter window and close the data you just loaded, the previews should be empty
- Load some data and open the filter window, modify some filter parameters, the previews should update (intensity cut off makes an obvious change to the image)
- Zoom in on either image and trigger a preview update, the preview should show the same region of the image

### DIFF
--- a/mantidimaging/core/filters/background_correction/background_correction_gui.py
+++ b/mantidimaging/core/filters/background_correction/background_correction_gui.py
@@ -9,11 +9,16 @@ from mantidimaging.core.utility import value_scaling
 from . import execute
 
 
-def _gui_register(form):
+def _gui_register(form, on_change):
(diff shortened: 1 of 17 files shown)


martyngigg:
@DanNixon There are a some conflicts here now too.

martyngigg:
Sorry, @DanNixon. Now #183 has gone in it's got conflicts again

PR #185, opened 2017-11-14
Basic UI for finding COR and tilt

### LINKED ISSUE #127: COR/Tilt angle finding GUI
A GUI that will handle finding the centre of rotation and tilt angle.

DanNixon:
Fixes #127

Basic first implementation. Good enough for preliminary review by instrument scientists.

Issues still present:
- Processing is single threaded
- Processing is done in GUI thread

To test:
- Open GUI
- Load `babylon5/DanNixon/PSI_Cylinder/Sample` (entire angle space with a step size of 40 should do)
- Select a region of interest that encloses one of the two large cylinders (that contain the several thinner cylinders), exceed the cylinder on the X axis but remain within the cylinder on the Y axis
- Open the Find COR and Tilt UI and click Set ROI from stack, the image preview should show the ROI you selected on the stack visualiser
- Enter 25 slices
- Click OK, progress should be visible in the terminal output
- You should see the COR per slice plotted in blue with a linear fit in red and a yellow line showing the sample tilt over the image
- This sample is not very tilted but you can check on the large image on the stack visualiser that the reported centre of rotation is correct (this is measured from the bottom of the ROI you selected)


martyngigg:
Following the instructions I get the following error in the console when I click OK after setting `25` slices

cor_and_tilt_error

martyngigg:
@DanNixon I seem to get the same problem with Python 2.7

martyngigg:
This works now and seems to produce sensible numbers.

A few comments:

* None of the axes (actually anywhere) have units or labels on them - are there sensible defaults we can put there?
* In the COR "Find COR and Tilt" dialog I accidentally hit "OK" when I wanted to close the dialog as it felt a bit like a settings dialog where "OK" would close it. Should we separate the button that starts the finding and maybe move it above the results?

DanNixon:
@martyngigg Done.

Images have never had labels as the axes are just the pixel coordinates.

martyngigg:
@DanNixon Thanks.

Issue #119, opened 2017-10-06
TomoPy reconstruction GUI


DanNixon:
Add a GUI for TomoPy reconstruction.

PR #192, opened 2017-11-21
Add first cut of a TomoPy reconstruction UI

DanNixon:
Re #119 

To test:
- Load `babylon5/DanNixon/bg_corr_crop_sinogram`, all sinograms (step = 1)
- Open TomoPy Reconstruction
- Select a slice from the bottom cylinder
- Enter 590 in Rotation Centre (this is the COR relative to the cropped sinogram)
- Enter -0.003 in tilt angle
- Reconstruct a single slice, you should see a relatively OK reconstruction (most likely you'd have a better reconstructed slice if you selected a slice near the bottom of the cylinder)
- Reconstruct the entire volume (you'd need at least 16GB of RAM to do this)

There are a couple of UX issues with this so far:
- The COR is assumed to be at the bottom of the image, which here is not the case
- The COR is also assumed to be relative to the image given to the reconstruction UI, when it is measured relative to the entire image

(there are several "easy" solutions to these, choosing the best one will probably need more discussion)

### DIFF
--- a/mantidimaging/gui/ui/tomopy_recon_dialog.ui
+++ b/mantidimaging/gui/ui/tomopy_recon_dialog.ui
@@ -1,0 +1,192 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>Dialog</class>
+ <widget class="QDialog" name="Dialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>800</width>
+    <height>400</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Dialog</string>
+  </property>
+  <property name="sizeGripEnabled">
+   <bool>true</bool>
+  </property>
+  <layout class="QHBoxLayout" name="horizontalLayout">
+   <item>
+    <widget class="QSplitter" name="splitter">
+     <property name="orientation">
+      <enum>Qt::Horizontal</enum>
+     </property>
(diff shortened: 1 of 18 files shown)


mantidimaging/gui/ui/tomopy_recon_dialog.ui
```diff
@@ -0,0 +1,192 @@
...
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>800</width>
+    <height>400</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Dialog</string>
```
martyngigg: Minor point that the window title should probably say "Tomopy Reconstruction"?

martyngigg:
For the progress bar window, could the close button just minimise it to put a process bar in the status bar of the window or dock it in some way?

martyngigg:
One final thing on progress. I don't know how much control we have over this given tomopy is doing it but the progress bar really gives no indication of what has been done or what is left. It seems to jump to 50% and then 100% when it is finished.

martyngigg:
It also seems to take a long time to shutdown after the full volume reconstruction - is that expected.

martyngigg:
The second time I tried the full reconstruction (with python 2) the app hung completely when I tried to quit - I had to kill it.

DanNixon:
- The progress dialog will now remain open when the task is still running
- The delay on exit is likely the large memory deallocation (with a smaller dataset it is shorter)
- There does not seem to be a simple way to get fine grained progress from TomoPy

PR #213, opened 2017-12-13
Record filter params in metadata

DanNixon:
Saves a very rough history of filters that are applied in the GUI in the image properties.

To test:
- Load some data
- Apply some filters
- Save it
- Open the `.json` file saved alongside the images and see that the operations performed are listed

### DIFF
--- a/mantidimaging/core/data/images.py
+++ b/mantidimaging/core/data/images.py
@@ -66,6 +66,20 @@ class Images(object):
     def metadata_saves(self):
         return json.dumps(self.properties)
 
+    def record_parameters_in_metadata(self, func, *args, **kwargs):
+        if 'operation_history' not in self.properties:
+            self.properties['operation_history'] = []
+
+        def accepted_type(o):
+            return type(o) in [str, int, float, bool, tuple]
+
+        self.properties['operation_history'].append({
+            'name': func,
+            'args': [a if accepted_type(a) else None for a in args],
+            'kwargs': dict([(k, v if accepted_type(v) else None) for
+                           (k, v) in kwargs.items()])
+        })
+
     @staticmethod
     def check_data_stack(data, expected_dims=3, expected_class=np.ndarray):
         h.check_data_stack(data, expected_dims, expected_class)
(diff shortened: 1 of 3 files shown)


mantidimaging/core/data/images.py
```diff
@@ -66,6 +66,20 @@ def metadata_save(self, f):
...
+    def record_parameters_in_metadata(self, func, *args, **kwargs):
+        if 'operation_history' not in self.properties:
+            self.properties['operation_history'] = []
+
+        def accepted_type(o):
+            return type(o) in [str, int, float, bool, tuple]
+
+        self.properties['operation_history'].append({
+            'name': func,
+            'args': [a if accepted_type(a) else None for a in args],
```
martyngigg: Is it going to be a problem that this results in losing information?

mantidimaging/core/data/images.py
```diff
@@ -66,6 +66,20 @@ def metadata_save(self, f):
     def metadata_saves(self):
         return json.dumps(self.properties)
 
+    def record_parameters_in_metadata(self, func, *args, **kwargs):
+        if 'operation_history' not in self.properties:
+            self.properties['operation_history'] = []
+
+        def accepted_type(o):
+            return type(o) in [str, int, float, bool, tuple]
```
martyngigg: Is there a reason we accept `tuple` but not `list` or `ndarray`?

DanNixon: The ones listed are the ones that are definitely used (apart from `str`) that will not result in trying to save entire image stacks to a JSON file (which would happen if we accepted `ndarray`).

mantidimaging/core/data/images.py
martyngigg: I see. Is the idea really just to have a list of operations that can be viewed.

DanNixon: For now it is just to get the regions used when cropping the image so that the rotation centre parameter can be auto filled in the reconstruction dialog (the parameter must be relative to the cropped image, but the result given by the COR/Tilt dialog is relative to the entire image).

martyngigg: :+1:

PR #218, opened 2018-01-08
Reduce memory usage in COR/TIlt finding

### LINKED ISSUE #217: Reduce COR/Tilt finding memory consumption
The intended workflow using the COR/Tilt finding UI now uses far more memory than it needs to.

There are obvious ways to crop the sample down to just what is needed.

DanNixon:
Fixes #217.

Allows the dataset used for COR/Tilt finding to be downsampled in the GUI.

This allows a full dataset to be loaded, preprocessed, COR/Tilt finding run then saved as sinograms. This is now important as the full crop history and COR/Tilt results is expected to be found in the image metadata, which will only happen if the operations are carried out on a single dataset.

To test:
- Open your resource monitor of choice (`htop` sorted by memory usage will do)
- Open GUI
- Load `babylon5/DanNixon/PSI_Cylinder/Sample`, all images (step size = 1) (adjust this to a higher value if memory is limited on your system)
- At this point just over `16 / step size` GB should be consumed
- Select a region of interest that encloses one of the two large cylinders (that contain the several thinner cylinders), exceed the cylinder on the X axis but remain within the cylinder on the Y axis
- Open the Find COR and Tilt UI and click Set ROI from stack, the image preview should show the ROI you selected on the stack visualiser
- Enter 45 projections
- Enter 25 slices
- Click OK, memory usage should increase slightly here but definitely be less than `(16 / step size) + 2` GB
- You should see the COR per slice plotted in blue with a linear fit in red and a yellow line showing the sample tilt over the image

### DIFF
--- a/mantidimaging/core/reconstruct/cor_tilt.py
+++ b/mantidimaging/core/reconstruct/cor_tilt.py
@@ -35,13 +35,17 @@ def find_cor_at_slice(slice_idx, sample_data):
 
 
 def calculate_cor_and_tilt(
-        stack, roi, indices, cores=None,
+        stack, roi, indices, projections=None, cores=None,
         progress=None):
     """
     :param stack: Full image stack
     :param roi: Region of interest from which to calculate CORs
     :param indices: Indices of slices to calculate CORs (in full image
                     coordinates)
+    :param projections: Indices of projections to include in calculation, by
+                        default all are used.
+    :param cores: Number of CPU cores to execute over, by default use all
+                  available
     :param progress: Progress reporting instance
     """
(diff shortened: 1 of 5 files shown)


martyngigg:
Once I've run a COR/Tilt finding with 45 projections how can I go back to using the full number again? I tried deleting the number in the box but it put it back.

DanNixon:
@martyngigg At the moment you'd have to know the number of projections and enter it manually.

I'll add a "reset to default" button of some kind for this field.

martyngigg:
@DanNixon Good idea. I think that makes things a bit easier to use now.

PR #219, opened 2018-01-08
Remove in-source Tomopy code

DanNixon:
No longer needed as we use the latest version from Anaconda.


DanNixon:
@martyngigg Yes. I've pinned it to the version that I know works.

PR #221, opened 2018-01-11
Ensure shared array memory is freed

DanNixon:
Requires #260 to be merged first.

Fixes the bug where memory allocated via `sharedctypes.RawArray` was never being free'd until the program exits.

Also corrects a bug where the `Images` instance holding the allocated array would never get out of scope (at least for loaded images).

To test:
- Review alternative implementation
- Validate memory is being freed when images are closed
  - Load a large image stack (ideally one that just fits in memory twice)
  - Check mapped memory usage (`cat /proc/meminfo | grep Mapped`)
  - Close the image
  - Check mapped memory usage (`cat /proc/meminfo | grep Mapped`)
  - (it should be lower by approximately the size of the dataset)
- Try a few workflows (from recent PRs) and try to make something break
  - Following the steps of #213 and #242 is a good start

### DIFF
--- a/buildscripts/create_conda_environment.sh
+++ b/buildscripts/create_conda_environment.sh
@@ -14,7 +14,8 @@ then
   echo "Creating environment ${env_name}..."
   ${anaconda_dir}/bin/conda env create -n ${env_name} -f ${env_file}
 else
-  echo "Environment ${env_name} already exists"
+  echo "Environment ${env_name} already exists, eupdating..."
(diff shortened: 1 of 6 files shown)


martyngigg:
@DanNixon There's a conflict in the conda yaml file.

DanNixon:
79356704b1a3cb655c74d64d690bd1db35d6e01a should also get rid of the need to wipe the environment every time the spec changes.

Issue #224, opened 2018-01-17
Stack selector never disconnects signal/slot connection from main window


DanNixon:
```
Traceback (most recent call last):
  File "/home/dan/mantidimaging/mantidimaging/gui/widgets/stack_selector/view.py", line 31, in <lambda>
    lambda: self.presenter.notify(Notification.RELOAD_STACKS))
  File "/home/dan/mantidimaging/mantidimaging/gui/widgets/stack_selector/presenter.py", line 27, in notify
    self.show_error(e)
  File "/home/dan/mantidimaging/mantidimaging/gui/mvp_base/presenter.py", line 16, in show_error
    if hasattr(self.view, 'show_error_dialog'):
RuntimeError: wrapped C/C++ object of type StackSelectorWidgetView has been deleted
fish: “mantidimaging-gui” terminated by signal SIGABRT (Abort)
```

To reproduce:
- Open GUI
- Open a dialog with a stack selector (i.e. any of them apart from Load)
- Close it
- Load some data (generating a new stack, triggering a stack selector refresh)

PR #226, opened 2018-01-22
Show image metadata in GUI

DanNixon:
Adds a menu option to show image metadata in GUI.

When testing, be aware of: https://github.com/mantidproject/mantidimaging/issues/224

To test:
- Load some data
- Perform some basic filters on it
- Right click the data in the stack visualiser
- Click "Show Metadata"
- A dialog with image metadata should be shown

### DIFF
--- a/mantidimaging/gui/windows/stack_visualiser/metadata_dialog.py
+++ b/mantidimaging/gui/windows/stack_visualiser/metadata_dialog.py
@@ -1,0 +1,29 @@
+from __future__ import absolute_import, division, print_function
+
+from PyQt5 import Qt
+
+
+class MetadataDialog(Qt.QDialog):
+    """
+    Dialog used to show a pretty formatted version of the image metadata.
+    """
+
+    def __init__(self, parent, image):
+        super(MetadataDialog, self).__init__(parent)
+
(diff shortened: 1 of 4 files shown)


mantidimaging/gui/windows/stack_visualiser/metadata_dialog.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import absolute_import, division, print_function
+
+from PyQt5 import Qt
+
+
+class MetadataDialog(Qt.QDialog):
```
martyngigg: The pretty printing is good but when the dialog first opens for me I immediately have to resize it or the text wraps and messes up the pretty printing anyway. Should we try and set a slightly wider dialog when it first opens?

DanNixon: Widened to 600px by default.

I was hoping Qt could be able to dynamically resize it (as you can get the displayed document width of a `QTextEdit`), however the results were always too narrow.

martyngigg: I've always found Qt's sizing rules to be quite mysterious. I think the new default width is a good compromise. :+1:

PR #227, opened 2018-01-22
COR/Tilt finding refactor

DanNixon:
When testing, be aware of https://github.com/mantidproject/mantidimaging/issues/224#

The same instructions from https://github.com/mantidproject/mantidimaging/pull/218 can be used for testing.


DanNixon:
The issue seemed to be (for me at least) not properly initialising the slice index cache. This and projection index cache have been moved to be generated just before execution as they are not actually used before then anyway.

PR #228, opened 2018-01-23
Convert larger tools to windows

DanNixon:
Converts filters, COR/tilt and reconstruction to windows.

Fixes an issue with stock selector widgets not having connections to the main window removed.

To test:
- Open each window and make sure they still load
- Run a filter
- Open and close some of the mentioned windows
- Load and delete data (doing this after one of the mentioned windows has been closed demonstrates the mentioned bug is fixed)


martyngigg:
@DanNixon There are some conflicts now after the refactor.

martyngigg:
@DanNixon In Python 2, when I close the filters window I get:
```
  File "/media/data1/source/github/mantidproject/mantidimaging/mantidimaging/gui/mvp_base/view.py", line 24, in closeEvent
    super().closeEvent(e)
TypeError: super() takes at least 1 argument (0 given)
```
It's only Python 3 that lets you use super without any arguments.

DanNixon:
@martyngigg Fixed the `super()` issue (I'm too used to other projects being Python 3 only...)

I have no idea why the layout is tight like that, there should be padding on the layouts and they look fine in Qt designer. I half expected it to be an i3 thing but that likely isn't the case.

mantidimaging/gui/windows/main/view.py
```diff
@@ -69,13 +69,13 @@ def show_save_dialogue(self):
         self.save_dialogue.show()
 
     def show_cor_tilt_window(self):
-        CORTiltDialogView(self).show()
+        CORTiltWindowView(self).show()
 
     def show_filters_window(self):
-        FiltersDialogView(self).show()
+        FiltersWindowView(self).show()
```
martyngigg: Might there be a lifetime issue here now? I don't think `show` waits...

DanNixon: Yes, it is possible. However it's being masked by another issue with the way references are handled in the way the MVP pattern has been implemented.

It's something I plan to fix however I'd rather get the fix in this PR merged as users will almost certainly encounter this if they try and use it in the current state.

martyngigg: Okay. Let's merge this and figure out the weird layout issues later.

DanNixon:
I created #231 and #232 to keep track of the additional issues mentioned.

PR #229, opened 2018-02-06
COR/Tilt UI redesign

DanNixon:
Adds the ability to manually change the COR/Tilt finding results or perform finding by manual inspection of individual reconstructed slices.

To test:
- Load all images from `babylon5/DanNixon/sample_intensity_cutoff_crop/`
- Open *Reconstruct* > *Find COR and Tilt*
- (no cropping is needed as the dataset is already cropped to a correct region)
- Set 45 projections and 10 slices
- Click *Calculate*
- (*Results* tab should be automatically shown)
- Click *Clear All*
- Select somewhere near the top of the projection preview and click *Add*, enter a COR of around 610 (try to remove the halo around the small cylinders)
- Do the last step two or three more times at even intervals down the image
- Click *Fit*
- Click anywhere on the image and you should get a reasonable reconstruction (at least as reasonable as it was when you selected the CORs)
- Right click on the image and view its metadata
- Ensure the COR/Tilt data reflects the last entry in the *Results* tab

### DIFF
--- a/mantidimaging/core/cor_tilt/data_model.py
+++ b/mantidimaging/core/cor_tilt/data_model.py
@@ -37,32 +37,42 @@ class CorTiltDataModel(object):
         self._cached_c = None
 
     def populate_slice_indices(self, begin, end, count, cor=0.0):
-        self._points = [[idx, cor] for idx in np.linspace(
+        self.clear_results()
+
+        self._points = [[int(idx), cor] for idx in np.linspace(
             begin, end, count, dtype=int)]
         LOG.debug('Populated slice indices: {}'.format(self.slices))
 
     def linear_regression(self):
+        LOG.debug('Running linear regression with {} points'.format(
+            self.num_points))
         result = sp.stats.linregress(self.slices, self.cors)
         self._cached_m, self._cached_c = result[:2]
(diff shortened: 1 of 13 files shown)


DanNixon:
Yes, the left panel can't shrink anymore. I can't quite find which size policy is enforcing this though as I've tried to allow everything to shrink where possible.

What part of resizing the rest is difficult? Is it just the width and visibility of the resize handle?

martyngigg:
Yes it's mostly not too clear how to resize things but I guess that's more or a general Qt problem with those handles. The main thing would be the initial width of the central panel where it obviously cuts off the Y-axis labels. Would that be hard to fix?

DanNixon:
I can't find a simple fix for the Matplotlib issues, there are a couple of other annoyances that I haven't gotten to the bottom of yet to.

As this is image data I'm inclined to say it's low enough priority to leave for now as it has been mentioned a lot in meetings that the actual coordinates are pretty irrelevant compared to a visual indication of where something is (i.e. the yellow slice preview line is more useful than reading off the Y axis).

martyngigg:
Fair enough 👍

PR #236, opened 2018-02-09
Add the ability to open a large COR vs slice plot

### LINKED ISSUE #235: Open large COR vs slice index plot from COR/Tilt UI
Open a larger COR vs slice index plot when the user double clicks on the smaller plot in the *Results* tab of the COR/Tilt window.

DanNixon:
See commit messages for descriptions.

Fixes #235

To test:
- Load some data (anything will do here)
- Open *Find COR and Tilt* window
- Switch to *Results* tab
- Open the larger COR vs slice plot by double clicking on the mini plot at the top of the *Results* tab
  - When less than 2 points are added no plot is opened
  - When 2 or more points are added only the data line is shown
  - When 2 or more points are added and the fit has been run the data and fit lines are shown

### DIFF
--- a/mantidimaging/gui/windows/cor_tilt/presenter.py
+++ b/mantidimaging/gui/windows/cor_tilt/presenter.py
@@ -3,6 +3,8 @@ from __future__ import absolute_import, division, print_function
 from enum import Enum
 from logging import getLogger
 
+import matplotlib.pyplot as plt
+
 from mantidimaging.core.utility.progress_reporting import Progress
 from mantidimaging.gui.dialogs.async_task import AsyncTaskDialogView
 from mantidimaging.gui.dialogs.cor_inspection import CORInspectionDialogView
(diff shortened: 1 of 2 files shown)


mantidimaging/gui/windows/cor_tilt/presenter.py
```diff
@@ -118,6 +123,19 @@ def do_add_manual_cor_table_row(self):
...
+    def do_plot_cor_vs_slice_index(self):
+        if self.model.model.num_points > 1:
+            fig = plt.figure()
+            ax = fig.add_subplot(111)
+
+            ax.plot(self.model.model.slices, self.model.model.cors)
+
+            fit_data = self.model.preview_fit_y_data
+            if fit_data is not None:
+                ax.plot(self.model.model.slices, fit_data)
```
martyngigg: Is it worth adding a legend for when the fit curves are plotted? Or maybe consider a diff-style plot as well where the difference is plotted on a separate axis below?
Also, do the axes need labels? It seems odd not having them...

DanNixon: Yes, on the larger plot there is really no excuse for lack of axes labels and legend.

PR #239, opened 2018-02-13
Add dialog for manual rotation centre refinement

DanNixon:
Part of the work required for #237

Implements a GUI that allows manual rotation centre refinement by iteratively selecting the best reconstruction out of three possibilities.

To test, run this script (change path to data as required):
```
import logging
import sys

from PyQt5.Qt import QApplication

from mantidimaging.helper import initialise_logging
from mantidimaging.core.io.loader import load
from mantidimaging.gui.dialogs.cor_inspection import CORInspectionDialogView

initialise_logging(logging.DEBUG)

d = load('/home/dan/datasets/psi_cylinder/sample_intensity_cutoff_crop/', indices=(0, 940, 5))
print(d.sample.shape)

app = QApplication(sys.argv)

cid = CORInspectionDialogView(None, data=d.sample, slice_idx=5)
res = cid.exec()
print(res)
print('Found COR:', cid.optimal_rotation_centre)
```

### DIFF
--- a/mantidimaging/gui/dialogs/cor_inspection/view.py
+++ b/mantidimaging/gui/dialogs/cor_inspection/view.py
@@ -1,0 +1,96 @@
+from __future__ import absolute_import, division, print_function
+
+from matplotlib.backends.backend_qt5agg import FigureCanvasQTAgg
+from matplotlib.figure import Figure
+
+from mantidimaging.gui.mvp_base import BaseDialogView
+from mantidimaging.gui.utility import BlockQtSignals
+from mantidimaging.gui.widgets import NavigationToolbarSimple
+
+from .presenter import (
+        CORInspectionDialogPresenter,
+        Notification as PresNotification)
+
+
+class CORInspectionDialogView(BaseDialogView):
+    def __init__(self, parent, cmap='Greys_r', **args):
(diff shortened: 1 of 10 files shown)


mantidimaging/gui/dialogs/cor_inspection/view.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Completed button action
+        self.finishButton.clicked.connect(self.accept)
+
+        # Image canvas
+        self.image_figure = Figure(tight_layout=True)
+        self.image_canvas = FigureCanvasQTAgg(self.image_figure)
+        self.image_canvas.setParent(self)
+        self.imagePlotLayout.addWidget(self.image_canvas)
+        self.image_plots = self.image_figure.subplots(
```
martyngigg: I had `matplotlib` `2.02` installed and that doesn't have `Figure.subplots` - it looks like that was introduced in `2.1.0`. Should we have a minimum `matplotlib` version requirement?

DanNixon:
It is also used in the GUI, that is done in #242.

mantidimaging/gui/dialogs/cor_inspection/view.py
DanNixon: Done.

PR #254, opened 2018-02-20
Add sanity check for TomoPy version

### LINKED ISSUE #252: Check TomoPy version on startup
As older versions of TomoPy can lead to issues with workflows used here, there should be a check that an up to date version is being used.

At least 1.1.0 or higher should cover all the known issues.

```python
import tomopy
tomopy.__version__
```

DanNixon:
Adds a sanity check at startup time that the version of TomoPy installed matches that defined in environment.yml.

The check does nothing if TomoPy is not installed.

Fixes #252

To test:
- Create a blank miniconda environment
- Install mantidimaging (i.e. `python setup.py develop`)
- Try to start the GUI (i.e. `mantidimaging-gui`)
- (an exception should be thrown stating the TomoPy version is not correct and the GUI will not start)

### DIFF
--- a/mantidimaging/main.py
+++ b/mantidimaging/main.py
@@ -7,6 +7,7 @@ import warnings
 from mantidimaging import helper as h
 from mantidimaging.core.configs import recon_config
 from mantidimaging.core.utility import ExecutionTimer
+from mantidimaging.core.utility.optional_imports import safe_import
 
 formatwarning_orig = warnings.formatwarning
 warnings.formatwarning = \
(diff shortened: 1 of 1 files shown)


mantidimaging/main.py
```diff
@@ -7,16 +7,29 @@
...
 warnings.formatwarning = \
     lambda message, category, filename, lineno, line=None: \
     formatwarning_orig(message, category, filename, lineno, line='')
 
 
+def startup_checks():
+    tomopy = safe_import('tomopy')
+    if tomopy is not None:
+        ver = tomopy.__version__
+        if ver != '1.1.2':
```
martyngigg: Do we want to specify an exact version of a minimum? (`distutils.LooserVersion` is useful here).

DanNixon: We specify an exact version in the `environment.yml` and Conda package `meta.yml`, this is just mirroring the version that is pinned there.

martyngigg: Fair point 👍

PR #256, opened 2018-02-21
Remove Python 2.x special import logic

DanNixon:
Removes special import logic that was required for Python 2.x compatibility.

### DIFF
--- a/docs_misc/examples/cli_checks_performance.py
+++ b/docs_misc/examples/cli_checks_performance.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import, division, print_function
-
 from timeit import timeit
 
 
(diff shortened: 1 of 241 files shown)


martyngigg:
@DanNixon Is this still in progress?

DanNixon:
For now, yes.

I was going to have a look though to see if there was version handling anywhere else in the codebase. These were just the ones I knew about/came to mind instantly.

DanNixon:
I also wonder if it's worth removing the `future` imports yet?

I guess with them being at the top of files it makes it an easily revertable change if something came up and Python 2 was required again.

martyngigg:
Yep, maybe leave them just for now.

martyngigg:
Actually no I misunderstood your comment. Yes, let's remove them and they can be put back if necessary.

DanNixon:
@martyngigg This is done now.

PR #266, opened 2018-02-28
Clarify tool naming

DanNixon:
See commit messages for description.

To test:
- Open GUI and look at menu options
- See that the correct window still opens for each menu option


DanNixon:
test this please

PR #277, opened 2018-04-09
Documentation updates

DanNixon:
See commit messages.

To test:
- `python setup.py docs_api`
- `python setup.py docs`
- See that changed docs make sense


DanNixon:
If they are all from `*/api/*/test/*` then yes, they are fine. They are just the unit test modules not being included in the API documentation.

PR #311, opened 2019-07-29
Cache miniconda/pkgs and envs/test-environment

DTasev:


DTasev:
Not needed with environment from docker

PR #355, opened 2019-10-02
Auto start Savu backend

dtasev:
Automatically starts the docker back-end if it is not already running. If it is running, just connect to it.

Todo:
- [x] Create a config for the data mount/save directories
- [x] Create a docker image that has a copy of the hebi and savu source inside it, so that it does not have to be mounted on the client. (now at dtasev/hebi, for development use `dtasev/hebi:dev`)
- [ ] Consider: only start docker / connect Socket IO client when the Savu Filters GUI is opened. Fail gracefully if anything fails
- [x] Review socket io disconnect on close
- [ ] Refactor `preparation.py`
  - Maybe split into multiple files/another package
  - Move into `core.net`?

Probably better in a follow-up PR:

- [ ] Persistent config storage 
  - `QSetting` are a pain (from Mantid Workbench experience), maybe config with dataclasses loaded from JSON (should be simple as `**json.load(file)`)?
- [ ] GUI to edit config? Start of a settings window?


WHTaylor:
Work on savu features has been continued on the `savu_filters` branch, I think this PR is about as big as it should be.

PR #366, opened 2019-10-24
Adding a new centre of rotation row in the COR/Tilt GUI no longer shown an error

DTasev:
This PR fixes the error that was shown when the user clicked Add in the COR/tilt GUI. This function is available after the automatic COR calculation is performed.

To test:

- Load a stack
- Open cor/tilt GUI and run an automatic calculation
- On the results tab select a row and click add
- The new row should have the same slice index and COR
- Remove/clear all should still work

No related issue


WHTaylor:
Looks good, fixes error

PR #370, opened 2019-10-24
Option to set all CORs to match selected row

### LINKED ISSUE #294: Add "Set to All" button on Find COR and Tilt
This will make it easier to setup datasets that happen to have the same CoR on all indices.

Once a good Centre of Rotation has been found, clicking the set-all button will set it on all rows of the table. Doing a fit would then be a straight line.

WHTaylor:
Closes #294 

UI change:

Screenshot from 2019-10-24 16-25-05

This bit of the code, having multiple models etc. is very confusing, so there may be a simpler way of doing this.

### DIFF
--- a/mantidimaging/gui/windows/cor_tilt/presenter.py
+++ b/mantidimaging/gui/windows/cor_tilt/presenter.py
@@ -27,6 +27,7 @@ class Notification(Enum):
     ADD_NEW_COR_TABLE_ROW = 7
     REFINE_SELECTED_COR = 8
     SHOW_COR_VS_SLICE_PLOT = 9
+    SET_ALL_ROW_VALUES = 10
 
 
 class CORTiltWindowPresenter(BasePresenter):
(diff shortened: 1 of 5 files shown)


mantidimaging/gui/windows/cor_tilt/presenter.py
```diff
@@ -203,3 +206,11 @@ def _on_finding_done(self, task):
             msg = self.ERROR_STRING.format(task.error)
             log.error(msg)
             self.show_error(msg)
+
+    def set_all_row_values(self):
+        if self.view.point_model.empty:
+            return
+
+        selected_cor = self.model.cor_for_current_preview_slice
+        for slice_idx in self.model.model.slices:
```
DTasev: Could this be moved inside a function of `self.model` (to avoid `model.model`)?

WHTaylor:
I've gotten rid of the newly introduced one, but there are another dozen uses of `self.model.model`left in the presenter. If you think this class is likely to be changed much, I think it'd be worth refactoring it a bit, as it's quite hard to understand atm.

PR #376, opened 2019-10-28
Option to construct Images with existing properties

### LINKED ISSUE #321: Recon stack should inherit pre-processed history
Possibly after #352

Reconstructing a stack of sinograms currently produces a brand new stack with unknown history. 

This should copy over the `self.properties[const.OPERATION_HISTORY]` to the reconstructed stack before/after producing the reconstructed images. This ensures that if they are saved out the pre-processing parameters are not lost.

WHTaylor:
Closes #321. Relies on #373

### DIFF
--- a/mantidimaging/gui/windows/tomopy_recon/presenter.py
+++ b/mantidimaging/gui/windows/tomopy_recon/presenter.py
@@ -1,8 +1,9 @@
+from copy import deepcopy
 from enum import Enum
 from logging import getLogger
 from typing import TYPE_CHECKING, Dict, List
 
-from mantidimaging.core.data import Images
+from mantidimaging.core.data import const, Images
 from mantidimaging.core.reconstruct.utility import get_cor_tilt_from_images
 from mantidimaging.core.utility.progress_reporting import Progress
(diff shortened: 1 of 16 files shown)


mantidimaging/gui/windows/tomopy_recon/presenter.py
```diff
@@ -130,7 +132,7 @@ def _on_reconstruct_slice_done(self, task):
     def _on_reconstruct_volume_done(self, task):
         if task.was_successful():
             volume_data = task.result
-            volume_stack = Images(volume_data)
+            volume_stack = Images(volume_data, properties=self.stack_properties)
```
DTasev: I wonder if this could be a good PR to add recording the parameters of the reconstruction? Centres of rotation, projection angles (if provided), algorithm name and post-processing filter.

Do you think this can be done here or should it be moved to another PR?

WHTaylor: Yeah that'll be good to do - it seemed a bit weird to be ignoring it (and swapping axes) in the history.

PR #379, opened 2019-10-30
Pyqtgraph for filter previews

### LINKED ISSUE #330: Replace Matplotlib with PyQtGraph in Filters GUI
The current filters window `mantidimaging.gui.windows.filters` uses a matplotlib figure with 4 axes to construct the preview view on the side, 2 images and 2 plots. 

This should be replaced to use `pyqtgraph` plot and images. Investigate if it is possible to replicate matplotlib's axes using pyqtgraph (maybe it has a similar Figure/Axes structure?). If not it can likely be done with a Qt widget and a few layouts

WHTaylor:
Closes #330 

Currently using a class that is specific to the filter window, when looking at #331 and #332 will make something generic if it fits.

Haven't replaced the matplotlib toolbar, as that kind of functionality is accessible through right click menu in pyqtgraph instead.

Not if sure there's any equivalent for cmap.

### DIFF
--- a/mantidimaging/gui/ui/filters_window.ui
+++ b/mantidimaging/gui/ui/filters_window.ui
@@ -31,16 +31,7 @@
         <property name="sizeConstraint">
          <enum>QLayout::SetDefaultConstraint</enum>
         </property>
-        <property name="leftMargin">
-         <number>9</number>
-        </property>
-        <property name="topMargin">
-         <number>9</number>
-        </property>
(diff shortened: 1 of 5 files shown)


WHTaylor:
I think a separate PR might be best, leaving this one as a straight replacement of what existed before. It might end up worked into one of the other replacing matplotlib PRs, depending on how it ends up being generalised.

Issue #381, opened 2019-11-06
Rework process list functionality to work with class methods


WHTaylor:
Follow up to #327

The current process list functionality is based on plain functions being called. The above issue replaces those with class methods instead. `executor_test` has been disabled until this issue is dealt with.

PR #382, opened 2019-11-06
Replace _guiregister

### LINKED ISSUE #327: Remove _gui_register in favour of a base-class that forces the needed functions/params
Needed functions are:
- Register
- Execute
- Do Before - has to be present, doesn't need to do anything
- Do After - has to be present, doesn't need to do anything
- Parameters - what parameters are required from the stack visualiser, e.g. ROI

WHTaylor:
Closes #327 

Notable points of this way too big PR (sorry):
 - Consolidates filter functions into classes which inherit from BaseFilter.
 - Removes a lot of the closures that were in use for registering gui components. Instead, `register_gui` now returns a dictionary of all the dynamically added widgets which are passed to `execute_wrapper` as kwargs, which retrieves the needed values from them and creates a partial of `_filter_func` with those values.
 - A couple of new tests at the end of `filters/test/model_test` to check all filters loaded as expected and that metadata is fully recorded by `apply_filter`
 - Disabled 3 tests in `ExecutorTest` due to process lists being incompatible with this PR. #381 open to fix this.


mantidimaging/core/filters/rotate_stack/rotate_stack.py
```diff
@@ -1,20 +1,84 @@
...
+        _, rotation_count = add_property_to_form(
+            'Number of rotations', 'int', 1, (0, 99),
+            form=form, on_change=on_change)
+
+        return {
+            "rotation_count": rotation_count
+        }
+
+    def execute_wrapper(self, rotation_count=None):
+        return partial(self._filter_func, rotation_count=rotation_count.value())
```
dtasev: I get an error here as I think `rotation_count` is trying to be passed into `rotation`, which then doesn't match the kwargs

Issue #285, opened 2018-06-01
Image Operations UI


DanNixon:
- Some filters do not update previews
- No indication when filter is running
- Histogram should overplot the data for each image, and in the 4th window the difference of the images should be shown

WHTaylor:
All of these are now done

PR #383, opened 2019-11-07
Filter previews - difference image and histogram overplotting

### LINKED ISSUE #354: Show image of difference between pre/post processed image
Updating this issue to be a follow up discussed in https://github.com/mantidproject/mantidimaging/pull/379.

The PR made the filters windows have pyqtgraph widgets instead of matplotlib. The next step is to iterate on that and:
- Merge the two histograms into one
- Show the difference image in the lower-right square

WHTaylor:
Ref #285
Closes #354 

Features added to filter previews:
 - Option to overplot histograms
 - Optional legend on overplotted histogram
 - Difference image
 - Ability to link images together for scrolling, zooming etc

Slight problem at the moment is that images in rows without plots expand slightly to fill the space, making them a little larger than the others, which makes the linked scrolling a bit glitchy. Some kind of empty spacer widget would probably work to fix - if I can find something I'll add to this PR, but for now leaving as is.

filters

### DIFF
--- a/mantidimaging/core/utility/histogram.py
+++ b/mantidimaging/core/utility/histogram.py
@@ -1,9 +1,9 @@
 import numpy as np
 
-HISTOGRAM_BIN_SIZE = 2048
+DEFAULT_NUM_BINS = 2048
 
 
-def generate_histogram_from_image(image_data, bin_size=HISTOGRAM_BIN_SIZE):
-    histogram, bins = np.histogram(image_data.flatten(), bin_size)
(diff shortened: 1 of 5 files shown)


WHTaylor:
New layouts

combine

split

When histograms are combined the images scale don't have the same priority for expanding for some reason I can't track down, so will be different sizes if there isn't enough (horizontal) space. As long as there's enough space for all of them to fit the layout is fine.

PR #385, opened 2019-11-15
Misc refactoring

WHTaylor:
Converted functions to properties, a couple of new named tuples, additional typing, removed unused property

PR #390, opened 2019-11-21
Fix failing test

WHTaylor:
There's probably a nicer way to fix this with `mock.patch` but it's a bit tricky due to the filter class inheritance going on. This PR saves and restores the original state of the wrapper functions after each test, so that when the actual functions are required by a later test they are available.


WHTaylor:
Failed builds for #385 and #389 should be fixed by this PR.

PR #394, opened 2019-11-29
Savu docs

WHTaylor:


WHTaylor:
@dtasev I've written up (how I've come to understand) the way savu/hebi works in imaging at a high level, as it stands at the moment. Anything wrong, missed out, or in the process of being changed that should be mentioned?

A couple of thoughts I had:
 - It doesn't seem right for imaging to know how to build process lists. Might be over optimistic (probably a lot of work), but could it make sense to pull `savu_config` out of `savu` as it's own package, and use that?
 - If I read things correctly, imaging builds a process list which it saves as a file, then sends it as bytes to hebi, which saves it again, so savu can be pointed at it. At the least, would it make sense for imaging's version to be temporary, and deleted after sending?
 - The endpoint for submitting a job to HEBI should be a POST to /jobs/queue
 - It would probably be nice for hebi to abstract away more of the file stuff, assigning ids of some kind and using them instead of file paths in query strings everywhere. At the moment, it doesn't feel like it abstracts much away because of the reliance on paths.

dtasev:
> * It doesn't seem right for imaging to know how to build process lists. Might be over optimistic (probably a lot of work), but could it make sense to pull `savu_config` out of `savu` as it's own package, and use that?
> * If I read things correctly, imaging builds a process list which it saves as a file, then sends it as bytes to hebi, which saves it again, so savu can be pointed at it. At the least, would it make sense for imaging's version to be temporary, and deleted after sending?

You are right, but I think an easier approach would be for the API itself to build the process list and then call Savu with it, which solves both of these. It makes more sense for clients to not care about what is used internally to do the reconstruction. Maybe all that's needed are a few helper end-points to allow features we want, like "run process list so far and give me the output"

> * The endpoint for submitting a job to HEBI should be a POST to /jobs/queue
I think it is, [or something close](https://github.com/mantidproject/mantidimaging/blob/325_auto_start_savu_backend/mantidimaging/gui/windows/savu_filters/model.py#L206). The "queue" parameter is supposed to be a string with the name of the queue, e.g. `/jobs/preview` does a local reconstruct, `/jobs/cluster-name-1` does the reconstruction in that queue.

> * It would probably be nice for hebi to abstract away more of the file stuff, assigning ids of some kind and using them instead of file paths in query strings everywhere. At the moment, it doesn't feel like it abstracts much away because of the reliance on paths.

I agree, especially when dealing with 2 mostly separate filesystems it gets confusing. Then again because of that I am unsure how to enforce any structure beyond "use this folder for this and hope for the best".

docs/developer_guide/savu_hebi.rst
```diff
@@ -0,0 +1,48 @@
...
+--------------------------------
+
+How everything is connected:
+ - |container_setup|
+ - An http client is connected to hebi to submit jobs, ask for plugin details etc.
+ - A websocket client waits for messages from hebi, signalling when jobs complete.
+
+To apply a 'savu filter' to an image stack:
+ - |process_list_built|
+ - Submits a job to hebi with the pl and path to the input data.
```
dtasev:
```suggestion
 - Submits a job to hebi with the process list and path to the input data.
```

docs/developer_guide/savu_hebi.rst
```diff
@@ -0,0 +1,48 @@
...
+ - Hebi validates everything and starts savu as a subprocess.
+ - Once savu completes, hebi sends a message to imaging's ws client.
```
dtasev:
```suggestion
 - Once savu completes, hebi sends a message to imaging's WebSocket client.
```

docs/developer_guide/savu_hebi.rst
dtasev: There is one more step after this. After the job is submitted, the client checks the HTTP Response code. If it's 200 (HTTP OK) then it [joins](https://github.com/mantidproject/mantidimaging/blob/4deeaa847d5120a57b5fcec26df63986d0fbd814/mantidimaging/gui/windows/savu_filters/sio_model.py#L30) a WebSocket [room](https://socket.io/docs/rooms-and-namespaces/). The output from the job is sent to that "room", and clients in it get notified when something is broadcast into it.

WHTaylor:
> You are right, but I think an easier approach would be for the API itself to build the process list (via a `/process_list` endpoint) and then run Savu with it, which solves both of these. It makes more sense for clients to not care about what is used internally to do the reconstruction. Maybe all that's needed are a few helper end-points to allow features we want, like "run process list so far and give me the output"

I guess it depends on if hebi should change the interface to savu substantially? At the moment, it seems like hebi is primarily for abstracting where savu is running, but maybe it could also provide (in addition to, or possibly as a replacement to, the existing 'run a process list' workflow) something which, as you say, hides what is used internally.

Thinking about it more though, the clients would still need a format for 'run this filter with these parameters, then this one with these', and then hebi would translate that into a process list, and maybe we've just added another layer that's not very useful unless it's a much simpler format than the nxs process lists, and multiple programs end up using hebi.

> I think it is, [or something close](https://github.com/mantidproject/mantidimaging/blob/325_auto_start_savu_backend/mantidimaging/gui/windows/savu_filters/model.py#L206). The "queue" parameter is supposed to be a string with the name of the queue, e.g. `/jobs/preview` does a local reconstruct, `/jobs/cluster-name-1` does the reconstruction in that queue.

This was only a minor nitpick, not having `submit` in the URL is probably a bit more standard for REST as it's  implied by the request being a POST anyway.


> I agree, especially when dealing with 2 mostly separate filesystems it gets confusing. Then again because of that I am unsure how to enforce any structure beyond "use this folder for this and hope for the best".

Yeah, sticking with this is probably best.

Issue #318, opened 2019-08-06
Option to apply the history of one stack to another


dtasev:

WHTaylor:
Still required to complete issue:
- [ ] Tests (Some included in #399)
- [ ] Replay reconstructions (requires some reworks to way cor/proj angles are handled)
- [ ] Previews

PR #395, opened 2019-12-03
Retain history when copying operations

WHTaylor:
Part of #318


WHTaylor:
Going to combine a couple of PRs to avoid conflicts

PR #399, opened 2020-01-13
History copy fixes

WHTaylor:
Fixes accidental change from recording the class name to the module name introduced in #396 and improves a couple of tests.


WHTaylor:
(Part of #318)

PR #444, opened 2020-04-30
443 cupy remove outlier

### LINKED ISSUE #443: Add an implementation of remove outliers that uses cupy

DolicaAkelloEgwel:
Closes #443

Implements remove outlier filter. Also added some documentation for users and developers that should *hopefully* explain how things work. I am very much open to feedback if something in the documentation I provided doesn't make much sense.

I also want to emphasise what what I mention in the documentation isn't necessarily the best way to get GPU stuff working, it simply describes how I managed to solve the problems I encountered...


dtasev (commented):
Fixed the present merge conflicts, but I'm seeing the execution taking a lot longer than I would expect (each image takes a couple of seconds or so). I haven't had a chance to profile which step is at fault for this yet, but that would be a good first step

PR #451, opened 2020-06-12
Flat fielding stack selection

dtasev:
This PR adds quite a bit of refactoring, and starts the change of how data is handled internally in Mantid Imaging. This PR's major changes are:

- Loading now takes Flat and Dark images. It will also try to automatically find them based on the usual IMAT directory structure
  - Loading all 3 at the same time should create 3 stacks - one for sample, flat and dark
- Background correction is renamed to Flat fielding
- Flat fielding no longer wants file paths, now it uses other stacks
- Converting projections to sinograms now copies the data
- Adds Divide filter

You will need an IDAaaS PreProduction developer account.

### Setup
1. Create a PreProd Tomography development environment. 

2. Start a terminal from `Applications -> System -> Terminal`, then run `source /opt/miniconda/bin/activate /opt/miniconda/envs/mantidimaging`
   - Alternatively, run Mantid Imaging from `Applications -> Software -> Mantid Imaging` and then close it. The leftover terminal should have it

2. `pip install -U SharedArray`
    - This is a workaround for getting the new package, before the release is uploaded to `conda`

3. Copy test data from my babylon folder to your home folder. This `rsync` command will show you the progress of the copy too
    - `rsync --info=progress2 -ah /babylon/Public/dtasev/IAEA_Cylinder_WhiteBeam_Feb2019_PSI_Corr ~`

4. Clone Mantid Imaging repository, and enter the directory
    - `git clone https://github.com/mantidproject/mantidimaging`

5. Start it with `python -m mantidimaging`. The GUI should open

All commands
```bash
source /opt/miniconda/bin/activate /opt/miniconda/envs/mantidimaging
pip install -U SharedArray
rsync --info=progress2 -ah /babylon/Public/dtasev/IAEA_Cylinder_WhiteBeam_Feb2019_PSI_Corr ~
git clone https://github.com/mantidproject/mantidimaging

cd mantidimaging
python -m mantidimaging
```

### Sinograms copied on conversion
- File > Load, then select the projections (at `~/flower/IAEA_Cylinder_WhiteBeam_Feb2019_PSI_Corr/Tomo`). The Dark and Flat images should be automatically selected. Load those in
- Open `htop` in a terminal, or Applications -> System -> Task manager, then go on the `Resources" tab. Keep a track of the current memory usage
- Right click on the image, and do "Create sinograms from stack". The memory usage should (about) double
- Close one of the (big) stacks, the memory usage should reduce by the same amount. Close all stacks, you should be back to 

Note: if you copy too much (go over about 50GB RAM) Mantid Imaging will be terminated with a "Bus Error" (SIGBUS). This is a known issue and will be fixed in the IDAaaS configuration, more info: https://github.com/ddboline/shared-array#on-linux-i-get-segfaults-when-working-with-very-large-arrays

### Flat-fielding uses Stack selection
- File > Load images in, the flat and dark are needed for this step
- Create sinograms to verify that after applying flat-fielding to the main projections, the sinograms are unchanged!
- Go to Workflow > Filters (or CTRL+F)
- Find `Flat-fielding`
- Select the projection stack in the top selector (should not have "Flat" or "Dark" anywhere in the name, and is usually the shortest name)
- Flat and Dark should be selected automatically
- Make sure the preview works (change the stack for a bit, when selecting the right one you should be able to see the object)
- Make sure the "Invert difference" checkbox triggers a refresh, i.e. the difference gets inverted
- Apply, and check that the main stack has been changed

### Divide filter
- You can use the stack from the previous test steps
- Go to the Divide filter
- Run with `0.0000000` - the stack shouldn't change
- Run with something else, it shouldn't allow more than 7 significant digits
- This filter will be used to divide a reconstruction by the pixel size to get the attenuation coefficient, it's not supposed to do anything more complicated

Todo:
- [x] Add test for divide filter
- [x] Finish making the error not be in a pop-up window, but rather an in-window label

- (new issue) Cleanup of Images class (remove dark/flat) usage, instead handle as 3 independent Images objects


DanielMurphy22:
Test Instructions were so great, I initially forgot to pull and merge this actual PR!
- Are the build failures relevant?
- Have you added a test for the divide filter?
- Should the Filters interface be called `Filters` in the `Workflow` drop-down menu? It's currently called `Imaging Operations`
- In the divide filter, it is possible to set 11 significant digits (is this fine because only 7 decimal places?)
Screenshot 2020-06-15 at 11 41 24 (2)
- When using the Flat-fielding filter, setting the stack to `...Tomo_000_sino` (just to check the preview updated correctly) gave an uncaught exception: `TypeError: setPixmap(self, QPixmap): argument 1 has unexpected type 'StandardPixmap' `
Screenshot 2020-06-15 at 11 38 00 (2)
- the main functionality seemed to work fine!

dtasev:
Thank you very much for the review! I'll go over each question

> * Are the build failures relevant?
Yes, I'm working on fixing them now.

> * Have you added a test for the divide filter?
Just added it with the latest commit

> * Should the Filters interface be called `Filters` in the `Workflow` drop-down menu? It's currently called `Imaging Operations`
Done. I think it makes more sense for imaging to just call them filters (and in the code it's referred everywhere as filters).

> * In the divide filter, it is possible to set 11 significant digits (is this fine because only 7 decimal places?)
>   Screenshot 2020-06-15 at 11 41 24 (2)

Yes, my bad, it should say "up to 7 decimal places" (the precision of float32)! I'll fix that in the PR.

> * When using the Flat-fielding filter, setting the stack to `...Tomo_000_sino` (just to check the preview updated correctly) gave an uncaught exception: `TypeError: setPixmap(self, QPixmap): argument 1 has unexpected type 'StandardPixmap' `
>   Screenshot 2020-06-15 at 11 38 00 (2)
> * the main functionality seemed to work fine!

I think that is part of an unfinished error widget. I'll see if it's worth keeping here or moving onto a new PR.

PR #518, opened 2020-09-16
Adds Go to projection and angle

### LINKED ISSUE #501: Go to N proj or N angle

dtasev:
Adds Go to Projection and Angle in the Stack Visualiser context menu.

To test:
- Load some data
- Test that `Go to projection`, goes to the correct projection (test negative, more than max, etc)
- Test that `Go to angle` goes to the correct angle (in degrees)

Fixes #501


Pasarus (approved):
Looks good :+1:

PR #520, opened 2020-09-17
Refactors some logic bits from MIImageView into a presenter

### LINKED ISSUE #517: MIImageView refactor and test
MIImageView class is a view that contains too much logic. E.g. functions such as `_update_roi_region_avg`, `get_roi` don't really interact with the Qt part - only the image data.

The parts that do not interact with Qt should be moved and tested in a presenter.

dtasev:
Refactors two functions that don't need to be in the view into a presenter, and adds tests for them. 

To test: 
- Check that loading data doesn't have any errors when visualising a stack
- Check that clicking on the histogram changes to the correct place
- Check that dragging the ROI around is OK, and running a filter with it (e.g. Crop coords) runs OK

Fixes #517


Pasarus (approved):
Looks good :+1:

PR #523, opened 2020-09-21
Add a filter help button in the operations window

### LINKED ISSUE #514: Add help button that redirects to correct github page for each filter

Pasarus:
To test:
Open operations window
Click the help button (the ? button next to filter choices) and a website should open
Check that each filter opens the correct website that correlates to the correct filter

Fixes #514


dtasev (changes requested):
Button works and goes to the correct page, but would it be possible to have it positioned in bottom-left?

dtasev (approved):

PR #563, opened 2020-10-02
Automatically applies minus log before reconstruction

### LINKED ISSUE #550: Automatically apply minus log to the data before reconstruction
This needs to be careful to not modify the original source data! Maybe the minus log filter can be run right on the step before reconstruction call.

Note that currently there's two reconstructors, it should be added in a way that it applies to both from a single location (i.e. like a parent function they both call)

dtasev:
Also removes the `minus_log` operation as it is no longer necessary.

To test:
- Load a dataset
- Go right into reconstruction window
- Run correlate to get some COR
- Reconstructed image on the right should look OK and the values in a range of around 0.000-0.0100 (depends on sample)
- Run a full recon and see that the output looks the same as the image in the preview

Fixes #550


Pasarus (changes requested):
The PR's test suggestions lead me to an issue where the recon and the recon_preview do not look the same.

dtasev:
Is this ready for merging?

Pasarus:
No, we need to discuss these results, I am still waiting on you to be free haha

PR #569, opened 2020-10-06
Allow users to cancel a task in progress

### LINKED ISSUE #544: X on progress doesn't abort task

Pasarus:
Work:
Currently you can abort an in progress task, this literally terminates the thread in place and stops execution, it is destructive so a user must be sure so a second dialog is used to confirm this choice.

To Test:

- Load some data
- Open Operations window
- Start a "long" (enough) operation where you have the chance to click the x button on the progress window
- Click yes and observe it stops the operation
- Repeat but clicking No and observe it doesn't stop the operation

Fixed #544

### DIFF
--- a/mantidimaging/gui/dialogs/async_task/model.py
+++ b/mantidimaging/gui/dialogs/async_task/model.py
@@ -19,6 +19,14 @@ class AsyncTaskDialogModel(Qt.QObject):
 
         self.on_complete_function: Callable = lambda: None
 
+    def do_cancel_task(self):
+        try:
+            self.task.terminate()
+            self.task.wait()
+        except Exception as e:
+            # This is going to throw
+            getLogger(__name__).exception("Terminated thread: " + str(e))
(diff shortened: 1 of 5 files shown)


dtasev (changes requested):
Trying to stop while loading a dataset will break the process very badly, it's hard to shutdown without a `kill -9`. Stopping an operation seems to work but the CPU usage still goes on so I don't think it stops - this is likely because even though the process terminate is called, the underlying multiprocessing Pool is not closed, so it will keep going until it finishes and `.join()`s

Also a few thoughts:
- We should probably focus on making this work for stopping loading. This means we need an error/terminated callback that will free the allocated shared array if the user stops the loading
- Cancelling an operation is of debatable usefulness, especially when every time it will corrupt a portion of the dataset. And it might not be as important once we get #578

Pasarus:
This will no longer be a good implementation.

PR #571, opened 2020-10-07
Add Rescale to ROI Normalisation

### LINKED ISSUE #548: Flat-fielding after ROI normalisation produces garbage
This might happen because we start hitting float 32 precision limit due to dividing low values with low values.

To replicate: Apply ROI normalisation to data and then do flat-fielding.

Potential fix:
A way around this would be for ROI normalisation to keep the `images.data.min()/images.data.max()` before applying the normalisation, and then re-scale back to that range. This means the resulting data will continue to match the range of the flat/dark images and will not produce garbage result.

Pasarus:
Fixes #548 

To Test:
- Load data
- Open operations
- Perform a ROI normalization operation
- Perform a Flat-fielding operation
- The stack should not look like garbage

### DIFF
--- a/mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
+++ b/mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
@@ -7,6 +7,7 @@ import numpy as np
 from mantidimaging import helper as h
 from mantidimaging.core.data import Images
 from mantidimaging.core.operations.base_filter import BaseFilter
+from mantidimaging.core.operations.rescale.rescale import RescaleFilter
 from mantidimaging.core.parallel import two_shared_mem as ptsm
 from mantidimaging.core.parallel import utility as pu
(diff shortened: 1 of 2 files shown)


mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
```diff
@@ -46,8 +47,16 @@ def filter_func(images: Images, air_region: SensibleROI = None, cores=None, chun
 
         # just get data reference
         if air_region:
+            initial_image_max = images.data.max()
+
             progress = Progress.ensure_instance(progress, task_name='ROI Normalisation')
             _execute(images.data, air_region, cores, chunksize, progress)
+
```
dtasev, line 54:
```suggestion
            progress.update(1, "Rescaling to input value range")
```
Updates the progress message to say that the data is being rescaled, otherwise it stays on "complete" for a few seconds

PR #574, opened 2020-10-07
Ensure new stacks are added as a tab instead of windows

### LINKED ISSUE #553: Adding a new stack should be added as a tab on the current window
The current behaviour will open a new stack in a new window inside the QDockWidget, essentially splitting the window in two:
- To replicate, load a dataset twice - you'll get two windows, split roughly at the middle

What we want to move towards is having that appear as a single big window, and every new stack as a tab (like how Flat and Dark images are loaded already).

We should first add this behaviour to the reconstructed volume - it should be added as a new tab. The loading a completely new dataset behaviour doesn't need to be changed at this point, but if the change affects both that is OK.

Pasarus:
Fixes #553 

To test:
- Load data
- Load more data
- Notice there is more tabs

### DIFF
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -104,6 +104,10 @@ class MainWindowPresenter(BasePresenter):
         sample_dock, sample_stack_vis = self._make_stack_window(sample, title)
         self.model.add_stack(sample_stack_vis, sample_dock)
 
+        current_stack_visualisers = self.get_all_stack_visualisers()
+        if len(current_stack_visualisers) > 1:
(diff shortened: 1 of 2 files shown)


dtasev (changes requested):
Can the new stack steal focus? I imagine a lot of users will not see the new tab because it just silently appears

Pasarus:
Whilst yes it can steal focus, it will result in dark or flat taking focus over the main tomo stack, we have current definition of what is a tomo vs a dark bar the name at the moment, how do you suggest I go about selecting to focus the correct stack?

dtasev:
I think loading is the exception of the focus switching - we want to stay on the sample data, I guess we can just focus it at the end, or prevent the focus changing signal from happening while loading. Apart from that the latest stack that has been added should steal the focus.

PR #596, opened 2020-10-15
Focus last added stack

### LINKED ISSUE #595: Focus newest added stack
#574 makes it so that all stacks are added to the main tab bar, but they are not focussed.

The newest stack should be focussed after it has been added to the main window. The only exception being after loading: the sample stack should remain focussed.

dtasev:
Focuses the last added stack to the main window.

Not sure if there's a better way that can avoid calling the `.processEvents()`

Fixes #595

### DIFF
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -121,7 +121,7 @@ class MainWindowPresenter(BasePresenter):
             if tab_bar is not None:
                 last_stack_pos = len(current_stack_visualisers) - 1
                 # make Qt process the addition of the dock onto the main window
-                QApplication.processEvents()
+                QApplication.sendPostedEvents()
                 tab_bar.setCurrentIndex(last_stack_pos)
 
         self.view.active_stacks_changed.emit()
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/windows/main/presenter.py
```diff
@@ -116,6 +116,14 @@ def create_new_stack(self, container: Union[Images, Dataset], title: str):
             if container.sample.has_proj180deg() and container.sample.proj180deg.filenames:
                 self._add_stack(container.sample.proj180deg, container.sample.proj180deg.filenames[0], sample_dock)
 
+        if len(current_stack_visualisers) > 1:
+            tab_bar = self.view.findChild(QTabBar)
+            if tab_bar is not None:
+                last_stack_pos = len(current_stack_visualisers) - 1
+                # make Qt process the addition of the dock onto the main window
+                QApplication.processEvents()
```
Pasarus, line 124: I wonder if QApplication.sendPostedEvents() will do what we want here?

dtasev: 👍 Works with that

Pasarus (approved):

PR #599, opened 2020-10-16
Adds tests for MainWindowView

### LINKED ISSUE #598: version_check.py goes wrong if the environment name contains `mantidimaging` AND the `mantidimaging` package is not installed
Discovered when running `mantidimaging` in Docker from #597, but should also be replicable without docker if:
- You environment has `mantidimaging` in the name - i.e. the Docker one is called `mantidimaging_test`
- There is no `mantidimaging` package

All you need to do is run `mantidimaging` and it should fall over at the start when the version is being checked

### To fix
`conda list` takes a regex argument, so instead of `grep`ing for the `mantidimaging` package name, we should just
- `conda list mantidimaging`
- Discard all lines from the output that start with `#`
- What's left is either: 
  - 0 length list = mantidimaging not installed
  - 1 length list = the `mantidimaging` package
  - 1 or more length list = any package that contains `mantidimaging`
- Perhaps the regex could be better than just matching `mantidimaging` and could try to match the whole version `mantidimaging_MJ_MN_PT_commit`

dtasev:
Adds a bunch of missing tests for the view. 

Check that tests pass

Fixes #598

### DIFF
--- a/mantidimaging/gui/windows/main/test/view_test.py
+++ b/mantidimaging/gui/windows/main/test/view_test.py
@@ -1,5 +1,6 @@
 # Copyright (C) 2020 ISIS Rutherford Appleton Laboratory UKRI
 # SPDX - License - Identifier: GPL-3.0-or-later
+
 import unittest
 import mock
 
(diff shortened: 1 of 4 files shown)


dtasev:
Tests for this are failing due to #598, I'll look at fixing that first

Pasarus (approved):

PR #602, opened 2020-10-16
Adds a check for available memory before allocating

### LINKED ISSUE #600: Try to prevent SIGBUS (Bus Error) when requesting too much memory
Currently Mantid Imaging will be terminated by the OS with SIGBUS if it tries to allocate memory over the maximum physically available.

We could _try_ to prevent this with a check in `create_shared_array` that calculates requested data size (`shape*dtype`) and compares it to available memory.

The helper functions for this should already be available, calculating data size is used in the load dialog, and I think available memory is sitting in a utility file, but not used anymore.

dtasev:
Adds a free memory getter in memory_usage package, and changes
get_memory_usage_linux to use psutil which should also make it run on Windows

### To test
- Try to load a dataset that's over your available RAM, or duplicate the data a bunch of times to reach that.
- Run Rebin or Crop on data where the temporary necessary for the operation cannot be allocated
  - Rebin to anything that's super big 
  - To get Crop to fail you can just crop by 99% of the image size, so as long you're somewhere above 50% memory usage

Fixes #600

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -37,6 +37,7 @@ requirements:
     - pyqt
     - pyqtgraph=0.11
     - sarepy
+    - psutil
 
 build:
   number: 0
--- a/environment.yml
+++ b/environment.yml
@@ -27,4 +27,5 @@ dependencies:
       - sphinx
       - pytest-randomly
       - pytest-parallel
+      - psutil
       - isort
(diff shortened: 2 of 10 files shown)


Pasarus (changes requested):
Largely looks good, can we add a way to ensure the progress window is closed/never opens? otherwise I get a fun little error:
image

dtasev:
Which action caused this? Load, applying a filter, recon, etc - they all start and show the progress a bit differently so it'd be good to know to track it down

Pasarus:
Loading

dtasev:
I can't replicate this - I just get the error and no progress bar opens at all.

dtasev:
Also I wonder if it might be better to check versus 95-97% of the actual available RAM - it seems a bit optimistic to expect to be 100% accurate, and I'd rather warn preemptively than get SIGBUSsed

Pasarus:
That sounds like a good solution, because I managed to load all but 5 images of a 500ish stack so that inaccuracy is likely.

Pasarus (approved):

Issue #588, opened 2020-10-13
Separate auto-find methods into two buttons


dtasev:
Currently it's a drop down to select method (Correlate/Minimise sqsum), but it's not immediately obvious you can switch.

As there's only 2 it might be worth just changing them to two buttons with the corresponding name of the method.

PR #611, opened 2020-10-20
Fix disabling the COR buttons

dtasev:
The Auto find COR button was disabled during operation in #590, but the layout was changed in a PR at the same time #588 so the disabling of the buttons ended up out of sync with the layout.

This PR fixes this

### To test
- Load a dataset and go into the recon window
- Check that correlate and minimise auto-find COR buttons execute

No associated issue


DolicaAkelloEgwel (approved):
Just tried it out and I was able to run auto correlate and auto minimise error.

PR #616, opened 2020-10-22
Adds ROI field to Crop and ROI Norm 

dtasev:
Adds a ROI field that takes a string - this allows users to paste a known ROI, or use Select ROI to find a good one.

- No longer uses "stack parameters" concept - all params used for the operation should be coming from a field
- The Select ROI is now in a separate window, which is a child of the Operation View
- Moving the ROI in the Select ROI now updates the roi_field, and updates the previews

### To test
1. Load a dataset
1. Open Operations (CTRL+F)
1. Go to Crop, click select ROI, moving the red ROI square should update the preview image
1. Clicking OK should close the ROI select
1. Applying the filter should crop the data properly
1. Check that the history properly reflects the ROI used
1. Repeat 3 onwards but with ROI Normalise

### DIFF
--- a/mantidimaging/gui/utility/qt_helpers.py
+++ b/mantidimaging/gui/utility/qt_helpers.py
@@ -129,8 +129,7 @@ def add_property_to_form(label: str,
             box.setValue(cast_func(default_value))
 
     # some of these are used dynamically by Savu Operations GUI and will not show up in a grep
-    if dtype == 'str' or dtype == 'tuple' or dtype == "NoneType" or dtype == "list" or \
-            dtype == Type.STR or dtype == Type.TUPLE or dtype == Type.NONETYPE or dtype == Type.LIST:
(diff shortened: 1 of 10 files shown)


mantidimaging/gui/utility/qt_helpers.py
```diff
@@ -129,7 +129,8 @@ def set_spin_box(box, cast_func):
             box.setValue(cast_func(default_value))
 
     # some of these are used dynamically by Savu Operations GUI and will not show up in a grep
-    if dtype == 'str' or dtype == 'tuple' or dtype == "NoneType" or dtype == "list":
+    if dtype == 'str' or dtype == 'tuple' or dtype == "NoneType" or dtype == "list" or \
```
DolicaAkelloEgwel, line 132: Could this be replaced with a list?

dtasev: Done 👍

DolicaAkelloEgwel (approved):

PR #628, opened 2020-10-29
Sets pixel size field when stack is changed 

### LINKED ISSUE #627: Pixel size field not updated when stack is changed in Reconstruction Window
The default value of the Pixel Size parameter field is only set inside the Reconstruction Window initialiser, so changing the stack afterwards won't change the pixel size value.

The presenter handles the stack change inside `set_stack_uuid` - it might be better to move the pixel size field setting code to this location.

dtasev:
- Updates test to check for new behaviour
- Changes recon pixel size spinbox to a `QSpinBox`
- Moves some attributes of the `recon.model` to properties 

Fixes #627

### DIFF
--- a/mantidimaging/gui/ui/recon_window.ui
+++ b/mantidimaging/gui/ui/recon_window.ui
@@ -334,18 +334,30 @@
            <layout class="QVBoxLayout" name="verticalLayout_6">
             <item>
              <layout class="QGridLayout" name="optionsLayout">
-              <item row="2" column="0">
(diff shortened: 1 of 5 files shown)


Pasarus:
I get an AttributeError

  File "/home/sam/mantidimaging/mantidimaging/gui/windows/recon/presenter.py", line 115, in set_stack_uuid
    self.view.pixel_size = self.get_pixel_size_from_images()
AttributeError: can't set attribute

Pasarus (changes requested):
above

PR #630, opened 2020-10-29
Use the pixel size provided in recon preview image

### LINKED ISSUE #626: Use Pixel Size when reconstructing for preview
Currently the reconstruction preview in the recon window does not take into account the pixel size.

The pixel size is available through the Pixel Size parameter field. Note that this field takes the input in microns (value gets multiplied by `1e-4`)

The pixel size should also be used when reconstructing the previews, allowing a more accurate preview. This is currently only done for the full stack reconstruction (reconstruct volume).

dtasev:
Now uses pixel size in recon preview and reconstruct slice. Reconstruct volume already used it so nothing should be changed there.

Changing the histogram also resets the histogram range on the recon slice, so if you manually scroll up the pixel size you should see the range remaining at the edges of the data range.

Fixes #626

### DIFF
--- a/mantidimaging/gui/windows/recon/image_view.py
+++ b/mantidimaging/gui/windows/recon/image_view.py
@@ -115,3 +115,6 @@ class ReconImagesView(GraphicsLayoutWidget):
                 self.tilt_line.setPos(pos)
             self.tilt_line.setAngle(90 + tilt.value)
         self.projection_vb.addItem(self.tilt_line)
+
+    def reset_recon_histogram(self):
+        self.recon_hist.autoHistogramRange()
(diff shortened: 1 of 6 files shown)


Pasarus:
The preview doesn't seem to use pixel size 1 micron when it opens and still shows me: 
image
Would it be possible to make it do the same for when it first opens the window? To get it to do that it will probably just be setting the pixel size to 1 micron before generating the images.

Pasarus (changes requested):
Above

dtasev:
I rebased on master which has #628 and I don't think I can replicate this anymore, I did the following steps:

- Load data, set the pixel size to 1
- Go to recon window after loading
- The pixel size field was 1 and the data had been divided correctly

Can you check if this works for you?

Pasarus (approved):
It seems to work now :/ sometimes it be like that!

PR #631, opened 2020-10-30
Compare Operation Window

### LINKED ISSUE #578: Compare operation result with original data and offer choice to pick best looking one (or keep both)
The idea of this is to allow users to apply operations without too much fear or destroying their data (and having to reload). This will also take care of #554.

### To do
- Add a new option in Operations window, which is "Safe Apply" (better name is welcome!)
	- Safe Apply could be a checkbox too, but it probably shouldn't be used on the Apply to All, unless it can be easily done
- This option will copy the data, and then apply the operation on the copy
- After the operation is done it will display a new window with the FilterPreview widget in it, and a few buttons to allow the user to select the better image
Some thoughts on design, the first one might be more feasible (and the layout is very similar to the COR Refine window!)

image
image

### Thoughts

- After this is done consider whether the `Duplicate whole data` option is still needed
  - I think the `Duplicate current ROI` is still useful as it crops too - useful for cropping datasets that contain multiple samples

- Be careful that algorithms that change the shape (i.e. rebin, crop) don't accidentally free the original data
- This could be broken up into two pieces of work
  - Implement the data duplication and display of results in a window using the image view from the main window
  - Iterate on what else will be useful to see
  - Actually - it might be possible to just reuse the FilterPreview class from the Operations window, but with bigger image views..
- What happens if you apply to copy but run out of memory when:
  - The data is being copied
  - The data is copied, but there's not enough memory to apply the operation afterwards (free on fail, or cache?)

Pasarus:
Fixes #578 

To Test:
- Load Data
- Open Operations window
- Ensure safe apply is checked
- Ensure that all options in the GUI are working
- Ensure that all ROI and when changing images occurs, it syncs across both stacks.
- Check that normal operations occur like before if safe apply is unchecked


mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -106,6 +111,9 @@ def filter_uses_parameter(self, parameter):
             self.model.params_needed_from_stack is not None else False
 
     def do_apply_filter(self):
+        if self.view.safeApply.isChecked():
+            self.original_images_stack = self.stack.presenter.images.copy()
+
         self.view.clear_previews()
         apply_to = [self.stack]
         if self.stack.presenter.images.has_proj180deg():
```
dtasev, line 119: Having a 180 degree projection loaded causes an error as it tries to open the compare dialog for it as well

dtasev (changes requested):
- Needs to have some prompt while it's copying data otherwise the GUI looks frozen with bigger datasets (grep for `with operation_in_progress` -> can just stick the image.copy() inside that and it will show a wait message)
- Contrast syncing for the image comparison? Perhaps we can do this in a follow up issue so that we can handle flat fielding and big contrast differences appropriately 
- I seem to get a progress bar left over when applying circular mask - can you replicate that?

Functionally it works very well!

mantidimaging/gui/ui/filters_window.ui
```diff
@@ -369,6 +375,25 @@
...
+             <string>Give the option to choose which stack to use after a filter is applied, RAM intensive</string>
+            </property>
+            <property name="text">
+             <string>Safe Apply</string>
+            </property>
+            <property name="checked">
+             <bool>true</bool>
+            </property>
+           </widget>
+          </item>
```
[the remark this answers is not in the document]
Pasarus: Done :+1:

Pasarus:
I think a follow up issue with "Contrast syncing for the image comparison" would be best as in multiple places it would cause comparison issues in multiple filters.

mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -120,19 +125,54 @@ def do_apply_filter_to_all(self):
...
+            for stack in stacks:
+                self.original_images_stack.append((stack.presenter.images.copy(), stack.uuid))
 
+        if len(stacks) > 0:
+            self.applying_to_all = True
         self._do_apply_filter(stacks)
 
-    def _post_filter(self, updated_stacks, task):
+    def _wait_for_stack_choice(self, new_stack, stack_uuid):
+        stack_choice = StackChoicePresenter(self.original_images_stack, new_stack, self, stack_uuid)
```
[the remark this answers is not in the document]
Pasarus: That's already the parent of the operation window. It's actually assigned in the view.py of the StackChoiceView. It unfortunately does not do that with QMainWindow

mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -120,19 +127,54 @@ def do_apply_filter_to_all(self):
         if not confirmed:
             return
         stacks = self.main_window.get_all_stack_visualisers()
+        if self.view.safeApply.isChecked():
+            self.original_images_stack = []
+            for stack in stacks:
+                self.original_images_stack.append((stack.presenter.images.copy(), stack.uuid))
```
dtasev, line 133: The with `operation_in_progress` should also be done here

dtasev (changes requested):

mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -120,19 +127,54 @@ def do_apply_filter_to_all(self):
...
+            if stack.presenter.images.proj180deg == stack_to_check.presenter.images:
+                return True
+        return False
+
+    def _post_filter(self, updated_stacks, task):
+        do_180deg = True
         for stack in updated_stacks:
+            is_a_proj180deg = self.is_a_proj180deg(stack)
+            if self.view.safeApply.isChecked() and not is_a_proj180deg:
+                do_180deg = self._wait_for_stack_choice(stack.presenter.images, stack.uuid)
```
dtasev, line 164: This should check if the task failed before showing the comparison (and before applying to the 180 degree actually), because currently if you do:

- Load some data that you can copy for safe apply
- Rebin to 9999 9999
  - The goal is to get the copy to succeed - but the filter allocation to fail

I got:
- Comparison with an unchanged stack (as the filter failed)
- Proj 180 was just gone

It should probably just show an error and pretend the user clicked "keep original"

dtasev (changes requested):

mantidimaging/gui/windows/operations/presenter.py
Pasarus: Done!

mantidimaging/gui/windows/operations/presenter.py
Pasarus: Should be sorted now I think :+1:

mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -158,23 +159,29 @@ def is_a_proj180deg(self, stack_to_check):
...
-            is_a_proj180deg = self.is_a_proj180deg(stack)
-            if self.view.safeApply.isChecked() and not is_a_proj180deg:
-                do_180deg = self._wait_for_stack_choice(stack.presenter.images, stack.uuid)
-            self.view.main_window.update_stack_with_images(stack.presenter.images)
-
-            if stack.presenter.images.has_proj180deg() and do_180deg and not is_a_proj180deg \
-                    and not self.applying_to_all:
-                self.view.clear_previews()
-                self._do_apply_filter([self.view.main_window.get_stack_with_images(stack.presenter.images.proj180deg)])
+            # Check if the stack was an error and fix if error occured else do normal bits
```
dtasev, line 164: Can this be changed to something succinct, maybe:

```
If the operation encountered an error during processing, try to restore the original data
else continue processing as usual
```

dtasev (changes requested):
Now works well with error during operation. Other functionality still works as expected.

Just a small suggestion for improving the comment above the attempt repair logic

PR #632, opened 2020-11-02
Handles missing projection or projection angle

dtasev:
Verifies the number of images versus the number of proj angles in log 

- If they mismatch it raises a runtime error
- It tries to find the missing projection / image file

It shows an error message and should not select the log. The error message ought to prompt the user to inspect their data and/or logs and fix the change (we can't really do anything!)

### To test
- The flower dataset should have logs in it
- Copy your image files and the sample log, as you'll be editing those and don't want to lose the main copy
- Delete any image file 
- Try loading the data - it should tell you there's a missing image file (projection), and do a best guess as to which one (number might be off by 1 for the image file)
  - Edge cases are: missing first or last image file, try those too

---------------

- Restore the image file and delete a line from the log
- Try loading the data again - this time it should tell you there's a missing angle
  - Same edge case as above

---------------

- Keep the deleted line in the log and delete A DIFFERENT image file 
- Try loading the data - it should fail again

---------------

- Restore the previous projection, keep the deleted line in the log and delete the SAME image file as the line in the log
- This should now load as there is no mismatching information

### DIFF
--- a/mantidimaging/core/data/images.py
+++ b/mantidimaging/core/data/images.py
@@ -119,13 +119,13 @@ class Images:
 
         self.metadata[const.OPERATION_HISTORY].append({
             const.TIMESTAMP:
-                datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
+            datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
             const.OPERATION_NAME:
-                func_name,
+            func_name,
             const.OPERATION_KEYWORD_ARGS: {k: prepare(v)
                                            for k, v in kwargs.items() if accepted_type(v)},
             const.OPERATION_DISPLAY_NAME:
-                display_name
+            display_name
         })
 
     def copy(self, flip_axes=False) -> 'Images':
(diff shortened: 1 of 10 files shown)


Pasarus (changes requested):
When attempting to load the tile 1 pre reco. I get FileNotFoundError: [Errno 2] No such file or directory: ''
```
2020-11-05 09:27:25,471 [mantidimaging.gui.windows.main.view:L221] ERROR: Traceback (most recent call last):
  File "/home/sam/mantidimaging/mantidimaging/gui/windows/load_dialog/view.py", line 43, in <lambda>
    self.select_sample.clicked.connect(lambda: self.presenter.notify(Notification.UPDATE_ALL_FIELDS))
  File "/home/sam/mantidimaging/mantidimaging/gui/windows/load_dialog/presenter.py", line 39, in notify
    self.do_update_sample()
  File "/home/sam/mantidimaging/mantidimaging/gui/windows/load_dialog/presenter.py", line 83, in do_update_sample
    self.last_file_info.filenames)
  File "/home/sam/mantidimaging/mantidimaging/gui/windows/load_dialog/presenter.py", line 209, in set_sample_log
    self.ensure_sample_log_consistency(sample_log, sample_log_filepath, image_filenames)
  File "/home/sam/mantidimaging/mantidimaging/gui/windows/load_dialog/presenter.py", line 203, in ensure_sample_log_consistency
    log = load_log(file_name)
  File "/home/sam/mantidimaging/mantidimaging/core/io/loader/loader.py", line 99, in load_log
    with open(log_file, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: ''

```

dtasev:
I did a hasty merge conflict resolution, and while everything looked OK I haven't verified that all the functionality is still working - please retest fully

Pasarus (approved):

PR #633, opened 2020-11-03
Fixes divide ignoring the unit from op window

dtasev:
- The value of the widget was not actually being passed in

No issue associated


Pasarus (approved):
Looks good!

PR #634, opened 2020-11-03
475: Refine iterations window

### LINKED ISSUE #475: Refine window, but for iterations
Add a similar dialog to the Refine window but instead of COR it refines the number of iterations. The Refine window should be re-usable for this

DolicaAkelloEgwel:
Closes #475 

Didn't get very far with having it switch between two modes, so this is the solution I currently have.


dtasev (changes requested):
I'm not sure if this is ready for review, but I'd like to note that we don't want to add a full new GUI unless necessary, as the functionality is too similar to the `cor_inspection` one. I think the `adjust_iterations` functionality from the new one can be merged into the original `cor_inspection`

dtasev (changes requested):
Can the refine iterations button be moved to the reconstruct tab? Perhaps above the help button 
image.

It should also be active only if SIRT has been selected

dtasev (approved):

dtasev:
Looks really good and all works!

PR #635, opened 2020-11-04
Load flat and dark, before and after

### LINKED ISSUE #477: Load flat/dark images from both `N_After` and `N_Before`
#### Current
Only a single instance of Flat/Dark images is found, either Before or After.

#### Desired
Find both Flat/Dark Before and After. Load as separate stacks, and add the option to use for flat-fielding:
- Only Before
- Only After
- Both, concatenated

If both flat and dark are present this should load 4 stacks: `Flat_After`, `Flat_Before`, `Dark_After`, `Dark_Before`

A more advanced feature would be to:
- Apply `X_Before` to first half, `X_After` to second half
  - Even better - specify indices to apply before to, and after to. Default to half/half

----------

Notes:
- This can/should all be handled in the `flat-fielding` filter, by extending the available parameters for the filter

```
Flat Before: <stack selector widget>
Apply To: <start, end index QSpinBox>
... Same for flat after, dark before/after
```

Pasarus:
Fixes #477 

Test:
- Open mantidimaging
- Load Data that has before and after for flat and dark
- Open operations
- Select flat fielding operation
- Test out the data and ensure it "looks right"

### DIFF
--- a/mantidimaging/core/data/dataset.py
+++ b/mantidimaging/core/data/dataset.py
@@ -7,5 +7,7 @@ from mantidimaging.core.data import Images
 @dataclass
 class Dataset:
     sample: Images
-    flat: Optional[Images] = None
-    dark: Optional[Images] = None
+    flat_before: Optional[Images] = None
+    flat_after: Optional[Images] = None
+    dark_before: Optional[Images] = None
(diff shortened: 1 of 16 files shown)


dtasev (approved):
Looks good and works well

PR #640, opened 2020-11-05
Disable SAVU window and apply history to new stack

### LINKED ISSUE #614: Evaluate features necessity for Octopus Replacement release
There are some features present that are of questionable necessity for the next release (Octopus Replacement), and time might be better spent on something else, rather than maintaining them.

Features in mind:
- Savu support - it is unstable at best, if manages to work at all. While potentially beneficial, it's real benefit and necessity to satisfy the requirements for the next release is at this point none.
	- Estimate to make it usable: very long, past the project release date
- Applying history from one stack to another
	- This might be a quick win but needs to be properly manually tested
	- If time cannot be allocated to test and fix the history re-apply, it would be best to disable it for the next release, and come back to it afterwards
- ... Other features that are not explicitly required for the project delivery?

dtasev:
These features are disabled as they are unfinished/unstable and are not essential for the project delivery

### To test
- Check that the GUI still opens and all the buttons from the main window menus still work
- Load some data
- Right click on the image and check that the Apply history option is gone

Fixes #614

### DIFF
--- a/mantidimaging/gui/windows/main/view.py
+++ b/mantidimaging/gui/windows/main/view.py
@@ -61,7 +61,7 @@ class MainWindowView(BaseMainWindowView):
         self.actionAbout.triggered.connect(self.show_about)
 
         self.actionFilters.triggered.connect(self.show_filters_window)
-        self.actionSavuFilters.triggered.connect(self.show_savu_filters_window)
+        # self.actionSavuFilters.triggered.connect(self.show_savu_filters_window)
         self.actionRecon.triggered.connect(self.show_recon_window)
 
         self.active_stacks_changed.connect(self.update_shortcuts)
--- a/mantidimaging/gui/windows/main/view.py
+++ b/mantidimaging/gui/windows/main/view.py
@@ -61,7 +61,6 @@ class MainWindowView(BaseMainWindowView):
         self.actionAbout.triggered.connect(self.show_about)
 
         self.actionFilters.triggered.connect(self.show_filters_window)
-        # self.actionSavuFilters.triggered.connect(self.show_savu_filters_window)
         self.actionRecon.triggered.connect(self.show_recon_window)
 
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/windows/main/view.py
```diff
@@ -61,7 +61,7 @@ def setup_shortcuts(self):
         self.actionAbout.triggered.connect(self.show_about)
 
         self.actionFilters.triggered.connect(self.show_filters_window)
-        self.actionSavuFilters.triggered.connect(self.show_savu_filters_window)
+        # self.actionSavuFilters.triggered.connect(self.show_savu_filters_window)
```
Pasarus, line 64: Is it possible to delete this line?

mantidimaging/gui/windows/stack_visualiser/view.py
```diff
@@ -108,16 +108,20 @@ def close_view(self):
...
-                   ("Goto angle", self.goto_angle)]
+        actions = [
+            ("Set ROI", self.set_roi),
+            ("Copy ROI to clipboard", self.copy_roi_to_clipboard),
+            ("Toggle show averaged image", lambda: self.presenter.notify(SVNotification.TOGGLE_IMAGE_MODE)),
+            ("Create sinograms from stack", lambda: self.presenter.notify(SVNotification.SWAP_AXES)),
+            ("Duplicate whole data", lambda: self.presenter.notify(SVNotification.DUPE_STACK)),
+            ("Duplicate current ROI of data", lambda: self.presenter.notify(SVNotification.DUPE_STACK_ROI)),
+            ("Show history", self.show_image_metadata),
+            #    ("Apply history from another stack", self.show_op_history_copy_dialog),
```
Pasarus, line 119: Could we also delete this line?

Pasarus (changes requested):
Just a few lines that are commented that we should probably remove opposed to comment out.

PR #693, opened 2020-11-17
Adds initial reconstruction docs

### LINKED ISSUE #663: Reconstruction filter description - link to ASTRA?
Except that ASTRA itself has no filter descriptions so I guess we point to our own docs page and we will be filling it out for now.

dtasev:
Adds initial reconstruction section as part of the user guide, and some docs for the current available algorithms. Adds some brief explanation as well, because the actual ASTRA and TomoPy docs don't really say anything themselves..

- Also refactors the help calling code to make it more re-usable

To test:

- Check that operation pages still open properly
- Check that the reconstruction help button (in the reconstruct tab) opens to the correct URL - you will get a 404 as this PR needs to be published to see it.
- Have a read through the `reconstruction/algorithms.rst`

Fixes #663


mantidimaging/core/reconstruct/astra_recon.py
```diff
@@ -18,7 +18,8 @@
 
 
 # Full credit for following code to Daniil Kazantzev
-# Source: https://github.com/dkazanc/ToMoBAR/blob/master/src/Python/tomobar/supp/astraOP.py#L20-L70
+# Source:
+# https://github.com/dkazanc/ToMoBAR/blob/5990aaa264e2f08bd9b0069c8847e5021fbf2ee2/src/Python/tomobar/supp/astraOP.py#L20-L70
```
dtasev, line 22: moved to new line because flake8 complains for line too long otherwise

mantidimaging/gui/ui/recon_window.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1495</width>
```
Pasarus, line 9: Would it be worth defining the height and weight on a more permanent section as it seems we keep setting new values for this probably by accident in the designer.

Pasarus (changes requested):
Just some comments

mantidimaging/gui/ui/recon_window.ui
```diff
@@ -37,7 +37,7 @@
            </sizepolicy>
           </property>
           <property name="currentIndex">
-           <number>0</number>
+           <number>1</number>
```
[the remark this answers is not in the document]
dtasev: Good catch.. always easy to miss the default tab. Maybe it's time to force the first tab being selected at runtime...

mantidimaging/gui/ui/recon_window.ui
dtasev: We can enforce a height/width in the code but the .ui files will keep changing anyway in diffs. 

Not sure whether that will end up being more confusing when someone new comes in and changes the UI file and then doesn't see it actually change

mantidimaging/gui/ui/recon_window.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1495</width>
-    <height>831</height>
+    <width>1411</width>
```
dtasev, line 9:
```suggestion
    <width>1495</width>
```

mantidimaging/gui/ui/recon_window.ui
```diff
@@ -6,8 +6,8 @@
...
+    <height>813</height>
```
dtasev, line 10:
```suggestion
    <height>831</height>
```

Pasarus (approved):

PR #694, opened 2020-11-17
Add histograms to image previews, adds button to collapse left side of op window

### LINKED ISSUE #675: Operations window - add histogram to the image previews
Add histogram like in the Refine window for each image

Do we have the screen real estate to fit it?

dtasev:
Adds contrast-adjustable histograms to the image previews.

Additional:
- Adds a collapse button `<<` next to the collapsible section
- Changes the height of the histograms with windows resize. (rather caps the height of the image previews to a max of 80% of the desktop's height)

To Test:
- Load some data and go to the operations window
- Check that the contrast adjustment works
- Resize the window and check that the bottom histograms can be seen properly 
  - We might want to consider capping the min height of either the window or the image previews, so that on small screen the histogram is still visible
- Test out that the collapse works properly
  - Note if you manually collapse the button doesn't change as I'm not sure there's a Qt event for that
  - If you click the button it should change anyway, then click it again to open the section

Fixes #675

### DIFF
--- a/mantidimaging/gui/ui/filters_window.ui
+++ b/mantidimaging/gui/ui/filters_window.ui
@@ -315,10 +315,16 @@
       </widget>
       <widget class="QWidget" name="layoutWidget_2">
        <layout class="QVBoxLayout" name="previewsPaneLayout">
+        <property name="topMargin">
+         <number>2</number>
+        </property>
         <item>
          <layout class="QHBoxLayout" name="previewsPaneToolbar">
           <property name="sizeConstraint">
            <enum>QLayout::SetMaximumSize</enum>
+          </property>
+          <property name="topMargin">
+           <number>3</number>
           </property>
           <item>
            <widget class="QPushButton" name="collapseToggleButton">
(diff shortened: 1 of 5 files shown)


Pasarus (changes requested):
image
2 things:

By Default could the window open larger by setting height and width for the window? or could we make the histograms optional, and not present by default (I realize this might be a pain)

Could we add a small top margin/buffer for the collapse button from the top of the window?

dtasev:
> By Default could the window open larger by setting height and width for the window? 
I will do that - it seems the height is capped to the height of the screen (at least on IDAaaS) so even if it's too big for the user's screen it will just take 100% of the height, which is fine

> could we make the histograms optional, and not present by default (I realize this might be a pain)
Possibly, but I'd rather not if we can avoid it!

> Could we add a small top margin/buffer for the collapse button from the top of the window?
👍

dtasev:
Added a small margin, now looks like:
image

dtasev:
Ready for another look, I've made the window bigger and added the margin

Pasarus (approved):

PR #695, opened 2020-11-17
Fixes inverting the overlay difference not showing properly

### LINKED ISSUE #644: Overlay doesn't show properly when diff image is inverted
It just makes the whole thing red - it should be calculated before negation

dtasev:
To test:
- Load image
- Go to operations
- Use flat fielding and invert the difference checkbox option
- Check that the overlay difference doesn't change, but the difference image inverts as expected

Fixes #644


Pasarus (approved):
The overlay difference

PR #697, opened 2020-11-18
Recon Window: Cor Table automatically does fit

### LINKED ISSUE #659: Recon window - cor table always automatically do fit?
Remove "Calculate COR/Tilt from slice COR table" and just do it automatically

Pasarus:
Fixes #659 

To test:
- Load Data
- Clean the data (ROI Norm, Crop, Remove Outliers and Flat fielding)
- Open Reconstruction window
- Ensure each time the table changes that the fit is performed


mantidimaging/gui/windows/recon/presenter.py
```diff
@@ -217,7 +217,10 @@ def do_cor_fit(self):
         self.model.do_fit()
         self.view.set_results(*self.model.get_results())
         self.do_update_projection()
-        self.do_preview_reconstruct_slice()
+        try:
+            self.do_preview_reconstruct_slice()
+        except Exception:
+            pass
```
dtasev, line 223: Can this be a specific exception? Not quite clear what we're trying to capture

mantidimaging/gui/windows/recon/view.py
```diff
@@ -112,7 +111,10 @@ def on_row_change(item, _):
                 cor = row_data.cor
                 self.presenter.set_row(item.row())
                 self.presenter.set_last_cor(cor)
-                self.presenter.set_preview_slice_idx(slice_idx)
+                try:
+                    self.presenter.set_preview_slice_idx(slice_idx)
+                except Exception:
+                    pass
```
dtasev, line 117: Same - can we narrow down the exception, perhaps add a comment about why it's capturing it too

dtasev (changes requested):

mantidimaging/gui/windows/recon/presenter.py
Pasarus: Astra literally raises type Exception, here so unfortunately not.
```  File "/home/sam/miniconda3/envs/mantidimaging/lib/python3.7/site-packages/astra/projector.py", line 36, in create
    return p.create(config)
  File "astra/projector_c.pyx", line 67, in astra.projector_c.create
Exception: Unable to initialize Projector2D.```

Comment (Pasarus) [reply mantidimaging/gui/windows/recon/presenter.py:223]: I will add a comment to explain this

File: mantidimaging/gui/windows/recon/view.py (hunk 2 above)
Comment (Pasarus) [reply mantidimaging/gui/windows/recon/view.py:117]: Moved one layer lower in the stack, as this is not the best place to catch the exception. Same as above with astra throwing the exception as "Exception" will add the same comment.

File: mantidimaging/gui/windows/recon/presenter.py (hunk 1 above)
Comment (dtasev) [reply mantidimaging/gui/windows/recon/presenter.py:230]: That's annoying but not much we can do

File: mantidimaging/gui/windows/recon/presenter.py [hunk 3]
```diff
@@ -126,7 +126,12 @@ def set_preview_projection_idx(self, idx):
     def set_preview_slice_idx(self, idx):
         self.model.preview_slice_idx = idx
         self.do_update_projection()
-        self.do_preview_reconstruct_slice()
+        # Astra raises type exception during auto fit here, attempting to fit without data to fit to is a problem, it is
+        # not easy to force the method to not be called given no table data.
+        try:
+            self.do_preview_reconstruct_slice()
+        except Exception:
+            pass
```
dtasev, line 134: Actually I think it's a better idea to handle this inside `do_preview_reconstruct_slice` itself - and show an error message to the user to check their COR/Tilt values.

Otherwise this exception can be seen raised from the `notify` itself with the steps from the main comments above

dtasev (changes requested):
I found a way to break this a bit, but I'm not sure exactly why it happens, also doesn't seem to happen on master - perhaps it's because it's busy fitting so something gets corrupted?

To break:
- Load some data
- Go to recon window and click add a few times fast
- It will add a COR, once or twice and then start adding nan's
- At this point the COR/tilt are at the max values and ASTRA projector can't initialise properly anymore.
  - This is recoverable - you just need to correlate a new COR and it will be fine, but we should avoid getting into this state anyway

I think we may need to disable the `Add` button so it can't be spammed until the fit has completed. This is probably good to have otherwise the fit data model is getting changed while it's doing the fit

mantidimaging/gui/windows/recon/presenter.py
Pasarus: Ok sounds like a plan!

dtasev (changes requested):
So I've tracked it down to the fit itself, and it's not an issue caused by changes in this PR, but it is exposed with the auto fit.

When click Add multiple times without changing the slice index we provide multiple CORs for the same point, to which the fit just gives back NaNs, which makes sense as the data doesn't make any sense.

I think this is essentially a 'data entry' issue so the least effort fix: disable the Add button until the slice index is changed. We might also want to add some label to tell the user that they need to change the slice index 

Patch:

```python
diff --git a/mantidimaging/gui/windows/recon/presenter.py b/mantidimaging/gui/windows/recon/presenter.py
index 502d287a..9698e341 100644
--- a/mantidimaging/gui/windows/recon/presenter.py
+++ b/mantidimaging/gui/windows/recon/presenter.py
@@ -128,6 +128,8 @@ class ReconstructWindowPresenter(BasePresenter):
         self.do_update_projection()
 
+        self.view.addBtn.setDisabled(False)
+
     def set_row(self, row):
         self.model.selected_row = row
 
@@ -144,6 +146,8 @@ class ReconstructWindowPresenter(BasePresenter):
             self.view.update_projection(img_data, self.model.preview_slice_idx, self.model.tilt_angle)
 
     def do_add_cor(self):
:...skipping...
diff --git a/mantidimaging/gui/windows/recon/presenter.py b/mantidimaging/gui/windows/recon/presenter.py
index 502d287a..9698e341 100644
--- a/mantidimaging/gui/windows/recon/presenter.py
+++ b/mantidimaging/gui/windows/recon/presenter.py
@@ -128,6 +128,8 @@ class ReconstructWindowPresenter(BasePresenter):
         self.do_update_projection()
 
+        self.view.addBtn.setDisabled(False)
+
     def set_row(self, row):
         self.model.selected_row = row
 
@@ -144,6 +146,8 @@ class ReconstructWindowPresenter(BasePresenter):
             self.view.update_projection(img_data, self.model.preview_slice_idx, self.model.tilt_angle)
 
     def do_add_cor(self):
+        self.view.addBtn.setDisabled(True)
+
         row = self.model.selected_row
         cor = self.model.get_me_a_cor()
         self.view.add_cor_table_row(row, self.model.preview_slice_idx, cor.value)
diff --git a/mantidimaging/gui/windows/recon/view.py b/mantidimaging/gui/windows/recon/view.py
index 8d619891..793a2b11 100644
--- a/mantidimaging/gui/windows/recon/view.py
+++ b/mantidimaging/gui/windows/recon/view.py
@@ -229,9 +229,7 @@ class ReconstructWindowView(BaseMainWindowView):
         self.clearAllBtn.setEnabled(not empty)
 
         if self.tableView.model().num_points >= 2:
-            self.addBtn.setDisabled(True)
             self.presenter.notify(PresN.COR_FIT)
-            self.addBtn.setDisabled(False)
 
     def add_cor_table_row(self, row: Optional[int], slice_index: int, cor: float):
         """
```

mantidimaging/gui/windows/recon/view.py
```diff
@@ -81,6 +80,7 @@ def __init__(self, main_window: 'MainWindowView'):
         self.cor_table_model.rowsInserted.connect(self.on_table_row_count_change)  # type: ignore
         self.cor_table_model.rowsRemoved.connect(self.on_table_row_count_change)  # type: ignore
         self.cor_table_model.modelReset.connect(self.on_table_row_count_change)  # type: ignore
+        self.cor_table_model.dataChanged.connect(self.on_table_row_count_change)
```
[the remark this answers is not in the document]
Pasarus: I have now ensured that on_data_change performs the fit as since these rows can change other than when a row is added, the fit needs to be ran again to ensure it is current with the data the user has entered.

Pasarus:
I am not sure we should disable the functionality of the Add button without changing the slice index, would it be best to discuss this in person?

dtasev:
Pushed a change to the YAPF formatting - will merge when the builds pass

dtasev (approved):

PR #698, opened 2020-11-18
Add 180 degree projection protection in operations window

### LINKED ISSUE #672: Signify that proj180 shouldn't have any stand-alone operations performed on it
All operations should match with the main projections otherwise they will get out of sync and the Correlation COR might mismatch.

We need a way to signify to the user that they shouldn't apply filters on it.

Suggested fixes:
- Perhaps we should forbid the application of filters to the proj180 stack to prevent this completely?

Pasarus:
Fixes #672 

To test:
- Load data with a 180 degree projection
- Open operations window
- Perform an operation on tomo stack ensure 180 degree also has it applied
- Perform an operation on all stacks and ensure the 180 degree has it applied
- Perform an operation on only the 180 degree stack and select cancel and ensure it didn't apply
- Perform an operation on only the 180 degree stack and select ok and ensure it did apply


mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -196,8 +197,22 @@ def _post_filter(self, updated_stacks, task):
...
+        for stack in apply_to:
+            is_180_proj = self.is_a_proj180deg(stack)
+            if not is_180_proj:
+                confirmed_stacks.append(stack)
+                continue
+            if is_180_proj and allow_180_degree:
+                confirmed_stacks.append(stack)
+            else:
+                if is_180_proj and confirm_with_user_for_180degree and\
+                        self.view.ask_confirmation("Are you sure you want to apply to 180 degree projection?"):
```
dtasev, line 211:
```suggestion
                        self.view.ask_confirmation("Operations applied to the sample are also automatically applied to the 180 degree projection. Please avoid applying an operation unless you're absolutely certain you need to.\nAre you sure you want to apply to 180 degree projection?"):
```

We need to find a way to signify that the users shouldn't be operating on the 180 degree projection (unless there's a reason to, and I'm not sure what that reason might be)

Maybe a better warning message would be enough?

dtasev (changes requested):

mantidimaging/gui/windows/operations/presenter.py
Pasarus: Sounds like a plan!

mantidimaging/gui/windows/operations/test/test_presenter.py
```diff
@@ -181,3 +182,62 @@ def test_original_stack_assigned_when_safe_apply_checked(self, _):
...
+                               allow_180_degree=allow_180_degree,
+                               confirm_with_user_for_180degree=confirm_with_user_for_180degree)
+
+    if not allow_180_degree:
+        if confirm_with_user_for_180degree:
+            model.do_apply_filter.assert_called_once_with([images], "Partial")
+            partial_mock.assert_called_once_with(post_filter, [images])
+        else:
+            model.do_apply_filter.assert_not_called()
+            partial_mock.assert_not_called()
```
dtasev, line 239: I think with the change above you can also add a check if `self.do_update_previews` was called - I think this might do it. Not sure if this will need some more mocking out though
```suggestion
            partial_mock.assert_not_called()
            self.view.clear_previews.assert_called_once()
```

dtasev (changes requested):

dtasev:
Functionally works great, just some minor comments

mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -196,8 +197,27 @@ def _post_filter(self, updated_stacks, task):
...
+                continue
+            if is_180_proj and allow_180_degree:
+                confirmed_stacks.append(stack)
+            else:
+                if is_180_proj and confirm_with_user_for_180degree and\
+                        self.view.ask_confirmation("Operations applied to the sample are also automatically applied to "
+                                                   "the 180 degree projection. Please avoid applying an operation "
+                                                   "unless you're absolutely certain you need to.\nAre you sure you "
+                                                   "want to apply to 180 degree projection?"):
+                    confirmed_stacks.append(stack)
```
dtasev, line 215: This bit and the similar part of code in `_post_filter` has become very hard to understand, can it be refactored into something more straightforward? On a glance it seems like we should be able to branch out based on whether the current stack is a 180deg projection one or not, but I haven't looked too much at it

dtasev (changes requested):
It's ignoring the Safe apply now when applied to a proj180 stack. And from looking at the code I'm not even sure why - there's a lot of conditions floating around, we need to simplify it

mantidimaging/gui/windows/operations/presenter.py
dtasev: Also this whole asking code should be in `do_apply_filter` to avoid cluttering this function that should be remain simple and just do the running.

I think this can simplify it to essentially being `if is_180_proj and self.view.ask_confirmation(...)` in `do_apply_filter`, so if the user clicks OK it just runs the stack through the rest

Pasarus: The code is in _do_apply_filter() due to the fact it is called in 3 places and I do not want to repeat the code. It makes sense to centralize it here, I will attempt to simplify the code as I agree it was not as clear as it could have been.

dtasev: I think we only care to show this message if the user is trying to apply directly on 1 stack, and that stack happens to be proj 180 degree. The only entry point for this case is `do_apply_filter` if I'm not mistaken

mantidimaging/gui/windows/operations/presenter.py
```diff
@@ -196,8 +193,25 @@ def _post_filter(self, updated_stacks, task):
...
+            if not is_180_proj:
+                confirmed_stacks.append(stack)
+            elif allow_180_degree:
+                confirmed_stacks.append(stack)
+            elif confirm_with_user_for_180degree and\
+                self.view.ask_confirmation("Operations applied to the sample are also automatically applied to the 180 "
+                                           "degree projection. Please avoid applying an operation unless you're "
+                                           "absolutely certain you need to.\nAre you sure you want to apply to 180 "
+                                           "degree projection?"):
+                confirmed_stacks.append(stack)
```
dtasev, line 209: I'm still not convinced we need this here when the only thing that uses it is `do_apply_filter`. 

All other paths ignore the choices via the function parameters:
- Do apply on 180 proj comes with allow 180 and no confirm
- DO apply all comes with don't allow and don't confirm

So it seems this is better placed inside `do_apply_filter` itself, rather than cluttering this function

dtasev (changes requested):
Also needs a rebase as I changed the call that processes the 180deg in https://github.com/mantidproject/mantidimaging/pull/715 to `_do_apply_filter_sync`

dtasev (approved):

PR #699, opened 2020-11-18
Check package label and switch between unstable and main icons/title

### LINKED ISSUE #638: Use proper icon for Release and Unstable
https://github.com/mantidproject/mantidimaging/pull/637 adds the icon files for both - but only uses the main one.

Handling which one to use is a bit more tricky:
- First the package needs to know which one it is, this could be done in the version_check.py parsing perhaps?
- If not unstable: do nothing
- If unstable:
  - Needs to change the app name to Mantid Imaging Unstable
  - Needs to change the app icon to the unstable image file
  - Needs to suggest the correct installation command on IDAaaS when unstable! (install script needs REPO_LABEL=unstable parameter!!)

dtasev:
Improves the package version parsing to also check the label that was used. From this determines whether the package is "main" or "unstable".

If unstable it will:
- Change the icon and window title to contain Unstable, and the warning message shown to the user
- Adjusts the update command to update the `unstable` package in the `mantidimaging_unstable` IDAaaS environment

This doesn't affect IDAaaS as the icon is set manually.

To test:

### Unstable
- Install an unstable package with: `conda create -c mantid/label/unstable -n mantidimaging_unstable mantidimaging-1.1.0_1167`
  - You'll also need to `pip install PyQt5==5.15` or `mantidimaging` won't open at all
  - This is not the latest as of writing the PR so it should work fine, but check https://anaconda.org/mantid/mantidimaging/files if not working for some reason
- Upon opening you should get the message that it's not the latest, and a command to update that has the command `ENVIRONMENT_NAME=mantidimaging_unstable REPO_LABEL=unstable source ...`
- Check that the Unstable icon is used and the title is Mantid Imaging Unstable

### Main
- Install a main package with: `conda create -n mantidimaging_main mantidimaging-1.1.0_1096` (note this is an older `main` package version
  - I labelled the same version as both `unstable` and main for testing, but we'll surpass both so it doesn't really matter
- Upon opening you should get a message that the latest one is `1.1.0_1167`
- The command in the terminal shouldn't have any variables in front and just be `source ...`
- Check that the icon is the normal one and the title is just "Mantid Imaging"

Fixes #638


Pasarus (changes requested):
Do we cache the version of the icon? I am just considering the possibility for offline usage of the software.

dtasev:
The icon file is in the package, but nothing is cached.

As we have the local package label information anyway, I've moved the check to be before that, and refactored the logic to check for "is_main" first rather than "is_unstable" (basically flipped all the bool checks)

PR #700, opened 2020-11-18
Add toggle average functionality to ROI selector

### LINKED ISSUE #678: Select ROI - add back toggle show averaged image
This functionality is present on the main window, can we add it on the Select ROI view as well?

Pasarus:
Fixes #678 

To test:
- Load data
- Open operations window
- Select ROI
- Right click on the stack
- Select toggle average
- Toggle it back and forth a few times and ensure the ROI works correctly


mantidimaging/gui/windows/operations/view.py
```diff
@@ -231,6 +233,24 @@ def roi_visualiser(self, roi_field):
         window.setCentralWidget(self.roi_view)
         self.roi_view.setWindowTitle("Select ROI for operation")
 
+        def toggle_average_images(images_):
```
dtasev, line 236: Can we move this into `windows/roi_selector`? It's getting pretty big to be a function in the Operations window

dtasev (changes requested):

mantidimaging/gui/windows/operations/view.py
Pasarus: I think if we need to add anything else to the ROI selector, then sure, but right now I was trying to keep it as a "Dialog" similar to QMessageBox

dtasev (approved):

PR #726, opened 2020-11-25
Adds Lock Histogram button and functionality

### LINKED ISSUE #705: Compare window - option to lock histograms to each other
This is done in the Refine window right now

dtasev:
- Refactors the common lock function into a base presenter mixin
- Adds tests

Not 100% sure about the choice of button here - currently it's a toggleable button, but maybe we should have a checkbox. This could be argued is more consistent with the Operations presenter, but that doesn't have an option to lock histogram ranges. I don't really mind either way.

To test:
- Load some data
- There's 2 ways to get to this dialog now:
	- Apply a filter with Safe Apply, try the Lock Histograms there
	- Go to Workflow > Compare, try the lock histograms there too

Fixes #705


Pasarus:
I think we should use a check box, and would it be possible to lock the zoom levels of the histograms?

Pasarus:
When you click lock histogram could it sync both of the histogram's data immediately? So for example say the ranges are mismatched when you toggle it on, then set them to the same value, with the left/original histogram being the "main" histogram that sets the range? maybe just the range that is largest.

dtasev:
I think that should do it

Pasarus (approved):
Great new feature, great new tests :+1:

PR #733, opened 2020-11-30
Adds source file path to IMATLogFile, shows it in metadata

### LINKED ISSUE #702: Add ability to see if log is loaded somewhere in the GUI
Could be one of:
- Entry in the history dialog saying: `Logs loaded: Yes`
  - Can also have a `view` link/button
  - Or just contain the path to the log file
- `View Log` context menu in the `StackVisualiser`

dtasev:
Adds source file path to IMATLogFile, shows the file path in the metadata.

Also: 
- Adds separators in StackVisualiser context menu (main window and select ROI)
- Re-orders the main StackVisualiser context menu

Fixes #702


Pasarus (changes requested):
If I attempt to load a log, that has a different number of projections in it to the images loaded, it adds a log_file section to the history/metadata window, but the log is not used. Fix that and this should be ready to go!

dtasev:
Latest changes should fix that

dtasev:
Coverage should be fine now 👍 Added tests for the new branches I introduced in https://github.com/mantidproject/mantidimaging/pull/733/commits/dd46fc62b6a3f5adce4a340a67d19f6efb89b255

Pasarus (approved):

PR #735, opened 2020-12-01
625: Fix GPU tests

### LINKED ISSUE #625: Fix failing GPU/Median filter tests
A few of them are using a numpy array rather than an `Images` object.

DolicaAkelloEgwel:
Closes #625


Pasarus (changes requested):
Would it be possible to just add cupy to the test environments, docker, and the conda build? This way the GPU tests will run, without having to install cupy manually.

dtasev:
The conda build would be the easy one to start with. I don't know if we're allowed to access the GPUs on the build servers, so best to start with the conda build one.

Rebuilding the docker images might be a bit of a PITA, and the `docker run` command might need to share the GPUs. Let me know if you struggle with that one.

DolicaAkelloEgwel:
I have tried a variety of ways to add cupy to the yaml file and all of them seem to get the libcuda error. The last time I encountered this error I think it had something to do with the cuda installation path. Paging @dtasev @Pasarus...

DolicaAkelloEgwel:
I've removed the changes to the environment.yml file for the time being.

PR #739, opened 2020-12-02
Move python=3.7 specification to the conda build

dtasev:


dtasev:
Not needed as it was trying to workaround a temporary anaconda bug or outage

PR #740, opened 2020-12-03
Adds user manual in RST, converted from Arianna's LaTeX

### LINKED ISSUE #664: User manual links from GUIs and revamping of the User Manual
Will need user manuals for Operations and Reconstruction

And:
- Delete these pages https://mantidproject.github.io/mantidimaging/user_guide/gui/index.html as they are out of date

dtasev:
To test: 
Read through the new content in `user_guide/index.rst`

The `docs/developer_guide/release.rst` will be updated further with #741, once we actually do those steps!

Fixes #664


docs/user_guide/index.rst
```diff
@@ -7,6 +7,101 @@ User Guide
...
+1. First select a slice index close to the top of the sample using the Preview box at the bottom of the reconstruction window.
+2. Then press the "Add" button at the bottom of the screen. This should add your slice to the table.
+3. Press the "Refine" button whilst selecting this slice. This brings up the following window:
+
+
+.. image:: ../_static/refine_window.png
+   :alt: Loading screen
+   :align: center
+
+4. Continue by selecting which window shows the most accurate depiction of the sample. Mantid image will highlight the window it feels is best in green.
```
Pasarus, line 94:
```suggestion
4. Continue by selecting which window shows the most accurate depiction of the sample. Mantid Imaging will highlight the window it feels is best in green.
```

docs/user_guide/index.rst
```diff
@@ -7,6 +7,101 @@ User Guide
...
+6. After this the "Calculate COR/Tilt from slice COR table" will now be selectable. Feel free to repeat the process for intermediate slices before pressing this button.
+
+Now move to the "Reconstruct" tab. At this stage there should several different settings to use. For this sample use the following settings.
+
+- Maximum Projection Angle: 360.00
+- Algorithm: FBP-CUDA
+- Reconstruction filter: hann
+- Pixel size (microns): 103
+
+There are many filter options. Experiment with the filters by looking at the slice preview and the corresponding hitogram next to it. Filters like "hann" will strongly filter out higher frequency components. In contrast "ram-lak" preservers these higher frequency components resulting in a visibly less smooth histogram.
```
Pasarus, line 105:
```suggestion
There are many filter options. Experiment with the filters by looking at the slice preview and the corresponding histogram next to it. Filters like "hann" will strongly filter out higher frequency components. In contrast "ram-lak" preserves these higher frequency components resulting in a visibly less smooth histogram.
```

Pasarus (changes requested):
Looking really really good right now, just a few questions.

docs/user_guide/index.rst
```diff
@@ -7,6 +7,101 @@ User Guide
...
+    :alt: View of the application window
+    :align: center
+
+
+Operations
+##########
+
+Next let's take the sample we loaded and let's tidy it up with operations. To open the operations go to "Workflow" then "Operations".
+
+1. **Remove Outliers** will be the first operation we will apply. This has to be completed in 2 modes "Bright" and "Dark". We'll complete this for both modes with difference set to 1000 and median kernel set to size 3. Apply this to all stacks.
```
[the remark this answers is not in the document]
dtasev: Yeah during that discussion we found out that they use ImageJ's 50, which is equivalent to 50 for us. I checked that afterwards and the confusion seems to be because we show all pixels affected in a binary mask, while ImageJ doesn't use a binary mask, so you can only see the brightest affected pixels.

Really it's very based on the data, and the Tilt 4 is already pre-cleaned in ImageJ with 50, so with 1000 there might not be a lot left to change

Pasarus (approved):

docs/user_guide/index.rst
```diff
@@ -7,6 +7,115 @@ User Guide
...
+
+2. Then press the "Add" button at the bottom of the screen. This should add your slice to the table.
+3. Press the "Refine" button whilst selecting this slice. This brings up the following window:
+
+
+.. image:: ../_static/refine_window.png
+   :alt: Loading screen
+   :align: center
+
+4. Continue by selecting which image shows the most accurate depiction of the sample. Mantid Imaging will highlight the image it feels is best in green. This can be innacurate for very noisy samples, but should be accurate for this dataset.
```
samtygier-stfc, line 102: "innacurate" -> "inaccurate"

samtygier-stfc (changes requested):

samtygier-stfc (approved):

dtasev:
@samtygier-stfc Feel free to merge this if you think it's OK

Issue #500, opened 2020-08-10
Plot monitor counts against the integrated counts for that image


dtasev:
> It should be a y=x line to show good correlation, any spikes away from that would give you a  good indication you have a rubbish projection.

Could be another line on the ROI histogram? Might be enough

This is a new feature rather than part of Octopus

PR #744, opened 2020-12-03
Duplicate context menu in the a new main window menu

### LINKED ISSUE #676: Move right click context menu to a window menu?
The right click on the image is getting pretty confusing - there's tons of options and #674 is going to be adding another one.

Perhaps we should move this into an "Image" submenu in the main menu?

We could duplicate the actions there and connect to the same place - essentially just making a more obvious way of accessing them

Pasarus:
Fixes #676 

2 Things have happened in this PR:
- I have disabled most main window options until a stack is loaded, if no stacks are present they should be disabled, and re-enabled later on.
- The context menu for a loaded stack, should also be available in a menu called Image at the top of the Main window

To Test:
- Open imaging, ensure that Image, Workflow, File->Save, File->LoadSampleLog, File->Load180DegreeProjection are disabled.
- Load a stack
- Ensure that all of the disabled items are now enabled.
- Check that the Image menu only applies operations/actions to the currently viewed stack.


mantidimaging/gui/windows/main/view.py
```diff
@@ -34,14 +36,19 @@ class MainWindowView(BaseMainWindowView):
...
+    menuWorkflow: QMenu
+    menuImage: QMenu
+    menuHelp: QMenu
 
     actionRecon: QAction
     actionFilters: QAction
     actionSavuFilters: QAction
     actionCompareImages: QAction
-    actionLoadLog: QAction
+    actionSampleLoadLog: QAction
```
dtasev, line 51: This can load logs for other things.. we don't use it but there's also logs for the flat images for #500

mantidimaging/gui/windows/stack_visualiser/view.py
```diff
@@ -121,26 +144,8 @@ def roi_changed_callback(self, roi: SensibleROI):
...
-
-        for (menu_text, func) in actions:
-            if func is None:
-                menu.addSeparator()
-            else:
-                action = QAction(menu_text, menu)
-                action.triggered.connect(func)
-                menu.addAction(action)
-
+        populate_menu(menu, self.actions)
```
dtasev, line 148: 👍

dtasev (changes requested):
Looks good, just a comment on changing the `actionLoadLog` name

mantidimaging/gui/utility/qt_helpers.py
```diff
@@ -13,7 +13,7 @@
 from PyQt5 import uic  # type: ignore
 from PyQt5.QtCore import QObject
 from PyQt5.QtWidgets import QLabel, QLineEdit, QPushButton, QSpinBox, QDoubleSpinBox, QCheckBox, QComboBox, QWidget, \
-    QSizePolicy
+    QSizePolicy, QAction
```
dtasev, line 16:
```suggestion
    QSizePolicy, QAction, QMenu
```

mantidimaging/gui/utility/qt_helpers.py
```diff
@@ -235,3 +235,13 @@ def delete_all_widgets_from_layout(lo):
         # layout and marks them for deletion)
         elif item.widget() is not None:
             item.widget().setParent(None)
+
+
+def populate_menu(menu, actions_list):
```
dtasev, line 240:
```suggestion
def populate_menu(menu: QMenu, actions_list: List[QAction]):
```
This should work if you commit both, but perhaps best to do it locally

dtasev (changes requested):
Small type hint addition

mantidimaging/gui/windows/main/view.py
Pasarus: I realize that, but that is the actual name of the action, and actionLoadLog points to nothing in the UI file.

dtasev: Ah I see

PR #750, opened 2020-12-03
Coverage Config additions

Pasarus:
Code review and coverage check

Coverage will be lower by around 10% from this PR but we will only be checking production code now, so the coverage actually means way more.


dtasev (approved):

dtasev:
RIP good coverage on `master`

PR #753, opened 2020-12-03
Welcome screen implementation

### LINKED ISSUE #671: User welcome screen
Should show links to docs, manuals

Features:
- "Don't show this again" option that is cached locally
- About > Show welcome screen
- Link to docs and manuals

samtygier-stfc:
Fixes #671

- [x] Shows at startup or from Help->About
- [x] Show version
- [x] Checkbox to disable show are start
- [x] Re-show for new version
- [x] Useful links
- [x] Branding
- [x] Avoid displaying during tests


samtygier-stfc:
image
image
 
I think `version_check.py` could be refactored, to avoid multiple conda and web calls. And make getting the current version easier.

samtygier-stfc:
This now does everything the old dialog did, plus the new features.

New links can be added to the list in mantidimaging/gui/windows/welcome_screen/presenter.py

New issue checks (e.g. missing conda, insufficient free memory) could be added to check_issues()

mantidimaging/core/utility/version_check.py
```diff
@@ -5,92 +5,153 @@
...
-                 "source /opt/miniconda/bin/activate /opt/miniconda && "
-                 f"{command_prefix}source "
-                 "<(curl -s https://raw.githubusercontent.com/mantidproject/mantidimaging/master/install.sh)")
-        action(msg)
-    else:
-        LOG.info("Running the latest Mantid Imaging")
-
-
 def _make_version_str(parsed: ParsedVersion) -> str:
     return f"{'.'.join([str(v) for v in parsed.version])}_{parsed.commits}"
```
dtasev, line 150: I don't think this function is used anymore (only in a test, but we can remove both)

mantidimaging/core/utility/version_check.py
```diff
@@ -5,92 +5,153 @@
...
+                                    _parse_version(self.get_conda_available_version()))
+
+    def conda_update_message(self) -> tuple:
+        suffix = self.get_conda_installed_label()
+
+        msg = f"Not running the latest Mantid Imaging{suffix}.\n"
+        msg += f"Found version {self.get_conda_installed_version()}, "
+        msg += f"latest: {self.get_conda_available_version()}.\nPlease check the terminal for an update command!"
+
+        detailed = f"Not running the latest Mantid Imaging{suffix}.\n"
```
dtasev, line 95:
```suggestion
        detailed = f"Not running the latest Mantid Imaging-{suffix}.\n"
```
I think this needs some separator before the suffix as it looks like this without one:

image

It may be worth not appending "main" at all, but only add "unstable" for the unstable version

dtasev (changes requested):
Looks really good - I love the purple recolor!

Functionally works, just a couple of small comments

mantidimaging/core/utility/version_check.py
```diff
@@ -5,92 +5,153 @@
...
+
+        return _version_is_uptodate(_parse_version(self.get_conda_installed_version()),
+                                    _parse_version(self.get_conda_available_version()))
+
+    def conda_update_message(self) -> tuple:
+        suffix = self.get_conda_installed_label()
+
+        msg = f"Not running the latest Mantid Imaging{suffix}.\n"
+        msg += f"Found version {self.get_conda_installed_version()}, "
+        msg += f"latest: {self.get_conda_available_version()}.\nPlease check the terminal for an update command!"
```
dtasev, line 93: If we have the user's attention here already, maybe we can just move the update command here?

It would have to be in a QLineEdit though (or a text edit), so the user can easily copy it

dtasev (changes requested):

mantidimaging/core/utility/version_check.py
samtygier-stfc: Removed in b90e47e

mantidimaging/core/utility/version_check.py
samtygier-stfc: I think it depends on the approach to #755 . The welcome dialog could just link to the entry on the trouble shooting webpage.

dtasev: You are right. I'm happy to leave it where it is for now.

dtasev (approved):

Issue #747, opened 2020-12-03
Distinguish data allocations to avoid corrupting already running Mantid Imaging instances


dtasev:
We currently **do not distinguish** between data allocated from Mantid Imaging release package and unstable package. Therefore if "start anew" deletes all previous data, it will corrupt other Mantid Imaging instances.

The simplest fix is to add a prefix to `parallel.utils.create_array` depending on whether it's running a release or unstable. Then whenever any deletion of *all* is done (such as in #746 starting a new session) it will _ONLY_ delete data relevant to its package label

Issue #745, opened 2020-12-03
Change the `atexit` event to close only stacks handled by the current instance of Mantid Imaging


dtasev:
Problem:

The current `atexit` event will free _all_ stacks allocated via the `SharedArray`, regardless of who allocated it. Including other instances of Mantid Imaging.

Solution:
- Remove the `atexit` event, and in the MainWindow's close loop through all StackVisualisers and call `.close()`

PR #762, opened 2020-12-08
Reverts shared array creation to python multiprocessing

dtasev:
I had some time over the weekend to consider the best way to fix the issues on IDAaaS caused by SharedArray. Namely https://github.com/mantidproject/mantidimaging/issues/745 and https://github.com/mantidproject/mantidimaging/issues/747.

Then I remembered that I changed it from `multiprocessing` to `SharedArray` as I suspected that the former was leaking the memory, but this turned out to be wrong after the change [1]. It was just never reverted as it took a bit of time to implement in the first place (I definitely fell into the [sunk cost fallacy](https://www.behavioraleconomics.com/resources/mini-encyclopedia-of-be/sunk-cost-fallacy/) here, and hopefully I learned my lesson!).

But now it's time for it to go, as it's caused a lot of issues for it's supposed "benefits" that we may never use.

This PR reverts the shared array creation to implementation from https://github.com/mantidproject/mantidimaging/commit/df4b3e2a6983d299c9ce71ff53efeceb575d0c97#diff-6b06bb41e81d5a3b1ccac6da5b0f4098583cf306a3194550caa1a169fe1182b4 (look at `parallel/utility.py`)

The core of this change is in the first commit https://github.com/mantidproject/mantidimaging/commit/24d5b31388e8beb45d3835e8dfa070d638d6ecb4. The rest is just cleanup to `free_memory` and all the workarounds added for deleting arrays.

Interestingly, Python itself is adding a built-in module, much like `SharedArray`, called [`shared_memory`](https://docs.python.org/3/library/multiprocessing.shared_memory.html) which provides a similar functionality to `SharedArray`. Perhaps if we ever require the machine global named arrays in the future, that should be the way to go.

### To test

Before starting make sure to open `htop` or another RAM monitoring tool - you'll need to keep an eye on it.

#### Load and kill
- Load a stack
- Rebin it to dims 9999, 9999
	- If the stack is too big, this will fail with Not enough memory warning. Then just use `4096, 4096`, or skip and just duplicate data until you are close to 100% memory
- Once you get to 100%, open operations window, open recon window, and open and **close** a stack comparison window (from workflow -> compare)
- Close some of the stacks - memory usage should go down
- Copy it again or just kill the process from `htop` or similar
	- Note: this must be SIGKILL to prevent any `atexit`s running, although this PR also removes that

#### Safe apply
- Load a stack
- Keep an eye on used RAM post load
- Apply remove outliers (or anything else) with Safe Apply
- The RAM usage should double
- Choose the original - it should fall back to the post-load value
- Redo and choose the new stack, it should also fall back

[1] The real culprit was `ImageItem`, and it has bitten us again - if you [don't call imageItem.close()](https://github.com/mantidproject/mantidimaging/blob/master/mantidimaging/gui/windows/stack_choice/view.py#L164-L165) the memory gets leaked, even if you close the widget.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -25,7 +25,6 @@ requirements:
     - astropy
     - scipy
     - scikit-image=0.17.2
-    - sharedarray
     - numpy
     - tomopy=1.7.1=cuda*
     - cudatoolkit=10.2*
--- a/mantidimaging/core/data/images.py
+++ b/mantidimaging/core/data/images.py
@@ -67,25 +67,6 @@ class Images:
     def count(self) -> int:
         return len(self._filenames) if self._filenames else 0
 
-    def free_memory(self, delete_filename=True):
-        """
-        Delete the memory file containing the data, and the references to it within this class.
-
-        The memory will not be freed until _all_ references to it are gone, so local variables
-        can safely keep a reference even after deletion. This is used in unit testing data
-        generation, and ROI normalisation.
-
-        :param delete_filename: Whether to reset the memory filename attribute.
-                                Set this to False in cases where the data will be replaced with
-                                data with a new shape (rebin, crop), but the memory filename
-                                ought to remain the same.
-        """
-        if self.memory_filename is not None:
-            pu.delete_shared_array(self.memory_filename)
-            if delete_filename:
-                self.memory_filename = None
-        self.data = None
-
     @property
     def filenames(self) -> Optional[List[str]]:
         return self._filenames
--- a/mantidimaging/core/data/images.py
+++ b/mantidimaging/core/data/images.py
@@ -24,8 +24,7 @@ class Images:
                  filenames: Optional[List[str]] = None,
                  indices: Optional[Tuple[int, int, int]] = None,
                  metadata: Optional[Dict[str, Any]] = None,
-                 sinograms: bool = False,
-                 memory_filename: Optional[str] = None):
+                 sinograms: bool = False):
         """
 
         :param data: Images of the Sample/Projection data
@@ -42,7 +41,6 @@ class Images:
         self.metadata: Dict[str, Any] = deepcopy(metadata) if metadata else {}
         self._is_sinograms = sinograms
 
-        self.memory_filename = memory_filename
         self._proj180deg: Optional[Images] = None
         self._log_file: Optional[IMATLogFile] = None
         self._projection_angles: Optional[ProjectionAngles] = None
@@ -114,8 +112,7 @@ class Images:
 
     def copy(self, flip_axes=False) -> 'Images':
         shape = (self.data.shape[1], self.data.shape[0], self.data.shape[2]) if flip_axes else self.data.shape
-        data_name = pu.create_shared_name()
-        data_copy = pu.create_array(shape, self.data.dtype, data_name)
+        data_copy = pu.create_array(shape, self.data.dtype)
         if flip_axes:
             data_copy[:] = np.swapaxes(self.data, 0, 1)
         else:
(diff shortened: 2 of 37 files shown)


samtygier-stfc:
Looks good. Nice to remove something that is not giving an advantage. Tested above and it does free the memory in those cases.

Given that this is quite core, it maybe needs everyone to give it a quick test drive.

Is there some documentation for multiprocessing.heap? I could not see it on the python documentation site. Is it an internal API, is there something higher level in multiprocessing that we could use?

dtasev:
> Is there some documentation for multiprocessing.heap? I could not see it on the python documentation site. Is it an internal API, is there something higher level in multiprocessing that we could use?

I think you are right that it is an internal API. I looked into using `multiprocessing.Array` instead which is in the public API, and it's a small change [1], but it seems like it doesn't [free the memory properly until 3.8](https://bugs.python.org/issue32759). Currently the arrays never get freed after being closed.

I'm afraid I couldn't find anything else in 3.7, until 3.8 where we could use either `multiprocessing.Array` or https://docs.python.org/3.8/library/multiprocessing.shared_memory.html, which is kind of like `SharedArray` but built in


[1] 
```
diff --git a/mantidimaging/core/parallel/utility.py b/mantidimaging/core/parallel/utility.py
index ecfba673..5773236a 100644
--- a/mantidimaging/core/parallel/utility.py
+++ b/mantidimaging/core/parallel/utility.py
@@ -7,7 +7,7 @@ from contextlib import contextmanager
 from functools import partial
 from logging import getLogger
 # for some reason mypy can't find this import, nor can IDE suggestions
-from multiprocessing import heap  # type: ignore
+from multiprocessing import Array
 from multiprocessing.pool import Pool
 from typing import Any, Tuple, Type, Union
 
@@ -75,13 +75,8 @@ def _create_shared_array(shape, dtype: Union[str, np.dtype] = np.float32):
 
     LOG.info('Requested shared array with shape={}, length={}, size={}, ' 'dtype={}'.format(shape, length, size, dtype))
 
-    arena = heap.Arena(size)
-    mem = memoryview(arena.buffer)
-
-    array_type = ctype * length
-    array = array_type.from_buffer(mem)
-
-    data = np.frombuffer(array, dtype=dtype)
+    array = Array(ctype, length)
+    data = np.frombuffer(array.get_obj(), dtype=dtype)
 
     return data.reshape(shape)
 
```

samtygier-stfc:
> I'm afraid I couldn't find anything else in 3.7, until 3.8 where we could use either `multiprocessing.Array` or https://docs.python.org/3.8/library/multiprocessing.shared_memory.html, which is kind of like `SharedArray` but built in.
> 
> Maybe 3.8 is the version where this finally gets closed for good, but I think for this release we're stuck with the heap or `SharedArray`

Ok. So `multiprocessing.heap` is probably best for now. And we have a route to `multiprocessing.Array` in the near future.

samtygier-stfc:
Would it be useful to have a comment convention for marking things that are being kept for version compatibility, e.g.

`# COMPAT python 3.7 : Using heap instead of Array, see https://github.com/mantidproject/mantidimaging/pull/762`

Then whenever we bump up the minimum requirements, we can grep for "# COMPAT" and see what cleanups can be done.

samtygier-stfc (approved):

dtasev:
After a slack discussion with @Pasarus, in which we found that sometimes the memory can be leaked with these changes, we decided it's better to keep this PR out of the release, and instead add it as a first post-release change. This should give more time to track down any memory issues

DolicaAkelloEgwel (dismissed):
The tests worked fine for me.

dtasev:
Re-opening to merge in master now that the release branch is separated.

PR #768, opened 2020-12-08
736: CUDA not present error message

### LINKED ISSUE #736: Friendly error message if CUDA not present
If run on a system without CUDA (i.e. one without an NVidia GPU) then opening the reconstruction window with cause mantidimaging to crash.

To reproduce:
1) load a data set
2) Workflows -> Reconstruct

```
Error: CUDA error 35: CUDA driver version is insufficient for CUDA runtime version.
Error: Failed to allocate 2048x2048 GPU buffer
python: ./../../src/CudaReconstructionAlgorithm2D.cpp:293: virtual void astra::CCudaReconstructionAlgorithm2D::initCUDAAlgorithm(): Assertion `ok' failed.
Aborted
```

DolicaAkelloEgwel:
Closes #736 

Adds a check to see if CUDA is on the system. If the check fails, a message is displayed on the welcome screen and the CUDA algorithms cannot be selected from the recon window.


DolicaAkelloEgwel:
Just wanted to point out that I get this when combining the issues into a single string and calling `add_issue` once.
image

...But I get this when I call `add_issue` a second time.
image

mantidimaging/core/utility/cuda_check.py
```diff
@@ -0,0 +1,27 @@
...
+
+import os
+from typing import Tuple
+
+
+def _read_from_terminal(command: str) -> str:
+    """
+    Runs a terminal command and returns the result.
+    """
+    return os.popen(command).read()
```
dtasev, line 12: I think a better built-in to use is [subprocess.check_output](https://docs.python.org/3/library/subprocess.html#subprocess.check_output)

mantidimaging/core/utility/cuda_check.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def cuda_is_present() -> bool:
+    """
+    Checks if nvidia-smi is on the system + working, and that the libcuda file can be located.
+    """
+    return "Driver Version" in _read_from_terminal("nvidia-smi") and _read_from_terminal("locate libcuda.so") != ""
```
dtasev, line 19: I'm not sure that `_read_from_terminal("locate libcuda.so")` will be very accurate - `locate` seems to be returning `libcuda.so`s for me that have been installed in software (Avizo) or the manpages entry:

```
/home/dtasev/dev/containers/avizo/avizo/usr/local/cuda-10.0/compat/libcuda.so
/home/dtasev/dev/containers/avizo/avizo/usr/local/cuda-10.0/compat/libcuda.so.1
/home/dtasev/dev/containers/avizo/avizo/usr/local/cuda-10.0/compat/libcuda.so.410.104
/home/dtasev/dev/containers/mantidimaging/mi_savu.sif/usr/local/cuda-9.0/targets/x86_64-linux/lib/stubs/libcuda.so
/usr/lib/i386-linux-gnu/libcuda.so
/usr/lib/i386-linux-gnu/libcuda.so.1
/usr/lib/i386-linux-gnu/libcuda.so.440.33.01
/usr/lib/x86_64-linux-gnu/libcuda.so
/usr/lib/x86_64-linux-gnu/libcuda.so.1
/usr/lib/x86_64-linux-gnu/libcuda.so.440.33.01
/usr/local/cuda-10.2/doc/man/man7/libcuda.so.7
/usr/local/cuda-10.2/targets/x86_64-linux/lib/stubs/libcuda.so
/usr/share/man/man7/libcuda.so.7
```

Is there any way we can filter this to just be from the system libraries (note that they're not the same between distributions, `/usr/lib/x86_64-linux-gnu/` for x64 debian (ubuntu), and `/usr/lib64` for rhel/centos iirc. Best way would be if we can call `locate` with a flag that limits it to system dirs, but I'm not sure one exists.

Any other suggestions? @Pasarus @samtygier-stfc

dtasev (changes requested):
Code looks good, just a couple notes. Will test functionally a bit later today

samtygier-stfc:
Could it use the same checks as in mantidimaging/core/gpu/utility.py ? Trying to see if cuda will function?

dtasev (changes requested):

DolicaAkelloEgwel:
>Could it use the same checks as in mantidimaging/core/gpu/utility.py ? Trying to see if cuda will function?
@samtygier-stfc That only works for checking the cupy installation.

samtygier-stfc:
> > Could it use the same checks as in mantidimaging/core/gpu/utility.py ? Trying to see if cuda will function?
>  @samtygier-stfc That only works for checking the cupy installation.

My thought was that if cupy is working, then the whole CUDA stack must be working. If not we point them to the trouble shooting page that tells them how to install CUDA and cupy.

dtasev:
Unfortunately we're not guaranteed `cupy` will be there as it's not part of the conda package dependencies (iirc due to not being able to resolve the dependencies with it included, that may have changed)

Also for some reason my suggestion to change to `.*libcuda.so` isn't working when called from Python, but it's fine on the terminal. I'm not sure why that is happening.

While trying to figure out what changed, I also thought perhaps it would be nice to add which bit exactly failed (e.g. is it NVIDIA driver, or CUDA) in the logging

DolicaAkelloEgwel:
> My thought was that if cupy is working, then the whole CUDA stack must be working. If not we point them to the trouble shooting page that tells them how to install CUDA and cupy.
It's possible for CUDA working but for cupy to fail (like if there's a mismatch between cupy version and CUDA version) so I think that makes it a sufficient but not necessary condition...

mantidimaging/gui/windows/main/view.py
```diff
@@ -78,6 +79,8 @@ def __init__(self):
             self.setWindowTitle("Mantid Imaging Unstable")
             self.setWindowIcon(QIcon("./images/mantid_imaging_unstable_64px.png"))
 
+        self.use_cuda = cuda_is_present()
```
dtasev, line 82: I don't like this being here, but I also can't think of a better place as it's used in multiple windows. The model is possible, but then we'd just need a bunch of getters to reach the same thing..

Any ideas are welcome, if not I'm happy to approve with the current state

mantidimaging/gui/windows/welcome_screen/presenter.py
```diff
@@ -72,7 +75,13 @@ def show_at_start_changed(self):
...
         if not versions.is_conda_uptodate():
             msg, detailed = versions.conda_update_message()
-            self.view.add_issue(msg)
+            issues.append(msg)
+            LOG.info(detailed)
+        if not self.cuda_present:
+            msg, detailed = cuda_check.not_found_message()
+            issues.append(msg)
             LOG.info(detailed)
+        self.view.add_issues("\n".join(issues))
```
dtasev, line 87: This has made it so that I always see the issues box, even when it's empty

image

I think we need to add issues only `if len(issues) > 0`

dtasev (changes requested):
Now works for me on both Ubuntu and CentOS!

Just found a minor issue that the "issues" box is now always showing up, and a question

mantidimaging/gui/windows/main/view.py
samtygier-stfc: Might be cleaner if `ReconstructWindowModel` (and anywhere else that want to know) calls `cuda_is_present()` itself. `cuda_check.py` could store the answer, so it does not need to repeat the check.

samtygier-stfc (changes requested):
Fixes the issue. Check on VM without cuda and new laptop with.

I think calling the check from multiple places, is neater than passing a `use_cuda` flag around. But its not a show stopper, so I wouldn't be opposed to merging as is.

samtygier-stfc (approved):
Looks good. Tested on machines with and without CUDA.

samtygier-stfc:
@dtasev did you have any further comments?

PR #771, opened 2020-12-09
Adds troubleshoot page with initial entries

### LINKED ISSUE #755: Troubleshooting page
Add a docs page about common troubleshooting issues:
- CUDA
- Memory allocated from another mantid imaging

This should be linked to from the welcome screen

dtasev:
Adds troubleshooting for missing CUDA (check if GPU compatible, and suggest CUDA download).

Also:
- Removes Qt from the installation steps, the Qt lib that we use is shipped with PyQt5 via `pip install PyQt5`

Fixes #755

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -51,7 +51,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'MantidImaging'
-copyright = u'2017, Mantid Project'
+copyright = u'2017-2020, Mantid Project'
 author = u'Dimitar Tasev, Dan Nixon'
 
 # The version info for the project you're documenting, acts as replacement for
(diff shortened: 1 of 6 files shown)


samtygier-stfc:
In setting up the VM I found these solutions:

`SystemError: <built-in function connectSlotsByName> returned a result with an error set`
means that the pyqt packages are missing, so do
`pip install pyqt5==5.15 pyqtgraph==0.11`
(That command maybe needs adding to the dev guide page. I gave my conda dev env a different name, so did not get packages installed with the install.sh method)

`qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.`
Means that the Qt system libraries are missing. On ubuntu fix is
`apt install qtbase5-dev`

dtasev:
Added those in, as well as a missing display (that you see if starting it from a remote terminal, probably a rare case, but happens often to me)

dtasev:
@samtygier-stfc can you do a quick review of the new changes? I'd like to get this merged in

samtygier-stfc (approved):
Looks good

PR #773, opened 2020-12-09
Adds a unique prefix for all allocations of an instance

### LINKED ISSUE #745: Change the `atexit` event to close only stacks handled by the current instance of Mantid Imaging

### LINKED ISSUE #747: Distinguish data allocations to avoid corrupting already running Mantid Imaging instances

dtasev:
Only deletes the allocation with the instance prefix on shutdown. Will not touch any other arrays. Handling of hard crashes will be done via #774 

To test:
- Start up two mantid imagings
- Load data in each
- Close one
- The second one should still function as usual, try to applying an operation

Fixes #745

### DIFF
--- a/mantidimaging/core/parallel/test/utility_test.py
+++ b/mantidimaging/core/parallel/test/utility_test.py
@@ -3,9 +3,10 @@
 
 import mock
 import numpy as np
-
-from mantidimaging.core.parallel.utility import create_array, free_all_owned_by_this_instance, multiprocessing_necessary, execute_impl
 import SharedArray as sa
+
+from mantidimaging.core.parallel.utility import (create_array, execute_impl, free_all_owned_by_this_instance,
+                                                 multiprocessing_necessary)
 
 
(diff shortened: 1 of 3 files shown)


samtygier-stfc (approved):
Fixes multiple instances of mantidimaging interfering with each other. Looks good.

Issue #770, opened 2020-12-09
Move shared array allocation to `multiprocessing.Array` when on Python 3.8


dtasev:
Ref: https://github.com/mantidproject/mantidimaging/pull/762#issuecomment-741663482

PR #775, opened 2020-12-09
Now asks the user to delete data or ignore it

### LINKED ISSUE #774: Add user choice to delete previous arrays or ignore them
Remove the `free_all` by default behaviour to prevent corrupting multiple instances running simultaneously.

Now adds a user check to confirm the memory release - they should release if it's caused by a crash, but ignore if caused by multiple instances.

dtasev:
Should be merged after #773 

Shows a prompt to the user on startup if it finds any shared arrays present.

This is only a temporary "band-aid" for this release, to avoid having to merge #770 due to the seen memory leak issues. That's why it's not part of the user welcome screen, and also because I couldn't figure out a better way of grabbing the user's attention

To test:

- Start up a mantid imaging
- Load data in it
- Start up another one
- Select Ignore on the prompt
- Load data in it
- The first one should still work OK, try to applying an operation

Fixes #774

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -51,7 +51,7 @@ jobs:
       - name: pytest
         shell: bash -l {0}
         run: |
-          xvfb-run --auto-servernum python -m pytest -v --cov --cov-report=xml
+          xvfb-run --auto-servernum python -m pytest --cov --cov-report=xml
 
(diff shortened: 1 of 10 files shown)


samtygier-stfc (approved):
Looks good. I think a good compromise until switching to #762

PR #777, opened 2020-12-11
versioned docs using sphinx_multiversion

### LINKED ISSUE #743: Separate release and unstable packages documentation
What's the best way to handle this? Currently we publish the docs on the gh-pages branch. Note that if we change the structure there we will have to update the links in the code, or they might end up invalid.

Perhaps if we can publish versioned docs this could be easy. I've seen this on ReadTheDocs packages, but I'm not sure if it is a ReadTheDocs feature or a Sphinx feature.

The fastest, but perhaps not best or easiest way, would be to keep the docs inside gh-pages as sub-folders for each version:
```
.
|--- index.html
|
|--- v1.0.0/docs for this version
|
|--- v1.5.0/docs for this version
|
|--- v....
|
|--- unstable
```

samtygier-stfc:
Fixes #743

sphinx-multiversion can build a copy of the documentation for each tag
and place in a subfolder. The existing spinx setup can build the top
level latest version. versioning.html gives links to version in the
sidebar. layout.html overrides the main template to use correct versions
in the title and top navigation bar.

Note this requires a patch to sphinx_multiversion to allow calling a
prebuild command. This is used to call the docs_api for each checkout.
Holzhaus/sphinx-multiversion#62

To build the versioned docs:

```
python setup.py docs_versioned
```

This will build the api files, then the top level docs, then the
versioned docs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ import fnmatch
 import os
 import subprocess
 from distutils.core import Command
+import sys
 
 from setuptools import find_packages, setup
 from sphinx.setup_command import BuildDoc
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -137,6 +163,7 @@ def run(self):
     cmdclass={
         "docs_api": GenerateSphinxApidoc,
         "docs": BuildDoc,
```
dtasev, line 165: If `docs_versioned` is the go-to command, perhaps we can remove the commands `docs_api` `docs`, or is there some use-case for them still? 

I don't think we'll need to API on it's own, but perhaps unversioned docs, although we likely wouldn't want to publish those over the versioned ones so I'm not sure we need to keep it

dtasev (changes requested):
Looks good, will need to merge into master to make sure it fully works with the Actions automation.

Just a minor question added

setup.py
samtygier-stfc: We can't full remove the `docs_api` and `docs` rules, because the `docs_versioned` calls them. I couldn't find a way to trigger other rules from within the setup.py, other than using `subprocess`. (there is `distutils.core.run_setup` but I'd did not work for me),

Would it make sense to rename them? So there are internal `_docs` and `_api_docs` commands, and then rename `docs_versioned` to just `docs`?

samtygier-stfc: just pushed a new version with this

dtasev: That's great!

PR #780, opened 2020-12-14
761: Only use unittest mock

### LINKED ISSUE #761: Remove mock library
Several files use the external library `mock`. Should be possible to switch these to `unittest.mock` and remove the mock dependency.

DolicaAkelloEgwel:
Closes #761


samtygier-stfc (approved):
Looks good. Nice to remove an unneeded dependency.

Issue #565, opened 2020-10-05
Export histogram doesn't work on Main Window visualiser


dtasev:
Exporting the histogram from the main window doesn't seem to show the Plot as an option that can be exported:
![image](https://user-images.githubusercontent.com/9135965/95086958-02f32c80-0719-11eb-9be6-55a231a62938.png)

Compared to the operations window that has the Plots as choices (taken with combine histograms turned off):
![image](https://user-images.githubusercontent.com/9135965/95087029-17372980-0719-11eb-8136-f1f27954becb.png)

It seems that the same happens in `ImageView` itself, so maybe it is a bug in PyQtGraph? In that case the workaround might have to be manually recreating an `ImageView` by adding a Plot and setting an ImageItem on it (what Operations window does), then re-adding the ROI functionality

Some tutorial that I think does this: https://www.thetopsites.net/article/52389032.shtml

dtasev:
I did some investigation on this. The internal structure of the `ImageView` seems to be broken so it doesn't discover the correct widgets to export.

A way around this would be to just hide the Export button for `MIImageView`. The context menu is however buried pretty deep - we use the MIImageView.imageItem.menu for prepend our custom menu, but internally imageItem calls to parent widgets to get their menus too - and the graphics view at some point attaches the other options.

It doesn't seem like a lot of work, but I couldn't trivially just hide it.

Note that this also affects the Select ROI view (and other views if they use MIImageView)

samtygier-stfc:
#783 is the workaround, removing the Export item from the menu. I suggest that this bug remains open until export is actually fixed.

JackEAllen:
Closing for the time being, can re-open if IMAT ask for the ability to export data from various places which could be enabled by this.

PR #783, opened 2020-12-17
Remove Export action from StackVisualiser context menu

samtygier-stfc:
To workaround #565 hide the broken Export option

This removes the action for the GraphicsScene instance, so that it does not get
added to the context menu in GraphicsScene.addParentContextMenus()

### DIFF
--- a/mantidimaging/gui/dialogs/cor_inspection/recon_slice_view.py
+++ b/mantidimaging/gui/dialogs/cor_inspection/recon_slice_view.py
@@ -89,6 +89,18 @@ class CompareSlicesView(GraphicsLayoutWidget):
         for img in self.less_img, self.current_img, self.more_img:
             img.hoverEvent = lambda ev: self.mouse_over(ev)
 
+        # Work around for https://github.com/mantidproject/mantidimaging/issues/565
+        for scene in [
+                self.less_img.scene(),
+                self.less_hist.scene(),
+                self.current_img.scene(),
+                self.current_hist.scene(),
+                self.more_img.scene(),
+                self.more_hist.scene(),
+        ]:
+
+            scene.contextMenu = [item for item in scene.contextMenu if "export" not in item.text().lower()]
+
     def mouse_over(self, ev):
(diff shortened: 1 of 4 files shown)


mantidimaging/gui/windows/stack_visualiser/view.py
```diff
@@ -74,6 +74,12 @@ def __init__(self, parent: 'MainWindowView', dock: QDockWidget, images: Images):
 
         self.image_view = MIImageView(self)
         self.image_view.imageItem.menu = self._context_actions
+
+        # Work around for https://github.com/mantidproject/mantidimaging/issues/565
+        self.image_view.scene.contextMenu = [
+            item for item in self.image_view.scene.contextMenu if item.text() != "Export..."
```
Pasarus, line 80: Would it be possible to remove if it contains export i.e. item.text().lower().contains("export") opposed to exactly export because pyqtgraph might change under us and we want to make sure that the export doesn't pop back up randomly.

Pasarus (changes requested):

mantidimaging/gui/windows/stack_visualiser/view.py
samtygier-stfc: Yes, that's probably safer. Fixed

Pasarus (changes requested):
Would it be possible to remove the "Export..." option from the histogram context menu as well?

dtasev:
I think the above can be done if we move the change inside `MIImageView` itself (or it's presenter)

samtygier-stfc:
In some of those export is still useful, for example for saving images. Or should we hide it everywhere for now?

Pasarus:
I think we hide it for now, if a similar feature is required by users I think we should re-implement this window so we can fully support it, and only show options that work for our workflow.

samtygier-stfc:
This does all the widgets in main window and operations. A few more to go.

samtygier-stfc:
I think that is all of them.

PR #784, opened 2020-12-17
Load images when only flat or dark is selected

### LINKED ISSUE #648: Load images when only Flat or Dark is selected
Currently the Load dialog expects the selection to always be sample - but sometimes users might select only dark or flat.

~Potentially covered by #690 ?~

This can be done in multiple ways, but perhaps the simplest it so add a new `File > Load images` option which loads a single stack of images, and rename the current load to `File > Load dataset` which uses the current loading dialog and behaviour.

Another way is to just make the current load dialog accept a load when only the flat and dark fields are populated, but to me it seems more user friendly to go with a separate option for only images.

Pasarus:
Added functionality that allows users to load an individual stack instead of as a "sample" as part of a dataset.

Wonder if we have too many load options now, wizard time for loading?

To test:
- Open Imaging
- Click File->Load images stack
- Select an image from a stack and it should load the full stack from the same directory

Fixes #648

### DIFF
--- a/mantidimaging/gui/ui/main_window.ui
+++ b/mantidimaging/gui/ui/main_window.ui
@@ -35,6 +35,7 @@
      <string>File</string>
     </property>
     <addaction name="actionLoad"/>
+    <addaction name="actionLoadImages"/>
     <addaction name="actionSampleLoadLog"/>
     <addaction name="actionLoadProjectionAngles"/>
     <addaction name="actionLoad180deg"/>
@@ -76,7 +77,7 @@
   </widget>
   <action name="actionLoad">
    <property name="text">
-    <string>Load</string>
+    <string>Load dataset</string>
    </property>
    <property name="shortcut">
     <string>Ctrl+O</string>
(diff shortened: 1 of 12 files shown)


mantidimaging/gui/ui/main_window.ui
```diff
@@ -193,6 +194,11 @@
     <string>Load projection angles...</string>
    </property>
   </action>
+  <action name="actionLoadImages">
+   <property name="text">
+    <string>Load images stack</string>
```
dtasev, line 199:
```suggestion
    <string>Load images</string>
```
I don't like "Load images stack" but I'm also not sure this is much better

Pasarus: The reason I added "stack" was simply because I wanted to convey that it was a singular set of images we could go with "Load image set", with the myriad of options we now have the File menu I am starting to wonder if clarity is going to become an issue from a random User's perspective.

samtygier-stfc (changes requested):
Looks good.

Just needs to consistent with naming the load types.

mantidimaging/gui/ui/main_window.ui
```diff
@@ -35,6 +35,7 @@
      <string>File</string>
     </property>
     <addaction name="actionLoad"/>
```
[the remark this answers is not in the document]
Pasarus: Good catch :+1:

samtygier-stfc (approved):

PR #786, opened 2020-12-18
Pin all dependency versions

Pasarus:
Pinning versions so we control when an update change happens, should help with regression management.

Test is just checking builds pass on PR and on master afterwards.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -15,27 +15,27 @@ source:
 
 requirements:
   build:
-    - python==3.6.*
-    - pip==20.3.3
-    - setuptools==50.0.0
-    - sphinx==3.3.1
+    - python=3.6.*
+    - pip=20.3.3
+    - setuptools=50.0.0
+    - sphinx=3.3.1
   run:
-    - python==3.6.*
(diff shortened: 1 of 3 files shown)


conda/meta.yaml
```diff
@@ -15,26 +15,26 @@ source:
 
 requirements:
   build:
-    - python
-    - pip
-    - setuptools
-    - sphinx
+    - python==3.6.*
```
dtasev, line 18:
```suggestion
    - python==3.7.*
```

conda/meta.yaml
```diff
@@ -15,26 +15,26 @@ source:
...
+    - pip==20.3.3
+    - setuptools==50.0.0
+    - sphinx==3.3.1
   run:
-    - python
-    - pip
-    - astropy
-    - scipy
+    - python==3.6.*
```
dtasev, line 23:
```suggestion
    - python==3.7.*
```

dtasev (changes requested):

dtasev:
I will test this onto the release branch - I think it will be good to have the versions pinned for 2.0

PR #790, opened 2020-12-18
Add caching for Github Actions Conda and Pip to improve build time

Pasarus:
Just a CI change for actions with the intent to improve conda and pip install times.


dtasev:
Re-ran the jobs to see if cache gets hit. Will merge to master afterwards regardless. 

BTW does this also work for the build step installing anaconda (the one that takes hours) or do we need to do that separately?

Pasarus:
Having evaluated this PR I do not believe caching will actually benefit us for run time, this is due to the fact that it takes so long for the cache to download and de-compress for use that it is actually faster for the runners to just download fresh and create a new env.

PR #799, opened 2021-01-05
795: Issue and PR templates

DolicaAkelloEgwel:
PR template should be in the right place now. Have also cleared out the config.yml file. I think for the time being this means templates can only only be selected by manually changing the URL (https://github.blog/2018-01-25-multiple-issue-and-pull-request-templates/).


samtygier-stfc (approved):
Let merge and see how it looks

Issue #795, opened 2021-01-04
Issue and PR templates


Pasarus:
Create issue and PR templates to guide Developers through creating both.

PR template specifically needs to include release notes reminder.

PR #801, opened 2021-01-05
Fixing Issue Tags

DolicaAkelloEgwel:
### Issue

#795

### Description

Fixes the tags for the issue types. Removed the one for improvement / feature request as there isn't a single tag for this.

### Testing 

n/a

### Acceptance Criteria 

Check that the `Type: Bug` tag is added on my fork.

### Documentation

n/a

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -2,7 +2,7 @@
 name: Feature / Improvement Request
 about: Request an improvement or a new feature.
 title: ''
-labels: ''
+labels: 'Type: Enhancement'
 assignees: ''
 
 ---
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -2,7 +2,7 @@
 name: Feature / Improvement Request
 about: Request an improvement or a new feature.
 title: ''
-labels: 'enhancement'
+labels: ''
 assignees: ''
```
samtygier-stfc, line 6: We have an `Type: Improvement` bug tag, could use that.

samtygier-stfc (changes requested):

.github/ISSUE_TEMPLATE/feature_request.md
DolicaAkelloEgwel: There's also one for features, so it seems like it should be up to the issue-opener to select which one they want or I should split it up into two templates?

samtygier-stfc: Ah, I see. I think the developers are probably best placed to decide. So default to `enhancement` and one of us can change it if its a bigger change.

samtygier-stfc (approved):

PR #802, opened 2021-01-06
Enable drag and drop loading for 

### LINKED ISSUE #778: Allow drag-dropping folders to load them
- If a sample folder is dropped it opens the load dialog & searches for flat and dark images. Otherwise it could just load it as a stack.
	- If a the parent folder of a sample is dropped (e.g. `sample/..`) then do the same thing

Perhaps the easiest (but maybe not smartest) way to distinguish is just the number of images. Projections will usually be a number larger than 100, while flat/dark less. If uncertain we can always ask the user "Is this a sample or flat/dark/etc"

Pasarus:
### Issue
Closes #778 

### Description
Allows a user to drag a folder containing a sample onto the program and it will load in the stacks.

### Testing
- Open imaging
- Drag on a folder containing a sample + flat (before/after) + dark (before/after) + monitor normalisation
- The sample should be loaded fine
- Try to break the sample, by running a full recon, performing monitor normalization and all standard operations pre-recon.

### Documentation
No


mantidimaging/gui/windows/main/view.py
```diff
@@ -354,3 +355,23 @@ def set_images_in_stack(self, uuid: UUID, images: Images):
...
+            if os.path.isdir(file_path):
+                # Load directory as stack
+                sample_loading = self.presenter.load_stacks_from_folder(file_path)
+                if not sample_loading:
+                    QMessageBox.critical(
+                        self, "Load not possible!", "Please provide a directory that has .tif or .tiff files in it, or "
+                        "a sub directory that do not contain dark, flat, or 180 in their title name, that represents a"
+                        " sample.")
+            else:
+                QMessageBox.critical(self, "Load not possible!", "Please drag and drop only folders/directories!")
```
samtygier-stfc, line 377: If I select multiple images from a folder and drag them, then I get the "Please drag and drop only folders/directories!" for each one. Could that be set to only show once per drop.

mantidimaging/core/io/utility.py
```diff
@@ -153,11 +152,73 @@ def _alphanum_key_split(path_str):
     return [int(c) if c.isdigit() else c for c in alpha_num_split_re.split(path_str)]
 
 
-if __name__ == '__main__':
-    import doctest
+def get_prefix(path: str, separator="_"):
+    return path[:path.rfind(separator)]
```
samtygier-stfc, line 156: I prefer `path.rpartition(separator)[0]`

samtygier-stfc (changes requested):
Looks fine. Loads the Tilt_1 example I have. Just the couple of small things.

mantidimaging/core/io/utility.py
Pasarus: This is just a refactored method from somewhere else.

mantidimaging/gui/windows/main/view.py
Pasarus: I will find a fix! good spot!

samtygier-stfc (approved):

PR #812, opened 2021-01-11
Rescale now uses image.data.min() for offset to translate negative values

### LINKED ISSUE #808: Rescale does not handle negative values correctly
### Summary

Re-scaling an image that has negative values to **unsigned** int16 is causing data loss.

The issue is that we are not preparing the negative values, but directly converting to uint16, which causes the negative values to overflow and come up as artifacts.

### Steps To Reproduce

- Do a recon of tilt 4
- Rescale from operations, the issue should be seen in the preview
	- Or save to int16 and load

### Expected Behaviour

Rescale should "translate" the values to start at 0 - basically do `images.data += images.data.min()`.

This should be recorded as the "offset" in the metadata, so that we can come back to the float32 values range (for which we haven't yet added an operation)

### Current Behaviour

It doesn't translate the values so they end up overflowing at the top of uint16 range

dtasev:
Fixes #808 

### Description

Rescale now adds the `abs(min_value)` to the image to translate all the values in range [0, ...]. This prevents issues with negative values overflowing when converted to `uint16` before saving or during Rescale.

The single image takes this offset as a parameter, so that the same one can be used for all images.

Tests have been added for the new functionality.

### Testing 

- Load a stack and reconstruct it
- Check the histogram of the reconstructed images - they're likely to have some pixels with negative values
- Rescale manually to int16
	- Make sure you have a min input that takes the negative values as well, by default it starts at 0 so they will be cropped out
- The whole distribution should look the same, except now it will be in [0, 65535] value range
- Try rescaling to float32 and check that the distribution remains the same

### Acceptance Criteria 

`Conda` tests should pass. The others will fail due to Docker image incompatibility.

### Documentation

Single image takes offset as parameter, so that the same one can be used for all images

### DIFF
--- a/mantidimaging/core/io/saver.py
+++ b/mantidimaging/core/io/saver.py
@@ -102,9 +102,9 @@ def save(images: Images,
     make_dirs_if_needed(output_dir, overwrite_all)
 
     # Define current parameters
-    min_value = images.data.min()
+    min_value = np.nanmin(images.data)
     offset = abs(min_value)
-    max_value = images.data.max()
+    max_value = np.nanmax(images.data)
     int_16_slope = max_value / INT16_SIZE
 
     # Do rescale if needed.
--- a/mantidimaging/core/io/saver.py
+++ b/mantidimaging/core/io/saver.py
@@ -102,6 +102,8 @@ def save(images: Images,
     make_dirs_if_needed(output_dir, overwrite_all)
 
     # Define current parameters
+    min_value = images.data.min()
+    offset = abs(min_value)
     max_value = images.data.max()
     int_16_slope = max_value / INT16_SIZE
 
@@ -109,7 +111,8 @@ def save(images: Images,
     if pixel_depth is None or pixel_depth == "float32":
         rescale_params = None
     elif pixel_depth == "int16":
-        rescale_params = {"offset": 0, "slope": int_16_slope}
+        # turn the offset to string otherwise json throws a TypeError when trying to save float32
+        rescale_params = {"offset": str(offset), "slope": int_16_slope}
     else:
         raise ValueError("The pixel depth given is not handled: " + pixel_depth)
 
@@ -148,8 +151,12 @@ def save(images: Images,
             for idx in range(num_images):
                 # Overwrite images with the copy that has been rescaled.
                 if pixel_depth == "int16":
-                    write_func(rescale_single_image(np.copy(images.data[idx]), min_value, max_value, INT16_SIZE - 1),
-                               names[idx], overwrite_all)
+                    write_func(
+                        rescale_single_image(np.copy(images.data[idx]),
+                                             min_input=min_value,
+                                             max_input=max_value,
+                                             max_output=INT16_SIZE - 1,
+                                             offset=offset), names[idx], overwrite_all)
                 else:
                     write_func(data[idx, :, :], names[idx], overwrite_all)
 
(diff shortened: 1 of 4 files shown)


mantidimaging/core/io/saver.py
```diff
@@ -102,14 +102,17 @@ def save(images: Images,
     make_dirs_if_needed(output_dir, overwrite_all)
 
     # Define current parameters
+    min_value = images.data.min()
+    offset = abs(min_value)
```
samtygier-stfc, line 106: I don't think should be abs().

If `min_value` is positive, either `offset` should be zero or rescale should shift the values down.

mantidimaging/core/operations/rescale/rescale.py
```diff
@@ -33,6 +33,10 @@ def filter_func(images: Images,
                     data_type=None) -> Images:
         images.data[images.data < min_input] = 0
         images.data[images.data > max_input] = 0
+
+        # offset - it removes any negative values so that they don't overflow when in uint16 range
+        images.data += abs(images.data.min())
```
samtygier-stfc: abs() again

samtygier-stfc (changes requested):
I think using abs() for the offset is not really right.

I think the formula needs to be:
`image = (image - min_input) / (max_input - min_input) * max_output`
so that it handles positive or negative min

mantidimaging/core/operations/rescale/rescale.py
```diff
@@ -44,9 +47,17 @@ def filter_func(images: Images,
...
+    def filter_single_image(image: ndarray,
+                            min_input: float,
+                            max_input: float,
+                            max_output: float,
+                            offset: float,
+                            data_type=float32):
+        if offset < 0:
+            raise ValueError("The offset value provided must not be negative.")
         image[image < min_input] = 0
         image[image > max_input] = 0
```
samtygier-stfc, line 59: Clearing the out of range values will need to happen after rescaling now, because for some cases zero will be a value in range.

e.g. If the data spans from -10 to +10. and I set the limits to -5 to +5. then a pixel at -8 will get set to zero, then scaled into the middle of the final output range

mantidimaging/core/operations/rescale/rescale.py
samtygier-stfc: should just be `images.data -= min_input`

dtasev:
I think with the suggested changes it looks better. I'm just investigating whether the forced rescale to uint16 on save is causing any issues. 

I wonder if it's worth removing that mandatory rescale and replacing it with a user-facing warning instructing to rescale the data if they wish to save out in int16. I think the way Octopus did it, it essentially put you into a specific rescale GUI, which also acted as the save window, but we don't have that for 2.0

dtasev:
I was getting some errors in the test and it made me realise that having offset the data, I'd also have to offset the max_value so that the clip at the end is correct.

However, I think that can be avoided if I move the clip back to before, but instead of clipping to 0, it actually clips like it should have been the first time - to the `min`/`max` values used `np.clip(images.data, min_input, max_input, out=images.data)`

samtygier-stfc (approved):

samtygier-stfc:
I think this is good now. I feel like we should revisit after release and work out exactly what the use case is, and what is best to do with values outside the range. It seemed like it was being used partly to clip out noise around zero, so maybe that needs to be specific tool.

PR #814, opened 2021-01-12
Add parsing for IMAT CSV Log files

### LINKED ISSUE #809: Increase range of angle of rotate stack
### Desired Behaviour

Rotate stack can go from -359 to 359 degrees.

### Current Behaviour

It only goes from -180 to 180

### Comments
Making this a release essential only because it's a tiny change with no risk.

dtasev:
### Issue
Fixes #809

### Description

Adds parsing for CSV files. Keeps previous functionality to parse old TXT files

### Testing 

Load the flower dataset, **make sure to check the `Use?` checkbox for the Sample Log!** otherwise logs will not be loaded

- Test that the old logs are still loaded correctly
	- Either hook up a debugger 
	- or on the loaded stack check that `right click > show history` shows a logfile entry
	- or run monitor normalisation and check that the images get normalised
- Test the same steps with this CSV log (converted from the text one): [TomoIMAT00010675_FlowerFine_log.zip](https://github.com/mantidproject/mantidimaging/files/5803429/TomoIMAT00010675_FlowerFine_log.zip)
- The usage and results should be identical

### Acceptance Criteria 

Conda tests should pass. The docker tests will fail as this is based on release 2.0 branch.

### DIFF
--- a/mantidimaging/core/data/test/fake_logfile.py
+++ b/mantidimaging/core/data/test/fake_logfile.py
@@ -25,9 +25,7 @@ def generate_txt_logfile() -> IMATLogFile:
 
 def generate_csv_logfile() -> IMATLogFile:
     data = [
-        TextLogParser.EXPECTED_HEADER_FOR_IMAT_TEXT_LOG_FILE,  # checked if exists, but skipped
-        "",  # skipped when parsing
-        # for each row a list with 4 entries is currently expected
+        CSVLogParser.EXPECTED_HEADER_FOR_IMAT_CSV_LOG_FILE,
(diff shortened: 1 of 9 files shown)


DolicaAkelloEgwel (commented):
Should the `load_sample_log_dialog` method also have `.csv` in the file filter? It's for the "Load log for stack..." option that's accessed from the main window.

dtasev:
Great catch! I've pushed the change now

DolicaAkelloEgwel (approved):
I'm able to load CSV logs using both "Load" and "Load log for stack..."

PR #832, opened 2021-01-20
Fix help link lookup for operations

### LINKED ISSUE #815: Operations help URLs are off by 1/2 depending on how far down the filter is
### Summary

Clicking the `?` help buttons opens the wrong filter, for any filters below the separators.

### Steps To Reproduce

- Go to Operations window
- Select Rebin
- Click ?

### Expected Behaviour

The correct page is opened

### Current Behaviour

The Rescale help page will be opened

### LINKED ISSUE #820: Operation help goes to wrong page
### Summary

The help for any filter below the first separator (`--------`) goes to the wrong help page.

This seems to happen in the `mantidimaging/gui/windows/operations/view.py::open_help_webpage` function, the `currentIndex()` no longer returns the correct index as we now have separators inside the list.

A crude fix for this was done in #821 but that will be contained only for the release branch.

### Steps To Reproduce

- Open Operations window
- Select Divide
- Click the `?` for help
- The web page opened will not be for the divide filter

### Expected Behaviour

The help page should always be correct

samtygier-stfc:
### Issue

Fixes #820
Fixes #815

### Description

Use `_find_filter_index_from_filter_name()` to find by name rather than menu index.

### Testing 

From the operations window use the ? button to open the help pages for operations. Check that operations from each of the groups opens the correct page

### Acceptance Criteria 

Check that operations from each of the groups opens the correct page

### Documentation

Updated release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -11,3 +11,4 @@ Fixes
 -----
 
 - #816 : Error finding screen size in some dual screen set ups
+- #820 : Help links go to wrong page
--- a/mantidimaging/gui/windows/operations/view.py
+++ b/mantidimaging/gui/windows/operations/view.py
@@ -203,7 +203,9 @@ class FiltersWindowView(BaseMainWindowView):
         self.notification_text.setText(f"{operation_name} completed successfully!")
 
     def open_help_webpage(self):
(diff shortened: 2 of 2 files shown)


DolicaAkelloEgwel (approved):
Looks good to me.

PR #837, opened 2021-01-25
823: Transpose colour palette

### LINKED ISSUE #823: Automatically transpose grayscale palette to a color palette
### Desired Behaviour

Have a button (or some other way to access the feature) that converts the grayscale palette into a color palette that tries to assign different colors to each material in the sample.

Two things are needed:
- Check if we can programmatically control the color scale of ImageView
- Find an algorithm that can produce good scale points for each color from the histogram

DolicaAkelloEgwel:
### Issue

Closes #823

### Description

Allows users to change the colour palette using the Jenks/Otsu algorithms.

"thermal" default:
image
Jenks with three materials:
image
Multi Otsu with three materials:
image

It looks like the tilt dataset will be a better test for these algorithms, but I haven't been able to try this out yet.

### Testing 

Created various tests for the presenter in `PaletteChangerPresenterTest`.

### Acceptance Criteria 

Check that the tests pass and cover sensible test cases.
Run the Auto Palette Changer with Jenks and see that n_materials + 1 new ticks appear, and that they replace the original ticks.
Run the Auto Palette Changer with Otsu and see that 3 new ticks appear, and that they replace the original ticks. Also check that choosing Otsu causes the material spin box to become disabled, and that switching back to Jenks enables it again.

Check that the docs make sense, don't contain any mistakes, etc.

### Documentation

Added this to the list of new features. 
Created a colour palette section in the docs under GUI.


mantidimaging/gui/windows/recon/view.py
```diff
@@ -159,6 +163,15 @@ def on_row_change(item, _):
         self.reconHelpButton.clicked.connect(lambda: self.open_help_webpage("reconstructions/index"))
         self.corHelpButton.clicked.connect(lambda: self.open_help_webpage("reconstructions/center_of_rotation"))
 
+        self.hists = [self.image_view.recon_hist, self.image_view.sinogram_hist, self.image_view.projection_hist]
+        self.auto_colour_action = QAction("Auto")
+        self.auto_colour_action.triggered.connect(self.on_change_colour_palette)
+
+        for hist in self.hists:
```
samtygier-stfc, line 170: At the moment this is only applying to a single image, so there no need to add 'Auto' to all the menus.

mantidimaging/gui/windows/recon/view.py
```diff
@@ -381,3 +394,7 @@ def open_help_webpage(self, page: str):
 
     def change_refine_iterations(self):
         self.refineIterationsBtn.setEnabled(self.algorithm_name == "SIRT_CUDA")
+
+    def on_change_colour_palette(self):
+        change_colour_palette = PaletteChangerView(self, self.hists, self.image_view.projection.image)
```
samtygier-stfc, line 399: Can you move this all to work on the reconstructed view `self.image_view.recon.image`

samtygier-stfc (changes requested):
Looking good so far. If it can be shifted over to the reconstructed view I think that will show its usefulness better.

Later on it might be best to move it from ReconstructWindowView to MIImageView so that it can be used more widely.

samtygier-stfc (changes requested):
Looking good. Happy to merge it with the copyright date updated.

mantidimaging/gui/widgets/palette_changer/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (C) 2020 ISIS Rutherford Appleton Laboratory UKRI
```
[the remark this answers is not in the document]
DolicaAkelloEgwel: I've just changed this but the copyright check is failing. Will this fix itself after the merge...?

samtygier-stfc (approved):

samtygier-stfc:
Last test issues due to jenkspy module in docker. Should fix itself once this is in master.

PR #842, opened 2021-01-28
Add githubs security code analysis

Pasarus:
Add githubs security code analysis

it should just work in the background making sure we do not have any security vulnerabilities. It should run, if I didn't read my cron timers wrong, weekly, so should largely be a quiet little tool process in the background.


samtygier-stfc (approved):
Looks useful.

PR #879, opened 2021-03-03
865: Auto update from spinbox change

### LINKED ISSUE #865: Operations auto update not triggered on spinbox changes
### Summary

In the operations window some changes to filter parameters immediately trigger an update tot the preview, but changes to spin box values on cause an update  when moving focus or clicking update now.

### Steps To Reproduce

Most obvious in Reb,

Open a dataset, open operations, select rebin.

Change the rebin factor, by typing or using spin buttons.

### Expected Behaviour

Live update (as long as Auto Update enabled

### Current Behaviour

Update only happens when tabbing away

### Context

2.0 and master

DolicaAkelloEgwel:
### Issue

Closes #865

### Description

Makes the spin box update methods trigger during `valueChanged` rather than `editingFinished`. Also moved the disabling/enabling of the preview index spin box to the view as I feel like it belongs there rather than in the presenter.

### Testing 

None. There currently aren't any tests for the files I added code to.

### Acceptance Criteria 

Change the spinbox values for the operations and check that the preview updates, and the value only increases once.

### Documentation

Changed the release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -18,3 +18,4 @@ Fixes
 - #843 : Error loading stack
 - #856 : Move shared array allocation to `multiprocessing.Array` when on Python 3.8
 - #854 : Ops window, image selector in preview section skips by 2
+- #856 : Operations auto update not triggered on spinbox changes
--- a/mantidimaging/core/operations/rebin/rebin.py
+++ b/mantidimaging/core/operations/rebin/rebin.py
@@ -88,7 +88,7 @@ class RebinFilter(BaseFilter):
 
         def size_by_factor_toggled(enabled):
             factor.setEnabled(enabled)
-            on_change()
+            on_change(None)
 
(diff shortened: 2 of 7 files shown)


samtygier-stfc (changes requested):
Solves the original problem, but makes it much hard to change the value by typing in a new one.

For example if you open the rebin filter, select the factor box and try to type `0.6`. As soon as I press `0`, the value changes to `0.00`, if I put the cursor after the decimal places and type `6` nothing happens, I guess because there is a limit on the number of positions allowed.

Is it possible to trigger the on_change without resetting it to `#,##`?

Or if not could a timer be used. So valueChanged starts a 1 second timer, any additional change resets the timer to 1 second, and when the timer finishes it calls _on_change_and_disable.

samtygier-stfc (approved):
Thanks. Works nicely now.

PR #880, opened 2021-03-04
874: Operation window widgets resize

### LINKED ISSUE #874: Widgets in Operations window need minimum sizes
### Summary

Some parts of the operation window shrink to nothing is the window is resized small enough

### Steps To Reproduce

Open operations window, resize as small as possible

### Expected Behaviour

Minimum size is large enough for all widgets to be rendered

### Current Behaviour

image

Pixel value histogram view is lost

### Context
Current master

DolicaAkelloEgwel:
### Issue

Closes #874

### Description

Sets a fixed height for the pixel values histogram rather than the image layout.

### Testing 

Tested it out on my own machine and IDAaaS. 

### Acceptance Criteria 

Reduce the size of the operations window and check that everything remains visible.

### Documentation

Updated the release notes.


samtygier-stfc (approved):
Looks good.

PR #882, opened 2021-03-08
866: Flat fielding warning

### LINKED ISSUE #866: Warning if running Flat-fielding again
### Desired Behaviour

Running flat-fielding twice is unnecessary and looses data. Could give a warning if history of stack already has flat-fielding applied.

### Current Behaviour

No warning. Data is lost. With safe apply off this is easy to do by accidentally double clicking apply to stack button.

DolicaAkelloEgwel:
### Issue

Closes #866

### Description

Displays a warning message if the user runs flat-fielding more than once.

### Testing 

Wrote tests to go through the different cases:
- running flat fielding a second time -> message displayed
- not running flat fielding at all -> no message displayed
- running flat fielding for the first time w/ non-empty operation history -> no message displayed
- running flat fielding for the first time w/ empty operation history -> no message displayed
- running flat fielding a second time and press cancel -> filter isn't run

Also made tests for the changes to the apply buttons.

### Acceptance Criteria 

Check that the tests pass and cover appropriate test cases. Check that a message appears when you try to run flat-fielding twice. Check that pressing cancel when the message appears causes the operation to be cancelled. Check that the apply buttons become disabled until the operation is complete.

### Documentation

Updated the release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -1,4 +1,3 @@
-
 Mantid Imaging next release
 ===========================
 
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -1,3 +1,4 @@
+
 Mantid Imaging next release
 ===========================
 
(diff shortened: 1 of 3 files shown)


samtygier-stfc:
Could you also disable the apply buttons while the operation is running, so that its not possible start 2 operations at the same time.

samtygier-stfc (approved):
Looks good.

PR #892, opened 2021-03-18
886: Don't reset zoom when filter parameters are changed

### LINKED ISSUE #886: Don't reset zoom when changing operation parameters
### Summary

It is useful to be able to adjust parameters while zoomed in to see their effect.

### Steps To Reproduce

Open dataset, open operations window. Switch to median.

Zoom in so that pixel details are visible.

Change the kernel size

### Expected Behaviour

preview updates.

### Current Behaviour

preview zoom level resets

### Context

reported with 2.0. also occurs on master

### Screenshot(s)

image

Then click to adjust kernel size, and zoom resets.
image

DolicaAkelloEgwel:
### Issue

Closes #886

### Description

Adds checkboxes for locking the zoom and histogram scales.

### Testing 

Updated one test and added another for checking what happens when the boxes are checked.

### Acceptance Criteria 

Check the boxes and change the operation parameters. Check that the zoom/histogram scale remains the same.

### Documentation

Updated the release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -25,3 +25,4 @@ Fixes
 - #878 : Update update instructions in version check
 - #885 : RuntimeError after closing wizard
 - #805, #875 : Fix segmentation fault due to object lifetime
+- #886 : Don't reset zoom when changing operation parameters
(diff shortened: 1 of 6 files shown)


samtygier-stfc (approved):
Nice, This should make it much easier to fine tune parameters.

PR #893, opened 2021-03-19
873: Link histogram scales

### LINKED ISSUE #873: Link histogram scales in operations window
### Desired Behaviour

For operations such as Remove outliers it would be useful if the colour gradient could be link between the before and after images,

For other operations such as flat fielding, they should not be link because the input and output values are very different.

This can probably set per operation rather than adding another checkbox.

### Current Behaviour

The number scale is linked, but moving the yellow limit bars are independent between the images.

image

DolicaAkelloEgwel:
### Issue

Closes #873

### Description

Links the histogram scales of the before and after images for some of the filters.

### Testing 

Added a test for checking if the argument passed to the histogram linking function is True or False. Haven't actually tested what takes place in the filter_previews file as it doesn't have any tests.

### Acceptance Criteria 

Check that the following filters have linked histograms:

- Crop coordinates
- Remove outliers
- ROI Norm
- Circular mask
- Divide
- Monitor Normalisation
- Rebin
- Ring Removal
- Clip values
- Median
- Gaussian
- all Stripe operations

Check that changing to a non-linked operation (such as flat-fielding) breaks the connection, and then switching back to one of the above operations restores it.

### Documentation

Updated release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -8,6 +8,8 @@ New features
 
 - #823: Automatically transpose colour palette
 - #567: Add new user wizard
+- #886 : Don't reset zoom when changing operation parameters
+- #873 : Link histogram scales in operations window
 
 Fixes
 -----
@@ -25,5 +27,3 @@ Fixes
 - #878 : Update update instructions in version check
 - #885 : RuntimeError after closing wizard
 - #805, #875, #891 : Fix segmentation fault due to object lifetime
-- #886 : Don't reset zoom when changing operation parameters
-- #873 : Link histogram scales in operations window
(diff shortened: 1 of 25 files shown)


samtygier-stfc (changes requested):
Works well. I think median, gaussian and the rest of the stipe should be added as well.

Could you try making this a boolean class value inside the filter classes, (like filter_name currently is). Then do_register_active_filter() can ask the active filter if it should use linked histograms.

samtygier-stfc:
Can you make so that when it first opens it applies the settings from `Image before` to `Image after` rather than the other way around.

Currently the first time I open the operations window I get 
image

samtygier-stfc (changes requested):
On Flat fielding, if I adjust the before scale I get:

```
File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/operations/filter_previews.py", line 263, in link_image_after_to_before_hist_range 
  self.image_after_hist.sigLevelChangeFinished.disconnect()
TypeError: disconnect() failed between 'sigLevelChangeFinished' and all its connections
```

samtygier-stfc (approved):

samtygier-stfc:
Working well now :+1:

PR #909, opened 2021-03-26
Prevent simultaneous operations

### LINKED ISSUE #716: Parallel module can't run in parallel
Because the multiprocessing we use shares a global object across processes it has the limitation that running _two_ async parallel processing will have a guaranteed clash, as each one will change the shared data variable and cause some exception in the other.

Solution with `SharedArray`: because each array has a unique identifier, we can instead use this to get a handle to the data during processing.

Other solutions would probably require moving away from `SharedArray` completely, but as Python multiprocessing seems to be notoriously annoying to do, may have different drawbacks (i.e. if shared over a socket it may be slower).

samtygier-stfc:
### Issue

closes #716 

### Description

Some GUI changes so that a user can't start an operation and reconstruction at the same time.

Prevent ops and recon windows closing while something is in progress. Prevent both windows being open at the same time.

Check that ops and recon windows can't be closed with an action is progress.

Check that they cant both be opened at the same time

### DIFF
--- a/docs/release_notes/2.1.rst
+++ b/docs/release_notes/2.1.rst
@@ -30,6 +30,7 @@ Fixes
 - #878 : Update update instructions in version check
 - #885 : RuntimeError after closing wizard
 - #805, #875, #891 : Fix segmentation fault due to object lifetime
+- #716 : Prevent simultaneous operations
 
 Developer Changes
 -----------------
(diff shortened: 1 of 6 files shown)


DolicaAkelloEgwel (approved):
Looks good to me.

PR #913, opened 2021-04-01
Fix ROI normalisation scaling

### LINKED ISSUE #912: Fix ROI normalise due to rescale changes
### Summary

Part of  #903

#812 changed the behaviour of rescale to better handle negative values. This effected ROI normalisation which expected rescale to be a simple multiply.

The original method is to:
Record the max initial pixel value for the full stack
For each slice, take the mean of the ROI (air region), divide the whole slice by that mean
Rescale (multiply) to whole stack so that the new max pixel value is equal to the original max value

With the change in #812 the rescale effectively move the lowest value to zero by offsetting

### Context

Since PR #812

### Screenshot(s)
before
Screenshot at 2021-04-01 12-02-01_before

after (note that the low values have been shifted) Shifts can be extreme if there are low outliers
Screenshot at 2021-04-01 12-03-25_after

samtygier-stfc:
### Issue

Closes #912 

### Description

Instead of calling RescaleFilter, move the rescale step into RoiNormalisationFilter

Restore pre 2.0 behaviour so that offset is not applied

In the comparison window after running ROI normalise, ensure that there is no offset of small value introduced.

### Documentation

Added to release notes

### DIFF
--- a/docs/release_notes/2.1.rst
+++ b/docs/release_notes/2.1.rst
@@ -32,6 +32,7 @@ Fixes
 - #805, #875, #891 : Fix segmentation fault due to object lifetime
 - #716 : Prevent simultaneous operations
 - #901: Rebin Operation failed: '<' not supported between instances of 'Progress' and 'int'
+- #912: Fix ROI normalise due to rescale changes
 
 Developer Changes
 -----------------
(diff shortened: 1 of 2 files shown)


samtygier-stfc:
Expected change in GUI test
image

Pasarus (approved):

PR #923, opened 2021-04-09
Add ROI normalisation modes

### LINKED ISSUE #914: Better ROI norm options
### Current Behaviour

Normalisation is scaled so that the maximum pixel value in the stack after is equal to the maximum value before.

This means that outliers can have an effect.

### Desired Behaviour

It would be more stable to normalise to preserve the mean value of the air region. I.e scale the brightness of each slice to make the mean of the air region equal to the mean of the air region across the whole stack.

It would also be useful to normalise to the mean of the air region in the flat field.

samtygier-stfc:
### Issue

Closes #914 

### Description
Add ROI normalisation modes

'Preserve Max' : Normalisation is scaled such that the the maximum pixel value of the stack is equal before
	     and after the operation.
'Stack Average' : The mean value of the air region across all projections is preserved.
'Flat Field' : The mean value of the air regions in the projects is made equal to the mean value of the air
	   region in the flat field.

'Preserve Max' is equivilent to the old behaviour.

### Testing 

Tests are updated, and check for sensible behaviour of the 3 modes

### Acceptance Criteria 

Each mode works

### Documentation

Release notes and doc strings updated

### DIFF
--- a/mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
+++ b/mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
@@ -3,6 +3,7 @@
 
 from functools import partial
 from logging import getLogger
+from typing import List, Optional
 
 import numpy as np
 
@@ -15,6 +16,11 @@ from mantidimaging.core.utility.progress_reporting import Progress
 from mantidimaging.core.utility.sensible_roi import SensibleROI
 from mantidimaging.gui.utility import add_property_to_form
 from mantidimaging.gui.utility.qt_helpers import Type
+from mantidimaging.gui.widgets.stack_selector import StackSelectorWidgetView
+
+
+def modes() -> List[str]:
+    return ['Preserve Max', 'Stack Average', 'Flat Field']
 
 
 class RoiNormalisationFilter(BaseFilter):
@@ -31,7 +37,13 @@ class RoiNormalisationFilter(BaseFilter):
     link_histograms = True
 
     @staticmethod
-    def filter_func(images: Images, region_of_interest: SensibleROI = None, cores=None, chunksize=None, progress=None):
+    def filter_func(images: Images,
+                    region_of_interest: SensibleROI = None,
+                    normalisation_mode: str = modes()[0],
+                    flat_field: Optional[Images] = None,
+                    cores=None,
+                    chunksize=None,
+                    progress=None):
         """Normalise by beam intensity.
 
         This does NOT do any checks if the Air Region is out of bounds!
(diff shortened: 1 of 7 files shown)


samtygier-stfc:
Expected UI change
image

mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
```diff
@@ -45,41 +57,106 @@ def filter_func(images: Images, region_of_interest: SensibleROI = None, cores=No
                            from which sums will be calculated and all images will
                            be normalised.
 
+        :param normalisation_mode: Controls what the air regions are normalised to.
+            'Preserve Max' : Normalisation is scaled such that the the maximum pixel value of the stack is equal before
+                             and after the operation.
+            'Stack Average' : The mean value of the air region across all projections is preserved.
+            'Flat Field' : The mean value of the air regions in the projects is made equal to the mean value of the air
```
DolicaAkelloEgwel, line 64: Should this say projections?

samtygier-stfc: Yes

mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
```diff
@@ -45,41 +57,106 @@ def filter_func(images: Images, region_of_interest: SensibleROI = None, cores=No
...
+
         :param cores: The number of cores that will be used to process the data.
 
         :param chunksize: The number of chunks that each worker will receive.
         :param progress: Reference to a progress bar object
 
         :returns: Filtered data (stack of images)
         """
+        if normalisation_mode not in modes():
+            raise ValueError(f"Unknown normalisation_mode: {normalisation_mode}, should be one of {modes()}")
```
DolicaAkelloEgwel, line 77: It seems worth testing these two value errors.

mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
```diff
@@ -45,41 +57,106 @@ def filter_func(images: Images, region_of_interest: SensibleROI = None, cores=No
...
             progress = Progress.ensure_instance(progress, task_name='ROI Normalisation')
-            _execute(images.data, region_of_interest, cores, chunksize, progress)
+            _execute(images.data, region_of_interest, normalisation_mode, flat_field_data, cores, chunksize, progress)
         h.check_data_stack(images)
         return images
 
     @staticmethod
     def register_gui(form, on_change, view):
-        label, roi_field = add_property_to_form("ROI",
+        def string_contains_all_parts(string: str, parts: list) -> bool:
```
DolicaAkelloEgwel, line 98: Would it be worth taking these methods out and writing tests for them?

DolicaAkelloEgwel (approved):

Issue #906, opened 2021-03-26
Clip values


samtygier-stfc:
### Summary

Overlay difference on clip values does not seem to be showing the pixel values that have changed

### Steps To Reproduce

Load data set. choose min and max clip points near the edge of the histogram. enable Overlay Difference.

In the screen shot below I expect on the brightest and darkest pixels to be cropped, but the overlay colours almost everything red.

### Expected Behaviour

Only changed pixels should be hihglighted

### Current Behaviour

Many more are

### Context
Current master

### Failure Logs
None

### Screenshot(s)

![image](https://user-images.githubusercontent.com/74248560/112640018-d94dcf00-8e38-11eb-8d61-5092e44e4b21.png)

PR #930, opened 2021-04-13
Change outlier detection

### LINKED ISSUE #918: Outliers acts as median in dark mode
### Summary

Outliers should remove either bright or dark outliers (depending on mode), that are more than `diff` away from the local median. Removed pixels are set to the local median

In bright mode this works correctly. In dark mode it appears to be setting almost all pixels to the local median.

### Steps To Reproduce

Open a dataset, go to operations outliers. zoom in and compare bright and dark modes.

### Expected Behaviour

 Both should normally only change the values of a few pixels.

### Current Behaviour

Dark mode changes to value of almost every pixel.

### Context

Current master

### Screenshot(s)

image

samtygier-stfc:
Fix logic in outlier detection

### Issue

Closes #918 

### Description

Should always be looking for pixels with a difference greater than diff

Load a noisy image. Go to the remove outliers operation

On master you will see that with dark mode, the whole image is smoothed. With this patch only dark outliers will be effected.

Note that the overlay differences seems to misbehave here sometimes, highlighting unchanged pixels, I think this is #906

### Documentation
Release notes updated


samtygier-stfc:
Before
Screenshot at 2021-04-13 11-30-03

After
Screenshot at 2021-04-13 11-34-25

DolicaAkelloEgwel (approved):

PR #938, opened 2021-04-16
906: Clip values

### LINKED ISSUE #906: Clip values

DolicaAkelloEgwel:
### Issue

Closes #906

### Description

Makes the overlay colour all pixels where the absolute value of the change is > 10^-3.

### Testing 

Haven't done an, as there aren't any tests for the filter previews file.

### Acceptance Criteria 

Open the operations window and check that all red pixels in the image after overlay have a difference greater than 10^-3 while all unchanged pixels have a different smaller than this value.

### Documentation

Updated the release notes.

### DIFF
--- a/docs/release_notes/2.1.rst
+++ b/docs/release_notes/2.1.rst
@@ -39,6 +39,7 @@ Fixes
 - #914: Outliers acts as median in dark mode
 - #919: Median min kernel size
 - #905: Ring Removal has no effect
+- #906: Clip values
 
 Developer Changes
 -----------------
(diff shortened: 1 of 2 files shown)


samtygier-stfc (approved):
That's working how I'd expect now. Thanks.

PR #945, opened 2021-04-26
927: ROI norm: use averaged view in ROI selector

### LINKED ISSUE #927: ROI norm: use averaged view in ROI selector
### Desired Behaviour

The ROI selector has an averaged mode. This is useful to be sure that the selected air region is clear of the object at all angles.

It should be the default.

### Current Behaviour

Averaged mode must be enabled manually using the right click menu.

DolicaAkelloEgwel:
### Issue

Closes #927

### Description

Shows the averaged image by default in the ROI visualiser.

### Acceptance Criteria 

I had to check this using the debugger as the difference in appearance isn't always clear.

### Documentation

Updated release notes.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -20,3 +20,5 @@ centos7-base:
 ubuntu18-base:
 	docker build -f Ubuntu18_base.Dockerfile -t mantidproject/mantidimaging:ubuntu18-base ..
 	docker push mantidproject/mantidimaging:ubuntu18-base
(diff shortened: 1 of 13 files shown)


samtygier-stfc (changes requested):
If I open the ROI window a second time, it is no longer in averaged mode.

samtygier-stfc (approved):
Works well now.

PR #977, opened 2021-05-12
959: Prevent even values for median filter kernel

### LINKED ISSUE #959: Median GPU does not work with even values
### Summary

If the GPU option is enabled in the Median operations, then even values of kernel size give skewed output.

### Steps To Reproduce

In Median, choose an even kernel size and enable GPU.

### Expected Behaviour
Results match CPU version

### Current Behaviour
image

### Context

With current master, 2.1 and further back.

DolicaAkelloEgwel:
### Issue

Closes #959 

### Description

Prevents the user from using even numbers as the kernel size in the medial filter.

### Testing 

Wrote some tests for the `KernelSpinBox` and one of them uses `QTest`. It might be worth using it in other places.

### Acceptance Criteria 

Check that only odd numbers from 3 to 999 work with the kernel size.

### Documentation

Updated the release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -18,6 +18,7 @@ Fixes
 - #957 : Crop Coordinates ROI not changing
 - #961 : Test failure: ImportError: libcuda.so.1
 - #953 : NaNs in result when reconstructing with FBP_CUDA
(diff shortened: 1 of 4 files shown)


samtygier-stfc (approved):
Looks good

PR #993, opened 2021-05-21
964: Open Nexus files

### LINKED ISSUE #964: Allow loading NeXus files
### Desired Behaviour

Mantid Imaging should be able to load image data from NeXus files.

DolicaAkelloEgwel:
### Issue

Closes #964

### Description

Allows Mantid Imaging to read images from NeXus files. If the `image_key` isn't found, then all the images are treated as projections.

### Testing 

Wrote tests for the `NexusLoader` class and the `load_nexus_file` method in the 'MainWindowPresenter`.

### Acceptance Criteria 

Check that the tests pass. You can also try loading [this Diamond file](https://github.com/DiamondLightSource/Savu/blob/master/test_data/data/24737.nxs) or [this CLF file](https://stfc365-my.sharepoint.com/:u:/r/personal/stephen_dann_stfc_ac_uk/Documents/Tomography%20sample%20data/ncc7/lores/tomo_flat.nxs?csf=1&web=1&e=0VXQFp).

### Documentation

Updated release notes and updated loading and saving page of documentation.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -27,7 +27,7 @@ requirements:
     - scipy=1.5.3
     - scikit-image=0.18.*
     - numpy=1.19.4
-    - tomopy=1.10.*=cuda*
+    - tomopy=1.9.*=cuda*
     - cudatoolkit=9.2*
     - cupy
     - astra-toolbox=1.9.9.dev4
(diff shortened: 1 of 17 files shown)


samtygier-stfc (changes requested):
Load looks good. I can open both files.

I try to do operations such as ROI norm on the data I get type errors:
```Error applying filter for preview: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('uint16') with casting rule 'same_kind'```
When loading TIFFs MI converts to float32 or float64, which some of the operations require. So we should do the same with nexus files.

samtygier-stfc (approved):
Thanks, I can now load a file and get all the way through a reconstruction,

PR #997, opened 2021-05-24
Set viewport size

### LINKED ISSUE #996: Eyes tests use window size as viewport size
### Summary

Applitools eyes allows tests to be have different baselines to take into account viewport size. This makes sense for a webpage rendered in a particular browser size. For MI we upload a screenshot of a window, and applitools treats the size of the screenshot as the viewport size.

This means that if a UI change causes the window size to change, it will be taken as a new baseline at a new viewport size, and the test will pass. This may cause some UI changes to pass undetected.

### Steps To Reproduce

In https://eyes.applitools.com/app/test-results/00000251780687438218 the test_operations_window.py is marked as new. This is because the viewport size is 1465x733 compared to previous 1393x733.

### Expected Behaviour

A change in window size should be treated as a fail

### Current Behaviour

A change in window size is an automatic pass, even if the interface has change significantly.

### Context

From github actions

samtygier-stfc:
### Issue
Closes #996 

### Description

Set an explicit viewport size. This prevents changes in window size triggering create of a new baseline, rather than being compared to the existing baseline.

Thanks to Applitools support, for helping solve this.

### Testing 

This will trigger recreating all existing baselines, with a viewport of 1920x1080.

### Acceptance Criteria 

New baselines with a viewport of 1920x1080

### Documentation
Just a test fix. Comments in code.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -31,6 +31,6 @@ dependencies:
       - pytest-repeat==0.9.1
       - psutil==5.7.3
       - isort==5.6.4
-      - eyes-images==4.15.0
+      - eyes-images==4.20.*
       - parameterized==0.8.1
       - pre-commit
--- a/mantidimaging/eyes_tests/eyes_manager.py
+++ b/mantidimaging/eyes_tests/eyes_manager.py
@@ -12,6 +12,10 @@ from applitools.images import Eyes
 
 from mantidimaging.gui.windows.main import MainWindowView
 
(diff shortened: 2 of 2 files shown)


DolicaAkelloEgwel (approved):

samtygier-stfc:
Looks like this is not quite sufficient. The API returns an error when doing the comparison with different sized images. Following up with tech support.

PR #1033, opened 2021-06-30
Add center mode to ring removal

### LINKED ISSUE #904: Ring Removal display COR coordinates
The Ring removal filter needs the user to enter the centre of rotation coordinates. For most cases this will be the centre of the image.

### Desired Behaviour

The best option is probably to have a "Use image centre" or "manual", combobox. And entry for manual input if needed.

Investigate if the manual input could be done by clicking on the image.

Investigate if the COR could be shown on the image.

### Current Behaviour

"The xy COR coordinates are not displayed; I had to get those from the data window;"

samtygier-stfc:
When set to "image center", then pass None through to tomopy. This
causes it to default to the image center. This is right value unless the
image has been cropped after reconstruction.

### Issue

Closes #904 

### Description

When set to "image center", then pass None through to tomopy. This
causes it to default to the image center. This is right value unless the
image has been cropped after reconstruction.

Check that selecting "image center" disables the coordinate boxes.

For the tilt1_ready2 data, after a reconstruction I get this. Note that that corrections are centered on the image center.

Screenshot at 2021-06-30 18-35-29

with the same settings, but COR x and y set to 100, the corrections are offset 

### Documentation
release_notes

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -16,6 +16,7 @@ New features
 - #987 : Allow Flat-fielding without dark subtraction
 - #1011 : Pixel size should allow setting decimal places
 - #1013 : NeXus Loading Window
+- #904 : Default to using image center for ring removal COR
 
 Fixes
 -----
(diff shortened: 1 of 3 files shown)


samtygier-stfc:
image

GUI changes as expected.

DolicaAkelloEgwel (approved):

PR #1038, opened 2021-07-05
Revert screenshot paddding work around

samtygier-stfc:
Padding screenshots so that a fixed image size is sent to applitools
should no longer be needed.

### Issue
Reverts change in #1004

### Description

Remove padding

Also use a seed when generating a 'random' image, so that ignore regions are no longer needed.

Actions tests should pass.

### Documentation

Not needed


samtygier-stfc:
If I make a deliberate change to a window size, it gets picked up now:

image

DolicaAkelloEgwel (approved):

PR #1041, opened 2021-07-06
Replaces grep and awk usage in version check with Python equivalent

StephenNneji:
### Issue

None

### Description
In _version_check.py_, the installed version and label is extracted using

`conda list mantidimaging | grep '^[^#]' | awk 'END{print $2,$4}'`

The **grep** command gets every line that does not start with # then the **awk** command extracts version number and label from the last line.

This replaces grep and awk usage in _version_check.py_ with Python equivalent and updates unit test. 

### Testing 

The unit test is updated to reflect changes 

### Acceptance Criteria 

The code should extract the correct version and label (if present) from the string returned by `conda list mantidimaging` 

### Documentation

No change to docs

### DIFF
--- a/mantidimaging/core/utility/test/version_check_test.py
+++ b/mantidimaging/core/utility/test/version_check_test.py
@@ -74,9 +74,17 @@ class TestCheckVersion(unittest.TestCase):
 
     @mock.patch("subprocess.check_output")
     def test_retrieve_conda_installed_version(self, mock_check_output):
-        mock_check_output.return_value = b"1.1.0_1018 mantid/label/unstable"
+        no_package = "# packages in environment at mantidimaging-dev:\n#\n# Name  Version Build  Channel"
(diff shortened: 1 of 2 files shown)


samtygier-stfc:
The CI is hitting some formatting errors. Running `make yapf_apply` should fix them.

I recommend stetting up pre-commit, so you get the warnings at commit time. https://mantidproject.github.io/mantidimaging/developer_guide/testing.html#pre-commit

samtygier-stfc:
`401 Client Error: Unauthorized for url`

I think the GUI test is failing because actions is not allowing a non-member to use the projects secrets. I have sent an invite.

samtygier-stfc:
Not sure why the GUI testing is still failing. Maybe because its a pull from an external repository. I'll merge as is. For the next PR you could try pushing a branch to this repo.

samtygier-stfc (approved):

PR #1058, opened 2021-07-16
1055: Arithmetic operation breaks flat-fielding

### LINKED ISSUE #1055: After using Arithmetic operation flat-fielding does not work
### Summary

After using Arithmetic operation flat-fielding does not update the stack. The preview does work, but pressing apply does work.

Might be to do with the operation creating a new none shared array instead of operating in place. Maybe line 48.

### Steps To Reproduce

- Load a dataset.
- In Arithmetic operations do a multiply by 1.2
- Open flat fielding operation, notice the preview is correct
- Click Apply to Stack, notice that the stack in the main window is not updated

### Expected Behaviour

FF should still work

### Current Behaviour

FF does not work after Arithmetic

### Context

Current master.

DolicaAkelloEgwel:
### Issue

Closes #1055

### Description

Uses `Process` to do the arithmetic operations which ensures it remains as a shared array. Also added a log error for when the division/multiplication value is zero.

### Testing 

Added tests for the log messages.

### Acceptance Criteria 

Run the arithmetic filter and then flat-fielding on a dataset and check that it works.

### Documentation

Updated release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -36,6 +36,7 @@ Fixes
 - #1036 : Exception in ReconstructWindowView.show_recon_volume prevents recon closing
 - #1046 : Can't rotate NeXus images
 - #1048 : NeXus Loader: ValueError: Illegal slicing argument for scalar dataspace
+- #1055: Arithmetic operation breaks flat-fielding
 
 Developer Changes
 -----------------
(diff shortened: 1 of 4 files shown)


samtygier-stfc:
Could you do this with `ps.create_partial` and `ps.execute` like in flat_fielding, that the multiprocessing calls are abstracted away.

samtygier-stfc (approved):
Looks good and fixes for me.

PR #1131, opened 2021-09-07
1044: Open dataset with command line

### LINKED ISSUE #1044: Command line argument to load a dataset
### Desired Behaviour

It should be possible to load a dataset directly from the command line. This would be useful for saving some clicks while developing.

`mantidimaging /path/to/data`

or

`python -m mantidimaging /path/to/data`

### Current Behaviour

Dataset can only be loaded using the menu or by drag and drop.

DolicaAkelloEgwel:
### Issue

Closes #1044 

### Description

Allows a dataset to be loaded through the command line by using a `--path` argument.

### Testing 

Added a test to the `MainWindowView` for checking that the path is loaded if one has been set. Also a test that the load method isn't called in `__init__` if a path argument wasn't given.

### Acceptance Criteria 

Check that the tests pass. Check that giving a `--path` argument loads the images from that folder.

### Documentation

Updated release notes.


docs/release_notes/2.2.rst
```diff
@@ -25,6 +25,7 @@ New features
 - #816 : NaN Removal Filter
 - #1035 : NeXus Loader: Select every Nth image
 - #1035 : List of bad slices in error message could be compressed
+- #1044 : Command line argument to load a dataset
```
samtygier-stfc, line 28: This should be in `next.rst`

mantidimaging/core/utility/command_line_path.py
```diff
@@ -0,0 +1,21 @@
...
+    _instance = None
+    images_path = ""
+
+    def __new__(cls, path: str = ""):
+        """
+        Creates a singleton for storing the path from the command line argument.
+        """
+        if cls._instance is None:
+            cls._instance = super(CommandLinePath, cls).__new__(cls)
+            cls.images_path = path
```
samtygier-stfc, line 13: Can you validate that the path exists here, I think it would be good to exit without starting the GUI if path does not exist.

samtygier-stfc (changes requested):
It would also be good to give a similar error to drag and drop if the path can't be loaded. Some of the logic and error message in MainWindowView.dropEvent could be moved to MainWindowPresentor.load_stacks_from_folder, to avoid duplication.

samtygier-stfc (approved):

PR #1133, opened 2021-09-09
1118: Remove outliers does not remove NaN

### LINKED ISSUE #1118: Remove outliers does not remove NaN
### Summary

Remove outliers removes pixels that have a significant difference from their neighbours and replaces them with the local median. This should include NaNs

This may need a fix for  #1117 otherwise the replaced value might still be a NaN

### Steps To Reproduce

Open a data set with NaN pixels.
Remove outliers
Note that the NaNs are unaffected

### Expected Behaviour

NaNs should be removed

### Current Behaviour

NaNs remain

### Context
Current master

DolicaAkelloEgwel:
### Issue

Closes #1118 

### Description

Makes remove outliers remove NaN pixels.

### Testing 

Added a `test_nan_removal` test.

### Acceptance Criteria 

Check that the tests pass. Check that NaNs are removed when the filter is used.

### Documentation

Updated release notes.

### DIFF
--- a/docs/release_notes/2.2.rst
+++ b/docs/release_notes/2.2.rst
@@ -61,6 +61,7 @@ Fixes
 - #1112: Minimise Error, number of slices input should have minimum of 2
 - #1067: Failed to load stack "log file is not recognised"
 - #1121: FF when negative values in output, no completed message
+- #1118: Remove outliers does not remove NaN
 
 Developer Changes
 -----------------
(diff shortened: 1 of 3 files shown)


samtygier-stfc:
I have an evil file with neighbouring NaNs that shows that using `scipy_ndimage.median_filter` is going to cause problems. I don't think I can attach it so I'll email it.

samtygier-stfc:
Closing this. NaN removal must now be done with the NaN removal tool.

PR #1149, opened 2021-09-21
Switch from sarepy to algotom

### LINKED ISSUE #1023: Switch from sarepy to algotom
https://pypi.org/project/algotom/

samtygier-stfc:
### Issue

Closes #1023 

### Description

The methods in Sarepy have been integrated into Algotom. 

Docker tests will fail due to new dependency, conda one should pass.

If you create a new dev env
`python3 ./setup.py create_dev_env`
then tests should pass locally, and stripe removal operations should function

### Documentation

Release notes

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -39,6 +39,7 @@ requirements:
     - cil-astra=21.2.*
     - ccpi-regulariser=20.9
     - tomophantom=1.4.*
+    - algotom=1.0.*
 
 build:
   number: 1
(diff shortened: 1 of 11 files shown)


samtygier-stfc:
This might be tricky. Building docs requires importing all the modules in mantidimaging. To get the old doc releases sphinx-multiversion checks out each old release, and builds those docs. So to build the docs we need an environment with that can import all the old versions of mantidimaging. Removing sarepy breaks this.

Quick solution is to leave sarepy in the environment. Long term we might need a way to have multiversioned docs without having to rebuild them each time.

DolicaAkelloEgwel (approved):

Issue #1185, opened 2021-11-04
Bad data indicator overlays


samtygier-stfc:
When mousing over the bad data indicators, display an overlay showing the problem pixels.

PR #1188, opened 2021-11-08
Remove old overlay code

### LINKED ISSUE #1187: Remove old overlay code
Use the new overlays everywhere that they are needed.

Clean up from #1186

samtygier-stfc:
### Issue

Closes #1187

### Description

Remove old overlay code from operations and recon windows.

In stack choice dialog, enable the non-positive check if the current operation uses it.

When adding a new check, run it immediately (and handle the case where there is not an image yet)

Load a dataset with NaNs, the indicators should show in the main, operations and recon windows, overlays should show when mousing over.

Load the IMAT00010675 flower dataset, in flat fielding the yellow indicator should show and overlay should highlight the soil region. With safe apply on, apply to stack. Indicator should work here too. 

image

### Documentation

Release notes from #1185 ok

### DIFF
--- a/mantidimaging/gui/widgets/bad_data_overlay/bad_data_overlay.py
+++ b/mantidimaging/gui/widgets/bad_data_overlay/bad_data_overlay.py
@@ -2,7 +2,7 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 
 from collections import OrderedDict
-from typing import Callable, List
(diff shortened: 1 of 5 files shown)


samtygier-stfc:
Rebased and updated applitools, so should pass tests this time.

DolicaAkelloEgwel (approved):

PR #1192, opened 2021-11-09
1178: Operation argument opens wrong operation

### LINKED ISSUE #1178: --operation argument sometimes opens wrong operation
### Summary

`python -m mantidimaging --path '/data/imaging/IMAT00010675/Tomo'  --operation rebin`

Opens NaN Removal

### Steps To Reproduce

Start mantid imaging with a `--operation` argument, check it opens the correct operation filter.

### Expected Behaviour

Opens right filter

### Current Behaviour

Out by one. If below separator.

### Context

Current main.

Reminds me of #820

DolicaAkelloEgwel:
### Issue

Closes #1178

### Description

Fixes the issue of the index ignoring the dividers.

### Testing 

Added two tests for the new method.

### Acceptance Criteria 

Check that the tests pass. Check that using an operation argument gives the right result.

### Documentation

Updated release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -17,6 +17,7 @@ Fixes
 
 - #1138 : Improve version number handling
 - #1134 : NeXus Loader: OSError
+- #1178 : 
 
 
 Developer Changes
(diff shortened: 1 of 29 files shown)


samtygier-stfc:
Would it it work for `set_filter_by_name()` to use `setCurrentText()` instead of `setCurrentIndex()` to avoid needing to lookup the index?

samtygier-stfc (approved):

PR #1234, opened 2021-12-14
Change default save format to float32

### LINKED ISSUE #1200: Default save format should be float32
### Desired Behaviour
When saving a stack (File -> Save) the default bit depth should be float32.

### Current Behaviour

Currently the default is int16 (which means values have to be rescaled).

rbauststfc:
### Issue

Closes #1200

### Description

Changes the order of the options in the pixel bit depth dropdown so that float32 is at index 0 (the default selection). I changed the option order rather than the default index so that the dropdown menu will still open downwards if clicked.
Note that changing the option order seems to have automatically changed the dialog's width and height settings.

I've also added a couple of new screenshot tests for the save dialog - one for opening the dialog without any images selected (not actually possible via the UI I don't think, but the code tries to handle this scenario just in case). The second test is for opening with images selected, which worked either with mocking or by actually loading data, so I've gone for the first option here but happy to change if actually loading the data is preferred.

All tests pass.
Loading a dataset and then opening the save dialog shows the float32 option selected by default in the pixel bit depth dropdown.
Saving using either option applies the correct pixel bit depth to the saved images.

### Documentation

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -16,6 +16,7 @@ New Features
 - #1205 : Linearisation correction for beam hardening
 - #1197 : Dataset Tree View: Delete image stack
 - #1197 : Dataset Tree View : Show tab from treeview
+- #1200 : Default save option set to float32
 
 Fixes
 -----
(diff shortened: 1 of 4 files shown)


samtygier-stfc (approved):
Looks good and working well.

PR #1241, opened 2021-12-16
UI improvements to various input default settings

### LINKED ISSUE #1235: UI improvements
### Operations window

Remove outliers
Difference should have a range 0 - 10,000

ROI Normalisation
Normalisation mode: default should be Stack Average

Clip
All boxes, step size should be 0.1

Ring removal
Threshold, Threshold Min, Threshold Max, step size should be 0.1

Stripe filters
All the float / QDoubleSpinBox inputs.

Note:
For the step size changes, `add_property_to_form()` does not have a way to set this. `ArithmeticFilter` shows how to set it with `setSingleStep`.  I think it makes sense for `add_property_to_form()` to default to 0.1 step size for any float input not set manually.

### Recon window
Beam hardening correction
Show 4 decimal places, stepsize 0.01

rbauststfc:
### Issue

Closes #1235

### Description

Changes to a few input settings and defaults, as described on the issue. In particular, `float` spin boxes in the Operations window will now default to a step size of 0.1 unless specified otherwise, which I think was the suggestion on the issue. I've implemented it so that `int` spin boxes could now also have a step size specified, but this isn't currently in use and the default will be the `QSpinBox` default.

The Difference spin box for the Remove Outliers filter now has a range of 0 - 10000. The spin boxes are actually set to render with this range by default, but I've continued passing it in as a parameter as I took from the issue that this range was needed specifically here.

Ensure that all tests pass.
Check the following in the UI:

_Operations window_
1)  Remove Outliers -> Difference field: range is 0 - 10000
2) ROI Normalisation -> Normalisation mode field: default selection is Stack Average
3) For all filters the spin box step sizes are 0.1 for float and 1 for int, other than those that are set differently (the different ones will be Arithmetic `Add` and `Subtract`, Rebin `Factor` and Median `Kernel Size`). Please shout if there are any where the step size seems like it will no longer be appropriate.

_Recon window_
1) BHC tab inputs all have 4 decimal places and step size is 0.01

### Documentation

Issue details added to release notes

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -17,6 +17,7 @@ New Features
 - #1197 : Dataset Tree View: Delete image stack
 - #1197 : Dataset Tree View : Show tab from treeview
 - #1200 : Default save option set to float32
+- #1235 : Improvements to input value defaults and spin box step sizes
 
 Fixes
 -----
--- a/mantidimaging/core/operations/arithmetic/arithmetic.py
+++ b/mantidimaging/core/operations/arithmetic/arithmetic.py
@@ -86,18 +86,18 @@ class ArithmeticFilter(BaseFilter):
                                                    on_change=on_change,
                                                    default_value=0.0,
                                                    valid_values=(-MAX_SPIN_BOX, MAX_SPIN_BOX),
-                                                   tooltip="The add value.")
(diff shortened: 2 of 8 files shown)


samtygier-stfc (approved):
Looks good. Thanks.

PR #1246, opened 2022-01-06
Fix for AttributeError when click recon auto colour button

### LINKED ISSUE #1226: Recon auto color button : AttributeError
### Summary

Auto Change Colour Palette button in recon window gives
AttributeError: 'ImageItem' object has no attribute 'image_item'

### Steps To Reproduce

Open recon window. Press Auto Change Colour Palette button 

### Expected Behaviour

Auto color dialog.

### Current Behaviour
Error message

### Context

Current main

### Failure Logs

```
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/recon/view.py", line 470, in on_change_colour_palette
    change_colour_palette = PaletteChangerView(self, self.image_view.recon_hist, self.image_view.recon.image_item,
AttributeError: 'ImageItem' object has no attribute 'image_item'
```

rbauststfc:
### Issue

Closes #1226

### Description

The error was being caused because we were passing `self.image_view.recon.image_item` into `PaletteChangerView`, however `self.image_view.recon` is of type `ImageItem`, which doesn't have an `image_item` attribute. The `image` parameter of `PaletteChangerView` requires image data in the form of a numpy array, so the fix here is to pass `self.image_view.recon.image` instead.

I've also added a new screenshot test to check that the auto colour dialog displays correctly when the button is clicked.

All tests pass.
Opening the recon window and clicking the Auto Change Colour Palette button displays the dialog without any errors.

### Documentation

Added issue number to release notes.


samtygier-stfc (approved):
Looks good

PR #1249, opened 2022-01-10
Change ordering of axis linking in MIMiniImageView

### LINKED ISSUE #1236: Aspect ratio not preserved in COR insepctor or operations window
### Summary

image

### Steps To Reproduce

Load dataset
In Recon window press Minimise Error, press ok
Choose a slice in the COR table
Click Refine
Resize window

### Expected Behaviour

Aspect ratio should always be preserved.

### Current Behaviour
Resizing causes images to be stretched. Any pan or zoom operation resets them

### Context
Current main. Probably since the ImageView changes 9753383e04ea505f2a5374202146140fd8e982b7

I think it is still necessary to disable `setAspectLocked` on all but the first image, but maybe can do something when the resize is complete.

samtygier-stfc:
View will pull range from its linked view. Need range to be pulled from the first sibling which has the fixed aspect ratio.

### Issue
Closes #1236 

### Description

9753383 is still needed to prevent run away zooming when resizing linked views.

Fixes is to invert the links.
`vb1.linkView(ViewBox.XAxis, vb2)`
means that v1 will pull the range from vb2. So if vb1 has a fixed aspect but vb2 does not, then vb2 can end up with an odd aspect. 

See https://github.com/pyqtgraph/pyqtgraph/issues/1348

This might cause tiny changes that trigger applitools.

Check that aspect ratio is preserved in operations, recon, and COR refine (in recon, click `add` then `refine`).

Check that there is no weird zooming behaviour when resizing any of those windows. 

### Documentation

release_notes

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -32,6 +32,7 @@ Fixes
 - #1174 : TypeError in link_before_after_histogram_scales
 - #1161 : Remove run_ring_removal argument from Ring Removal operation and always run filter
 - #1226 : AttributeError when trying to open reconstruction auto colour dialog
+- #1236 : Aspect ratio not preserved in COR inspector or operations window
 
 
 Developer Changes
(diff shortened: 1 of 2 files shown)


rbauststfc (approved):
This is working well for me. Nice spot on the solution.

PR #1274, opened 2022-01-27
1190: Put recons in tree view group

### LINKED ISSUE #1990: Load TIFF Container into MI
### Desired Behaviour

As an user on IMAT, i would like to load a Tiff container directly into mantid imaging which is then unpacked in MI. 

This feature would allow me to skip the step of first opening up the container within ImageJ and unpacking it to then load into MI.

### Current Behaviour

Loading a Tiff container (e.g. 4000, 512, 515) is not possible. MI will return the following error in a dialog box after trying to load the selected Tiff container into MI.
image

DolicaAkelloEgwel:
### Issue

Closes #1990

### Description

Creates a recon group in the dataset tree view. Adds to it when more recons are created. Also did some tidying of the `MainWindowPresenter`. Deleting recons probably doesn't work for the tree view yet...

### Testing 

Added/changed a bunch of tests. 

### Acceptance Criteria 

Check that the tests pass. Check that a recon group gets added to the tree view. Check that different datasets have different recon groups.

### Documentation

Updated release notes.

### DIFF
--- a/mantidimaging/gui/windows/stack_visualiser/presenter.py
+++ b/mantidimaging/gui/windows/stack_visualiser/presenter.py
@@ -105,27 +105,28 @@ class StackVisualiserPresenter(BasePresenter):
             new_stack = self.images.copy(flip_axes=True)
             new_stack.name = self.images.name + "_sino"
             new_stack.record_operation(const.OPERATION_NAME_AXES_SWAP, display_name="Axes Swapped")
-            self.view.parent_create_stack(new_stack)
-            new_dataset = MixedDataset([new_stack], new_stack.name)
-            self.view._main_window.presenter.model.add_dataset_to_model(new_dataset)
+            self.add_mixed_dataset_to_model_and_update_view(new_stack)
 
     def dupe_stack(self):
         with operation_in_progress("Copying data, this may take a while",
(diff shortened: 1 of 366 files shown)


mantidimaging/gui/windows/stack_visualiser/presenter.py
```diff
@@ -106,7 +106,7 @@ def create_swapped_axis_stack(self):
             new_stack.name = self.images.name + "_sino"
             new_stack.record_operation(const.OPERATION_NAME_AXES_SWAP, display_name="Axes Swapped")
             self.view.parent_create_stack(new_stack)
-            new_dataset = StackDataset([new_stack], new_stack.name)
+            new_dataset = MixedDataset([new_stack], new_stack.name)
             self.view._main_window.presenter.model.add_dataset_to_model(new_dataset)
```
samtygier-stfc, line 110: Is something extra needed here now. Currently ( b0ff207bc ) if I load a dataset and do Image -> Create sinograms from stack. It makes the tab, and adds the stack to the model (I can see it in the operations window stack list), but it does not get an entry in the tree view.

samtygier-stfc (approved):

PR #1284, opened 2022-01-31
When using a slice from stack as 180, make sidebar entry and tabs

### LINKED ISSUE #1283: No tab or treeview entry for 180
### Summary

When the 180 projection is not a separate file (when it is automatically selected from the projection stack), then no treeview entry or tab is created.

### Steps To Reproduce

Open a dataset that does not have a 180 file. Or a nexus file.

### Expected Behaviour

180 treeview entry and tab is created

### Current Behaviour

The 180 is loaded, it can be seen in the stack selector in the operations view.

### Context

Current main

samtygier-stfc:
### Issue
Closes  #1283

### Description

In add_alternative_180_if_required() the new stack should be added to the dataset treeview and have a tab created.

Let get_stack_visualiser() search all visualiser, not just visible ones.

Add some mocking in the tests as these reach quite a way into the view.

Steps on #1283

### Documentation
Fix for a recent so does not need to be on release notes.

### DIFF
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -202,6 +202,10 @@ class MainWindowPresenter(BasePresenter):
                 dataset.proj180deg = Images(np.reshape(closest_projection, (1, ) + closest_projection.shape),
                                             name=f"{dataset.name}_180")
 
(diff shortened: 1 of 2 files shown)


samtygier-stfc:
Good spot, could you file a new issue for that.

rbauststfc (approved):

PR #1287, opened 2022-02-01
Add checkbox to allow optional live reconstruction

### LINKED ISSUE #1279: Add option to turn off live reconstruction
### Desired Behaviour

When trialling TV parameters it would be good to have an option to turn off live reconstruction.

### Current Behaviour

Live reconstruction is good for FBP but it is tedious for TV because every change takes 20 sec or longer.

rbauststfc:
### Issue

Closes #1279

### Description

Adds an auto-update checkbox to the recon window to allow the user to control whether or not a live reconstruction preview is provided. Unchecking the checkbox stops the live preview. Checking the checkbox triggers a preview to bring things up to date and, while the box remains checked, further parameter changes are previewed live.

Also adds an Update Now button, as described in the comments.

Screenshot of checkbox and button on the recon window:

ReconAutoUpdate

All tests pass (a couple of new tests have been added for the functionality).

Load a dataset and open the reconstruction window. The checkbox should be checked by default and any parameter changes should be previewed live.
Uncheck the checkbox and make further changes to parameters - the preview should not be updated, but any sinogram changes should render.
Check the checkbox - the preview should update straight away. Further changes to parameters should also now preview live.

Test Update Now button refreshes the preview when auto update is disabled.

### Documentation

Issue added to release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -20,6 +20,7 @@ New Features
 - #1235 : Improvements to input value defaults and spin box step sizes
 - #1230 : Update CIL to 21.3 reducing memory requirements
 - #1191 : Loading 180 should only show datasets
+- #1279 : Add auto update checkbox option to turn off live reconstruction
 
 
 Fixes
--- a/mantidimaging/gui/test/test_gui_system_operations.py
+++ b/mantidimaging/gui/test/test_gui_system_operations.py
@@ -137,3 +137,10 @@ class TestGuiSystemOperations(GuiSystemBase):
             self._wait_until(lambda: self.op_window.presenter.filter_is_running is False)
             self.main_window.filters.close()
             QTest.qWait(SHOW_DELAY)
+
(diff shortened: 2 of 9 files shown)


samtygier-stfc:
Could you also add an "Update now" buttons, similar to in the operations window.

rbauststfc:
> Could you also add an "Update now" buttons, similar to in the operations window.

I noticed that in the operations window it looks like selecting the checkbox doesn't trigger the preview to update, so you have to use the "Update Now" button. I wasn't sure whether to include it here because I've set the Auto Update checkbox to trigger an update when you select it again. Are you thinking users will want to turn live recon off, tweak parameters and then use the button rather than turning the update back on again when they're ready? That would make sense if they're going to go through several rounds of changing parameters in batches and want to see changes intermittently as they go along.

samtygier-stfc:
>  That would make sense if they're going to go through several rounds of changing parameters in batches and want to see changes intermittently as they go along.

Yes, I think thats a useful workflow. Tweak parameters, run a preview, tweak some more, run another preview.

samtygier-stfc (approved):

PR #1296, opened 2022-02-03
1273: Move sinograms into dataset

### LINKED ISSUE #1273: Move sinograms into the Dataset
### Desired Behaviour
Sinograms should be part of the dataset

### Current Behaviour

Currently making sinograms creates a new StackDataset

DolicaAkelloEgwel:
### Issue

Closes #1273

### Description

Gives the datasets a sinogram attribute. The sinogram also has an entry in the tree view. Datasets can have one sinogram so the previous one is replaced when the method is called again. Sinograms can also be deleted with the tree view.

### Testing 

Wrote some tests for main window model/presenter/view and dataset.

### Acceptance Criteria 

Check that the tests pass. Check that creating a sinogram updates the tree view. Check that double-clicking item on tree view changes stack. Check that creating another set of sinograms replaces the previous one.

### Documentation

Updated release notes.

### DIFF
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -585,6 +585,10 @@ class MainWindowPresenter(BasePresenter):
         for i in range(top_level_item_count):
             top_level_item = self.view.dataset_tree_widget.topLevelItem(i)
             if top_level_item.id == parent_id:
-                self.view.create_child_tree_item(top_level_item, sino_id, "Sinograms")
+                sinograms_item = self.view.get_sinograms_item(top_level_item)
+                if sinograms_item is None:
+                    self.view.create_child_tree_item(top_level_item, sino_id, self.view.sino_text)
+                else:
+                    sinograms_item._id = sino_id
                 return
         raise RuntimeError(f"Unable to add sinogram item to dataset tree item with ID {parent_id}")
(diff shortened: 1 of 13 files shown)


mantidimaging/gui/windows/main/presenter.py
```diff
@@ -563,3 +553,31 @@ def _add_recon_to_dataset(self, recon_data: Images, stack_id: uuid.UUID) -> None
...
         self.add_recon_item_to_tree_view(parent_id, recon_data.id, len(self.model.datasets[parent_id].recons))
         self.view.model_changed.emit()
+
+    def add_sinograms_to_dataset_and_update_view(self, sino_stack: Images, original_stack_id: uuid.UUID):
+        """
+        Adds sinograms to a dataset or replaces an existing one.
+        :param sino_stack: The sinogram stack.
+        :param original_stack_id: The ID of a stack in the dataset.
+        """
+        parent_id = self.model.get_parent_strict_dataset(original_stack_id)
```
samtygier-stfc, line 563: Can we make this work with the mixed dataset as well.

Use case would be loading some projections using "File->Load Images" and then creating sinograms. The sinograms would be added to the same MixedDataset.

samtygier-stfc (changes requested):

samtygier-stfc (approved):

PR #1302, opened 2022-02-07
Fix attribute errors when applying filters after closing and re-opening ops window

### LINKED ISSUE #1292: NoneType object has no attribute applyButton

rbauststfc:
### Issue

Closes #1292

### Description

The attribute error was because method `_set_apply_buttons_enabled` was being triggered by a signal from the main window (which itself was triggered because of the call to `self.view.filter_applied.emit()` in the presenter's `_post_filter` method). When the Operations window was closed this connection between the old presenter and the main window wasn't getting cleaned up, so subsequent filters were still triggering calls to the old presenter, which no longer had a view associated with it.

As both the `_post_filter` and `_set_apply_buttons_enabled` methods are located in the presenter, I've refactored so that we now call `_set_apply_buttons_enabled` directly from `_post_filter` rather than using signals.

Following the steps described on the issue results in filters being applied without any errors.

### Documentation

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -42,6 +42,7 @@ Fixes
 - #1190 : Add recon item to right part of tree view
 - #1239 : 180 warning when applying ring removal to recon
 - #1297 : Auto colour dialog not working in compare images
+- #1292 : Attribute error if apply filter after closing and re-opening Operations window
 
 
 Developer Changes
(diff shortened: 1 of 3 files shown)


samtygier-stfc:
I think you can also remove the self.main_window.filter_applied mock in FiltersWindowPresenterTest

samtygier-stfc (approved):

Issue #1299, opened 2022-02-04
Encountered error while trying to reconstruct: list index out of range


samtygier-stfc:
### Summary

Error reconstructing sinograms with CIL

### Steps To Reproduce

Open dataset
Images -> Create sinogram from stack
Open recon window
Select the sinogram stack
Click use COR/Tilt from above
On the reconstruct tab, choose CIL
Click Reconstruct volume

Note that reconstructing with FBP works fine

### Expected Behaviour

Successful reconstruction

### Current Behaviour

"Encountered error while trying to reconstruct: list index out of range"

or

"Encountered error while trying to reconstruct: Cannot fill with the provided array.Expecting (81, 81, 362) got (256, 81, 362)"

I think these 2 error messages relate to the same problem of the wrong axes being chosen. Probably to do with if the number angles is bigger or smaller than the other dimension.

### Context

Current main

### Failure Logs

```
  File "/home/sam/git/mantidimaging/mantidimaging/gui/dialogs/async_task/task.py", line 49, in run
    self.result = call_with_known_parameters(self.task_function, **self.kwargs)
  File "/home/sam/git/mantidimaging/mantidimaging/core/utility/func_call.py", line 11, in call_with_known_parameters
    return func(**ka)
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/recon/model.py", line 154, in run_full_recon
    recon = reconstructor.full(images, self.data_model.get_all_cors_from_regression(images.height), recon_params,
  File "/home/sam/git/mantidimaging/mantidimaging/core/reconstruct/cil_recon.py", line 171, in full
    rot_pos = [(cors[pixel_num_v // 2].value - pixel_num_h / 2) * pixel_size, 0, 0]
IndexError: list index out of range
```
or
```
  File "/home/sam/git/mantidimaging/mantidimaging/gui/dialogs/async_task/task.py", line 49, in run
    self.result = call_with_known_parameters(self.task_function, **self.kwargs)
  File "/home/sam/git/mantidimaging/mantidimaging/core/utility/func_call.py", line 11, in call_with_known_parameters
    return func(**ka)
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/recon/model.py", line 154, in run_full_recon
    recon = reconstructor.full(images, self.data_model.get_all_cors_from_regression(images.height), recon_params,
  File "/home/sam/git/mantidimaging/mantidimaging/core/reconstruct/cil_recon.py", line 183, in full
[...]

PR #1304, opened 2022-02-07
Trigger recon redraws when stack is modified

### LINKED ISSUE #1289: Mark as projections/sinograms does not trigger recon redraw
### Summary

MI allows a stack to be marked as either projections or sinograms. The reconstruction window can correctly use either.

If sinograms are loaded from "Load Images" they must be manually marked as sinograms, otherwise the sinogram space data will be shown in the top projection preview. The reconstruction window is not automatically updated when the stack is remarked.

### Steps To Reproduce

Load some sinograms with "File -> Load Images"
Open the reconstruction window, not that the sinograms are shown at the top
In the main window, "Image -> Mark as projection/sinograms", choose sinograms.
Note that the reconstruction window is unchanged.
In the reconstruction window, change the Projection or Slice index. These will force a redraw, and the sinograms should be shown in the correct place

image
vs
image

### Expected Behaviour

Marking as projection or sinogram should trigger a redraw.

### Current Behaviour
No auto redraw

### Context
Current main

samtygier-stfc:
### Issue

Closes #1289

### Description

If a stack is modified while the reconstruction window is open, then the views and preview should be updated.

This covers the case for marking a stack as sinograms/projections and also the common case of applying operations to a stack.

Steps in the report.

Also, try opening the recon window, and then in the operations window applying a operation such as guassian.

### Documentation

release notes

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -44,6 +44,7 @@ Fixes
 - #1297 : Auto colour dialog not working in compare images
 - #1292 : Attribute error if apply filter after closing and re-opening Operations window
 - #1294 : Dataset Tree View: Allow for deleting recons
+- #1289 : Trigger recon redraws when stack is modified
 
 
 Developer Changes
--- a/mantidimaging/gui/windows/main/view.py
+++ b/mantidimaging/gui/windows/main/view.py
@@ -56,9 +56,10 @@ class MainWindowView(BaseMainWindowView):
     NOT_THE_LATEST_VERSION = "This is not the latest version"
     UNCAUGHT_EXCEPTION = "Uncaught exception"
 
+    # Emitted when a new stack is created or an existing one deleted
     model_changed = pyqtSignal()
-    filter_applied = pyqtSignal()
-    recon_applied = pyqtSignal()
+    # Emitted when an existing stack is changed
+    stack_changed = pyqtSignal()
     backend_message = pyqtSignal(bytes)
 
     menuFile: QMenu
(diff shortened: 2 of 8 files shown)


rbauststfc:
These changes look good to me. The only thing was I tested with sinograms of the flowers dataset and I got some index out of bounds exceptions - I think similar to what's described in issue #1299. Presumably this PR can go in and the exceptions can be looked at under the other issue, though?

samtygier-stfc:
Do you just see the out of bounds error when actually running the reconstruction (i.e. click the Reconstruct Slice or Reconstruct Volume buttons)?

rbauststfc:
No, it's when the preview automatically updates. This was what I got:

```
  File "/home/rbaust/mantidimaging/mantidimaging/gui/windows/recon/presenter.py", line 168, in handle_stack_changed
  File "/home/rbaust/mantidimaging/mantidimaging/gui/windows/recon/presenter.py", line 228, in do_preview_reconstruct_slice
    self.view.update_sinogram(self.model.images.sino(slice_idx))
  File "/home/rbaust/mantidimaging/mantidimaging/core/data/images.py", line 206, in sino
    return self.data[slice_idx]
IndexError: index 571 is out of bounds for axis 0 with size 512
```

samtygier-stfc:
Ok, I think that it separate to #1299 .  I'll have a look.

samtygier-stfc:
The error was because if the dimensions of image changes then the selected slice number may no longer be in range. The quick solution is to always reset the cor model, as if we had switched to a new stack. This is forces everything to be in range.

rbauststfc (approved):

PR #1319, opened 2022-02-17
1295: Failed recon increments counter

### LINKED ISSUE #1295: Failed recon still increments recon counter
### Summary

Failed recons still increment the recon counter.

### Steps To Reproduce

Make a recon. Make another recon with bad parameters (I do this by using default settings then choosing the none filter).

### Expected Behaviour

First recon should be called "Recon," next successful recon should be called "Recon_2, etc.

### Current Behaviour

First recon is called "Recon." Failed recon doesn't work yet next successful recon is named "Recon_3."

### Context

Current master.

DolicaAkelloEgwel:
### Issue

Closes #1295

### Description

Replaces infs with NaNs in recons. This prevents the error caused by inf arguments to the pyqtgraph histogram.

### Testing 

Added tests that check the add data to view method is called with data that doesn't contain infs.

### Acceptance Criteria 

Check that the tests pass. Check that the error mentioned in #1295 doesn't happen.

### Documentation

Updated release notes.

### DIFF
--- a/mantidimaging/gui/windows/recon/presenter.py
+++ b/mantidimaging/gui/windows/recon/presenter.py
@@ -6,6 +6,7 @@ from enum import Enum, auto
 from logging import getLogger
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Callable, Set
 
+import numpy as np
 from PyQt5.QtWidgets import QWidget
 
 from mantidimaging.core.data import Images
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/windows/recon/presenter.py
```diff
@@ -304,6 +306,7 @@ def _on_volume_recon_done(self, task):
             self.view.show_error_dialog(f"Encountered error while trying to reconstruct: {str(task.error)}")
             return
 
+        self._replace_inf_nan(task.result)
```
samtygier-stfc, line 309: Can you add a comment here that this is needed because pyqtgraph has problems with arrays containing inf

samtygier-stfc (changes requested):

samtygier-stfc (approved):

PR #1335, opened 2022-02-24
1332: Fix for non-standard names

### LINKED ISSUE #1322: FiltersWindowPresenter._post_filter: attempt_repair
Investigate what cases this is triggered and if it does anything useful still.

If not needed then set_images_in_stack() can probably be removed.

DolicaAkelloEgwel:
### Issue

Closes #1322

### Description

Allows flat/dark images to be recognised if they don't have "Flat" or "Dark" in their names, provided a filename is given.

### Testing 

Wrote a test for checking the second call to `find_images`.

### Acceptance Criteria 

Check that the tests pass. Check that the other images are recognised. Not so sure if this works all the time or with just the test images...

### Documentation

Updated release notes.


mantidimaging/gui/windows/load_dialog/presenter.py
```diff
@@ -123,7 +123,11 @@ def do_update_flat_or_dark(self, field: Field, name: str, suffix: str):
         if not selected_file:
             return
         selected_dir = Path(os.path.dirname(selected_file))
-        field.set_images(find_images(selected_dir, name, suffix, image_format=self.image_format, logger=logger))
+        images = find_images(selected_dir, name, suffix, image_format=self.image_format, logger=logger)
+        if not images:
+            base_name = os.path.basename(selected_file).split("_")[0]
```
samtygier-stfc, line 128: Can you change this to split on the last `_`
`os.path.basename(selected_file).rpartition("_")[0]`

This means that if the filename is `spheres_bbb_0000.tiff` then it won't match `spheres_ccc_0000.tiff`

(I added a strange_names2 file, currently when I select `spheres_bbb_0000.tiff` for Flat Before, it then matches all 409 images)

samtygier-stfc (changes requested):

samtygier-stfc (approved):
Working well. Thanks

PR #1357, opened 2022-03-07
Use mixin class for auto colour menu

### LINKED ISSUE #1317: Unify auto color between MIMiniImageView and MIImageView
### Desired Behaviour

Move the code for Auto option in the menu into a mixin classs that can be shared between the image view classes.

Maybe use the `histogram_siblings` for setting the auto colour on neighbours.

rbauststfc:
### Issue

Closes #1317

### Description

Moves the code for creating and connecting the auto colour right click menu option to a mixin class to avoid duplicating it in the MIImageView and MIMiniImageView classes.

For the MIMiniImageView class, I've used the `axis_siblings` to find the other histograms (for making sure colour palette changes are applied to all three images), as not all image views are added to the `histogram_siblings` in the Operations window.

1) The auto colour menu option works as expected for all relevant windows. For the Operations and Reconstruction windows, changing the value for one histogram should apply the change to all.

2) Check that the auto colour palette button in the Reconstruction window is still working as expected.

3) In the Operations window, check that the "Auto" menu option enables and disables correctly depending on if there is a difference image.

### Documentation

Issue added to release notes.

### DIFF
--- a/mantidimaging/eyes_tests/reconstruct_window_test.py
+++ b/mantidimaging/eyes_tests/reconstruct_window_test.py
@@ -61,7 +61,7 @@ class ReconstructionWindowTest(BaseEyesTest):
         self._load_data_set()
 
         self.imaging.show_recon_window()
-        self.imaging.recon.image_view.imageview_recon.im.image = np.empty((512, 512))
+        self.imaging.recon.image_view.imageview_recon.setImage(np.zeros((512, 512)))
         self.imaging.recon.changeColourPaletteButton.click()
 
         self.check_target(widget=self.imaging.recon.change_colour_palette_dialog)
--- a/mantidimaging/eyes_tests/reconstruct_window_test.py
+++ b/mantidimaging/eyes_tests/reconstruct_window_test.py
@@ -61,7 +61,7 @@ class ReconstructionWindowTest(BaseEyesTest):
         self._load_data_set()
 
         self.imaging.show_recon_window()
-        self.imaging.recon.image_view.recon.image = np.empty((512, 512))
+        self.imaging.recon.image_view.imageview_recon.im.image = np.empty((512, 512))
         self.imaging.recon.changeColourPaletteButton.click()
 
         self.check_target(widget=self.imaging.recon.change_colour_palette_dialog)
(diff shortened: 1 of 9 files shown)


mantidimaging/gui/widgets/mi_image_view/view.py
```diff
@@ -104,16 +108,19 @@ def __init__(self,
...
-        action = self.ui.histogram.item.gradient.menu.actions()[12]
-        self.ui.histogram.item.gradient.menu.insertAction(action, self.auto_colour_action)
-        self.ui.histogram.item.gradient.menu.insertSeparator(self.auto_colour_action)
+        self.add_auto_color_menu_action(self)
 
         # Work around for https://github.com/mantidproject/mantidimaging/issues/565
         self.scene.contextMenu = [item for item in self.scene.contextMenu if "export" not in item.text().lower()]
 
+    @property
+    def main_histogram(self) -> 'HistogramLUTItem':
```
samtygier-stfc, line 117: Maybe just call this histogram. It is the only histogram of the image view.

mantidimaging/gui/windows/operations/filter_previews.py
```diff
@@ -199,22 +197,3 @@ def set_histogram_log_scale(self):
         """
         set_histogram_log_scale(self.image_before_hist)
         set_histogram_log_scale(self.image_after_hist)
-
-    def _add_auto_colour_action(self, img_view: MIMiniImageView):
```
samtygier-stfc, line 203: I there are a few things that are not used and can now be cleaned up.
self.all_histograms

Also now is probably a good time to remove the 
`self.image_before, self.image_before_vb, self.image_before_hist = self.imageview_before.get_parts()`
These were to ease the transition to MIMiniImageView. Now that there are only a few places that directly access these, they can used the MIMiniImageView accessors.

e.g. use of`self.image_after_vb` would become `self.imageview_after.viewbox`

Same changes can be made in `gui/windows/recon/image_view.py`

Then get_parts can be removed from MIMiniImageView.

samtygier-stfc (changes requested):

mantidimaging/eyes_tests/reconstruct_window_test.py
```diff
@@ -61,7 +61,7 @@ def test_reconstruction_window_colour_palette_dialog(self):
         self._load_data_set()
 
         self.imaging.show_recon_window()
-        self.imaging.recon.image_view.recon.image = np.empty((512, 512))
+        self.imaging.recon.image_view.imageview_recon.im.image = np.empty((512, 512))
```
samtygier-stfc, line 64: I think that can be
`self.imaging.recon.image_view.setImage(np.empty((512, 512)))`

Also should probably be `np.zeros` rather than `np.empty`.

samtygier-stfc (approved):

PR #1375, opened 2022-03-16
Add LeakTracker

### LINKED ISSUE #1374: Leak checking tool
### Desired Behaviour

A tool for tracking `leaks` in MI.

Most of MI's memory usage is the large arrays used for image stack storage. Python will keep these in RAM as long as there is a reference pointing to them. Even once a dataset has been closed by the user, it is possible some references remain.

Add a tool that uses weakreferences to keep track of ImageStacks, and can report about still alive instances.

samtygier-stfc:
### Issue

Closes #1374 

### Description

Class for tracking references to objects. Uses weak references to avoid keeping the objects alive. Then the leak_tracker can be queried later to see if the objects are still alive.

Use the tracker for the numpy array in `ImageStack`

Show leaked objects in the system test `tearDown` method

In the system test run leaked ImageStacks should be reported, e.g.

`xvfb-run --auto-servernum  pytest -vs -rs --run-system-tests -k test_run_operation_stack_safe_0`

If you change the `gui_system_base.py` line 50 to
`leak_tracker.pretty_print(debug_init=True, debug_owners=True, trace_depth=5)`

Then some extra debug information will be shown

### Documentation

Release notes updated

### DIFF
--- a/mantidimaging/core/utility/leak_tracker.py
+++ b/mantidimaging/core/utility/leak_tracker.py
@@ -1,0 +1,112 @@
+# Copyright (C) 2022 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+import gc
+import sys
+import traceback
+import weakref
+from typing import NamedTuple, List, Set, Iterable
+
+from numpy import ndarray
+
+
+class ItemInfo(NamedTuple):
+    ref: weakref.ref
+    msg: str
+    created: List[str]
+
+
+def obj_to_string(obj, relative=None) -> str:
+    """Shows object type, id and extra info for some classes"""
+    extra_info = ""
+    if isinstance(obj, dict) and relative is not None:
+        keys = {k for k, v in obj.items() if v is relative}
+        extra_info += f" keys={keys}"
+
+    try:
+        if hasattr(obj, "name"):
+            extra_info += f" obj.name={obj.name}"
+
+        if hasattr(obj, "id"):
+            extra_info += f" obj.id={obj.id}"
+    except RuntimeError as e:
+        # Can happen with a deleted QWidget
+        extra_info += str(e)
+
+    if isinstance(obj, ndarray):
+        extra_info += f" ndarray: {obj.shape} {obj.dtype}"
+
+    return f"{type(obj)} pyid={id(obj)} {extra_info}"
+
+
+def find_owners(obj, depth: int, path: List[str] = None, ignore: Set[int] = None) -> List[List[str]]:
+    """Recursively track though references to objects and return a list of routes"""
+    all_routes = []
+    if path is None:
+        path = [obj_to_string(obj)]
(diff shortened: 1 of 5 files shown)


mantidimaging/gui/test/gui_system_base.py
```diff
@@ -44,6 +45,10 @@ def tearDown(self) -> None:
         QTest.qWait(SHORT_DELAY)
         self.assertDictEqual(self.main_window.presenter.model.datasets, {})
 
+        if leak_count := leak_tracker.count():
+            print("\nItems still alive:", leak_count)
+            leak_tracker.pretty_print(debug_init=False, debug_owners=False, trace_depth=5)
```
rbauststfc, line 50: This looks like it's just passing in defaults, would it work to simplify to `leak_tracker.pretty_print()`?

mantidimaging/core/utility/leak_tracker.py
```diff
@@ -0,0 +1,113 @@
...
+        all_routes.append(path)
+    for owner in new_refs:
+        if type(owner).__name__ in ['frame', 'function', 'list_iterator']:
+            continue
+        route = [obj_to_string(owner, obj)] + path
+        ignore.add(id(route))
+        if depth > 0:
+            new_routes = find_owners(owner, depth - 1, route, ignore)
+            for new_route in new_routes:
+                all_routes.append(new_route)
```
rbauststfc, line 64: Would this also work if we used `all_routes.extend(new_routes)` instead of the `for` loop?

mantidimaging/core/utility/test/leak_tracker_test.py
```diff
@@ -0,0 +1,87 @@
...
+        ob1 = TestObject()
+        self.leak_tracker.add(ob1, "created_in_test_tracking")
+        container = TestContainer(ob1)  # noqa: F841
+
+        self.assertEqual(self.leak_tracker.count(), 1)
+        del ob1
+
+        # check that the object type and message are in the output
+        self.assertEqual(self.leak_tracker.count(), 1)
+        track_output = StringIO()
```
rbauststfc, line 69: These three sections that are checking the output messages look very similar - would it work to pull some of this into an internal `check_output` method, just to help with readability of the test?

rbauststfc (commented):
Changes look good. There are a couple of superficial tweaks I've suggested if you're interested, but not essential if you'd rather leave it as it is.

The only other thing to mention is that once we've switched to using multiprocessing.shared_memory instead of multiprocessing.Array I think we might need to be passing in the SharedMemory object rather than the numpy array, as it's the reference to the memory location that keeps things from being garbage collected with that.

mantidimaging/gui/test/gui_system_base.py
samtygier-stfc: I left them in to make it quick and easy to flick them on when debugging.

rbauststfc (approved):

PR #1380, opened 2022-03-23
Use small tag of mantidimaging-data to speed up tests

### LINKED ISSUE #1339: Speed up github actions - smaller test data
### Desired Behaviour

We currently use the 512x512x1142 flower whitebeam dataset from the mantidimaging-data repo. A small version would run faster and not reduce the value of the tests.

Need to consider the best place to host. Just adding to the mantidimaging-data repo would increase download size.

samtygier-stfc:
### Issue

Closes #1339

### Description

I have created a 128x128 branch of the flower dataset on mantidimaging-data. The change here is to use that instead of the full dataset. This reduces download time and speeds up gui test runtimes.

Compared to the last successful conda action run on a PR
Get test data:  47s -> 1s
GUI test system: 7:55 -> 3:23
GUI test screenshots: 1:48 -> 2:38 

Not sure why the screenshots got slower. Maybe dominated by time uploading to api?
Some more old screenshot times: 1:47, 1:48

I have updated all of the applitools baselines for the smaller image.

### Documentation

not needed

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -105,9 +105,9 @@ jobs:
       - name: Get test data
         shell: bash -l {0}
         run: |
-          wget -nv https://github.com/mantidproject/mantidimaging-data/archive/refs/heads/main.zip
-          unzip -q main.zip -d ~
-          mv ~/mantidimaging-data-main ~/mantidimaging-data
+          wget -nv https://github.com/mantidproject/mantidimaging-data/archive/refs/tags/small.zip
+          unzip -q small.zip -d ~
+          mv ~/mantidimaging-data-small ~/mantidimaging-data
         timeout-minutes: 5
 
       - name: GUI Tests System


samtygier-stfc:
Today, comparing with #1383 , I see almost identical times 2:06 vs 2:07. 

```
 	User time (seconds): 34.60
	System time (seconds): 1.96
	Percent of CPU this job got: 29%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 2:04.83
```
vs
```
	User time (seconds): 26.22
	System time (seconds): 1.41
	Percent of CPU this job got: 21%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 2:06.67
```
Assuming `user` is our actual CPU time, `system` is mostly disk IO, and the the gap to `wall clock` is network requests, then it looks like moving to smaller files saves CPU and disk, but the job is dominated by quite variable API speed.

rbauststfc (approved):

PR #1387, opened 2022-03-30
Fix async task leak

### LINKED ISSUE #1378: TaskWorkerThread ImageStack leak
### Summary

In the gui system tests an `ImageStack` is being leaked probably due to a reference in `TaskWorkerThread`.

The issue is probably, why is the `TaskWorkerThread` still alive.

### Steps To Reproduce

In `mantidimaging/gui/test/gui_system_base.py` change the leak tracker call to
`leak_tracker.pretty_print(debug_init=False, debug_owners=True, trace_depth=5)`

run 
`xvfb-run --auto-servernum  pytest -vs -rs --run-system-tests -k test_minimise`

### Current Behaviour
```
Items still alive: 44
 Object: <class 'numpy.ndarray'> pyid=139681282501104  ndarray: (1, 512, 512) float32
  item_info.msg='ImageStack untitled'
 Ownership:
	 <class 'mantidimaging.gui.dialogs.async_task.model.AsyncTaskDialogModel'> pyid=139681282527088 
	 <class 'dict'> pyid=139681283123008  keys={'task'}
	 <class 'mantidimaging.gui.dialogs.async_task.task.TaskWorkerThread'> pyid=139681282555968 
	 <class 'dict'> pyid=139681283122176  keys={'result'}
	 <class 'mantidimaging.core.data.imagestack.ImageStack'> pyid=139681282507776  obj.name=untitled obj.id=38915926-f953-41a5-8bdf-30f08c5a12bc
	 <class 'dict'> pyid=139681283123328  keys={'_data'}
	 <class 'numpy.ndarray'> pyid=139681282501104  ndarray: (1, 512, 512) float32

...
```

### Context
Current main

samtygier-stfc:
### Issue
Closes #1378

### Description

Use close instead of hide, because we are not going to reshow
    
Set parent to None, otherwise ReconstructWindowView holds a references to the child dialog
   
Drop the presenter and model references as well to ensure everything can be deleted. Use property to prevent access to the presenter after this.

Also, remove unused auto_close option

Note: I wanted to avoid WA_DeleteOnClose, because that can lead to a case where the python object still exists but the c++ object is gone, which can in turn lead to runtimeerrors

`xvfb-run --auto-servernum pytest -vs -rs --run-system-tests -k test_minimise`
should no longer report `Items still alive`

Test some operation and recon steps which show a progress bar. 

There is still 1 (or more) smaller leak remaining, but LeakTracker can't see their referrers. That will need some further investigation.

### Documentation

release notes

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -24,3 +24,4 @@ Developer Changes
 - #1148 : Re-enable coverage checking
 - #1375 : Leak tracking tool
 - #1376 : Fix leak in StackChoiceView
+- #1387 : Fix async task leak
--- a/mantidimaging/gui/dialogs/async_task/test/view_test.py
+++ b/mantidimaging/gui/dialogs/async_task/test/view_test.py
@@ -13,7 +13,7 @@ from mantidimaging.test_helpers import start_qapplication
 class AsyncTaskDialogViewTest(unittest.TestCase):
     @mock.patch('mantidimaging.gui.dialogs.async_task.view.AsyncTaskDialogPresenter')
     def setUp(self, mock_atd) -> None:
-        self.view = AsyncTaskDialogView(None, True)
+        self.view = AsyncTaskDialogView(None)
         self.view.infoText = mock.Mock()
         self.mock_qtimer = mock.Mock()
         self.view.show_timer = self.mock_qtimer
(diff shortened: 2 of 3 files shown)


samtygier-stfc:
This is not quite as clean as I'd hoped. We want to set some data members to None, to avoid references keeping things alive. We are doing this as cleanup when we are done with the instance. But mypy forces us to be strict, if we just set `self.presenter` to None, mypy requires that its type must be optional. That means that lots of places need guards. Its a genuine issue (good job mypy), because `show_from_timer` can be called after `handle_completion` if the task runs quickly. `self.show_timer.stop()` does not seem to stop the call, probably because the signal can sit in a queue for a while.

I have used a property to simplify the guard.

rbauststfc (approved):

Issue #1331, opened 2022-02-24
Make MI Windows compatible


rbauststfc:
Currently MI is only compatible with Linux. It would be good for it to run on Windows and potentially Mac too.

Following some initial investigation, the below is a suggested plan for the issue (see notes in the issue comments for more detail on the steps given):

1) Address the issues under [_Setting up the mantidimaging-dev environment_](https://github.com/mantidproject/mantidimaging/issues/1331#issuecomment-1049798985) - change the versions of the packages we're using so that our `mantidimaging-dev` conda environment is Windows compatible.

2) Set the style of the application to 'Fusion', as detailed under [_GUI features_](https://github.com/mantidproject/mantidimaging/issues/1331#issuecomment-1049798985) - this is a very small change that shouldn't make any visible difference on Linux, but would mean the application had a consistent look and feel when run on Windows.

3) Address the issues with checking the CUDA installation detailed in [_Running the application_](https://github.com/mantidproject/mantidimaging/issues/1331#issuecomment-1049798985) - update the code so it works on both Windows and Linux.

These first steps would mean developers could get Mantid Imaging up and running on both Windows and Linux without having to tweak settings for Windows. The application won't yet be fully functional on Windows.

4) Test the reconstruction speed on some other Windows machines to get a better sense of how slow it really is. This will help figure out how viable it is to use the system on Windows, or if we potentially need to chat with the CIL developers about algorithm speeds on Windows.

5) Work through the steps given in the [suggested approach for resolving the multiprocessing issues](https://github.com/mantidproject/mantidimaging/issues/1331#issuecomment-1057106957). Each of the three steps should be possible to do as a separate PR.
  **Step 1** will update the way we're using shared memory to lay the foundations for Windows compatibility, but the application will still only work with forked processes at this point.
  **Step 2** will change the application so that it should work with both spawn and fork. At this point the multiprocessing should work on both Windows and Linux, however performance on Windows will be very slow.
  **Step 3** will address the performance issues on Windows, but will require a design decision about whether we want to start using spawn on Linux and also whether we want to create processes up front on both Linux and Windows.

6) Fix the remaining issues listed under [_GUI features_](https://github.com/mantidproject/mantidimaging/issues/1331#issuecomment-1049798985) and any others that arise from testing on Windows. After this step the system should be fully functional on Windows.

rbauststfc:
Outcomes of investigation into Windows compatibility, Feb 2022:

**Summary**

[...]

PR #1403, opened 2022-04-08
Use shared memory for compatibility with spawn

rbauststfc:
### Issue

Refs #1331

### Description

Changes the way we work with shared memory when performing parallel processing tasks to be compatible with spawn (and consequently still fork). This completes step two of the multiprocessing changes described under point 5 of the issue description.

I have also updated one of the unit tests for the dead stripe removal functionality that was failing on Windows. This was because the algotom algorithm that we use for this filter uses `numpy.polyfit`, which gives slightly different results due to small differences in handling of floating point numbers between Linux and Windows. Increasing the size of the test dataset seemed get the test passing in this case.

At this stage the application should now be fully functional on both Windows and Linux, although performance will be quite slow on Windows. The operations that were updated in this PR should be tested on both platforms to confirm this, as well as testing the "Correlate 0 and 180" functionality in the recon window.

All unit tests should pass on both platforms. Further changes will be required to get the GUI system tests and Applitools tests running on Windows.

### Documentation

Issue to be added to release notes once all steps are completed.

### DIFF
--- a/mantidimaging/core/operations/arithmetic/arithmetic.py
+++ b/mantidimaging/core/operations/arithmetic/arithmetic.py
@@ -120,5 +120,4 @@ def _execute(images: ImageStack, div_val: float, mult_val: float, add_val: float
              progress):
     arg_list = [div_val, mult_val, add_val, sub_val]
     do_arithmetic = ps.create_partial(_arithmetic_func, fwd_function=ps.arithmetic, arg_list=arg_list)
-    ps.shared_list = [images.shared_array]
-    ps.execute(do_arithmetic, images.data.shape[0], progress, cores=cores)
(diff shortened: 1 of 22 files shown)


rbauststfc:
I removed a rebin unit test in commit [58183d1](https://github.com/mantidproject/mantidimaging/pull/1403/commits/58183d125a6f58fe2673ee3ce81f752cfd939e92) that checked that the order of parameters passed into the `execute` method (from the parallel utility module) were correct. I tried for a while but couldn't find a way to get this working using patching. I couldn't see similar tests for the other operations so I concluded it was OK to remove. I'm happy to re-instate it if there's a way to get it working that I've missed, though.

PR #1404, opened 2022-04-12
Cleanup FiltersWindowPresenter._post_filter

### LINKED ISSUE #1322: FiltersWindowPresenter._post_filter: attempt_repair

samtygier-stfc:
### Issue

Closes #1322

### Description

The `attempt_repair` did not actually do any repairing. It just called a method to redraw the the visualiser. (The only defence we have against an operations doing something bad tot he data and then failing is the safe mode). We need to redraw the visualisers no matter what happens, but we already emit `filter_applied` signal, so makes sense for the visualisers to just listen for that. The operations are all in-place, so the visualisers don't need to be told that there is a new image stack.

This means that several methods are no longer needed. `set_image_data_by_uuid()`, `update_stack_with_images()`, `set_images_in_stack()` 

### Testing 

Run some operations in the ops window.

### Acceptance Criteria 

Visualisers in the main window should be updated when an operation finishes.

### Documentation

Not needed.

### DIFF
--- a/mantidimaging/gui/windows/main/model.py
+++ b/mantidimaging/gui/windows/main/model.py
@@ -91,18 +91,6 @@ class MainWindowModel(object):
         else:
             raise RuntimeError(f"Failed to get StrictDataset with ID {dataset_id}")
 
-    def set_image_data_by_uuid(self, images_id: uuid.UUID, new_stack: ImageStack) -> None:
-        """
-        Updates the data of an existing dataset/images object.
-        :param images_id: The id of the image to update.
-        :param new_stack: The ImageStack containing the new data.
-        """
-        for dataset in self.datasets.values():
-            if images_id in dataset:
-                dataset.replace(images_id, new_stack)
-                return
-        self.raise_error_when_images_not_found(images_id)
-
     def get_existing_180_id(self, dataset_id: uuid.UUID) -> Optional[uuid.UUID]:
         """
         Gets the ID of the 180 projection object in a Dataset.
--- a/mantidimaging/core/data/dataset.py
+++ b/mantidimaging/core/data/dataset.py
@@ -52,13 +52,6 @@ class BaseDataset:
     def delete_stack(self, images_id: uuid.UUID):
         raise NotImplementedError()
 
-    def replace(self, images_id: uuid.UUID, image_stack: ImageStack):
-        for image in self.all:
-            if image.id == images_id:
-                image.shared_array = image_stack.shared_array
-                return
-        raise KeyError(f"Unable to replace: ImageStack with ID {images_id} not present in dataset.")
-
     def __contains__(self, images_id: uuid.UUID) -> bool:
         return any([image.id == images_id for image in self.all])
 
--- a/mantidimaging/gui/test/gui_system_operations_test.py
+++ b/mantidimaging/gui/test/gui_system_operations_test.py
@@ -5,7 +5,6 @@ from itertools import product
 from unittest import mock
 from uuid import UUID
 
-import pytest
 from parameterized import parameterized
 from PyQt5.QtTest import QTest
 from PyQt5.QtCore import Qt, QTimer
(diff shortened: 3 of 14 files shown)


rbauststfc:
I seem to be getting the below error if I load a dataset and then run the Remove Outliers operation. If I run it with Safe Apply and then choose not to keep the result it works fine on a subsequent run, but if I keep the result on the first run (or run without Safe Apply as the first run) I get the below:

```
2022-04-20 07:56:45,535 [mantidimaging.gui.windows.main.view:L431] ERROR: Traceback (most recent call last):
  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/presenter.py", line 56, in notify
    self.refresh_image()
  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/presenter.py", line 74, in refresh_image
    else self.images.data
AttributeError: 'NoneType' object has no attribute 'data'

During handling of the above exception, another exception occurred:

  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/view.py", line 78, in <lambda>
    self._main_window.stack_changed.connect(lambda: self.presenter.notify(SVNotification.REFRESH_IMAGE))
  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/presenter.py", line 66, in notify
    self.show_error(e, traceback.format_exc())
  File "/mantidimaging/mantidimaging/gui/mvp_base/presenter.py", line 21, in show_error
    if hasattr(self.view, 'show_error_dialog'):
RuntimeError: wrapped C/C++ object of type StackVisualiserView has been deleted
```

rbauststfc:
I seem to be getting the below error if I load a dataset and then run the Remove Outliers operation. If I run it with Safe Apply and then choose not to keep the result it works fine on a subsequent run, but if I keep the result on the first run (or run without Safe Apply as the first run) I get the below:

```
2022-04-20 07:56:45,535 [mantidimaging.gui.windows.main.view:L431] ERROR: Traceback (most recent call last):
  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/presenter.py", line 56, in notify
    self.refresh_image()
  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/presenter.py", line 74, in refresh_image
    else self.images.data
AttributeError: 'NoneType' object has no attribute 'data'

During handling of the above exception, another exception occurred:

  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/view.py", line 78, in <lambda>
    self._main_window.stack_changed.connect(lambda: self.presenter.notify(SVNotification.REFRESH_IMAGE))
  File "/mantidimaging/mantidimaging/gui/windows/stack_visualiser/presenter.py", line 66, in notify
    self.show_error(e, traceback.format_exc())
  File "/mantidimaging/mantidimaging/gui/mvp_base/presenter.py", line 21, in show_error
    if hasattr(self.view, 'show_error_dialog'):
RuntimeError: wrapped C/C++ object of type StackVisualiserView has been deleted
```

samtygier-stfc:
Error is when there is a `StackVisualiserView` instance from a closed stack, that is still listening to the signals.

The `StackVisualiserView`s are kept alive by:
A reference loop between StackVisualiserPresenter and StackVisualiserView
MIImageView.roi_changed_callback holding a method (MIImageView is kept alive in the graveyard)
The MainWindowView.stack_changed signal itself (lives in main window and holds reference to StackVisualiserPresenter.notify).

We can clean up the references. Also need to disconnect the signal.

mantidimaging/gui/windows/main/model.py
```diff
@@ -91,18 +91,6 @@ def do_nexus_saving(self, dataset_id: uuid.UUID, path: str, sample_name: str):
...
 
-    def set_image_data_by_uuid(self, images_id: uuid.UUID, new_stack: ImageStack) -> None:
-        """
-        Updates the data of an existing dataset/images object.
-        :param images_id: The id of the image to update.
-        :param new_stack: The ImageStack containing the new data.
-        """
-        for dataset in self.datasets.values():
-            if images_id in dataset:
-                dataset.replace(images_id, new_stack)
```
rbauststfc, line 102: I don't think this `replace` method from `BaseDataset` is used anywhere now. Would it be worth removing, or do we think it's worth keeping in case it's needed in the future?

rbauststfc (commented):
Looks like a good clean-up! I just had one small query, but looks like it's working well.

rbauststfc (approved):

PR #1412, opened 2022-04-21
NeXus Save Tests

DolicaAkelloEgwel:
Was supposed to be part of the last PR. Just bringing up test coverage.


samtygier-stfc (changes requested):
Can you also cover the case of `h5py.File()` failing. For example if the user chooses a path that can't be opened.

samtygier-stfc (approved):

PR #1416, opened 2022-04-26
Create pool of processes on start-up

rbauststfc:
### Issue

Refs #1331

### Description

Completes the work required for step 3 of bullet point 5 in the referenced issue description. The pool of processes are now started when the application begins (asynchronously to minimise the impact on start-up time) to improve performance on Windows. This change means that there is little performance difference between spawn and fork on Linux, and there are potentially other benefits to switching to spawn (one example being to avoid incorrect resource tracker warnings from the sub processes when the application ends), so this PR switches us to use spawn everywhere.

I've removed the core and chunksize parameters from the filter functions, as these were not being used by the code to run MI as a GUI application and we don't have any API users as yet. These changes mean that making calls to the API would always run the filters synchronously, so if we do have people interested in using the API in the future we may want to look at how to maximise performance from direct calls.

All tests should pass in a reasonable duration. Note tests are likely to become a bit slower because of the extra time needed to spawn the process pool when testing parallel processing.

Test the operations affected in the PR, and the reconstruction auto correlate button.

### Documentation

There is one final step required on the issue before it can be added to the release notes.

### DIFF
--- a/docs/ext/operations_user_doc.py
+++ b/docs/ext/operations_user_doc.py
@@ -29,7 +29,7 @@ def split_lines(s: str) -> List[str]:
     return s.split("\n")
 
 
-PARAM_SKIP_LIST = ["images", "cores", "chunksize", "progress"]
+PARAM_SKIP_LIST = ["images", "progress"]
 
 
 def get_params(s: str) -> List[str]:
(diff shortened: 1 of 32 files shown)


samtygier-stfc (approved):
Looks good.

PR #1423, opened 2022-04-27
Windows GUI fixes

rbauststfc:
### Issue

Refs #1331

### Description

Includes the following GUI and GUI test fixes:

1) Ensure background image shows in welcome dialog
2) Hide the `?` (context help) button that appears at the top right of the dialog title bar on Windows
3) Update the licence check pre-commit command to be compatible with Windows
4) Start up the process pool to speed up the operations and recon GUI system tests

A number of things are still outstanding and will be covered in later PRs:

1) Look at window resizing, as this doesn't seem to work on Windows
2) Look at documentation updates to support Windows users (to be done at release)
3) Figure out how to run Applitools tests on Windows locally
4) Set up GitHub Actions to run system and eyes tests. Not yet sure if/how this will work without xvfb. The GUI system tests work on Windows (locally) if the test data is located at `C:\Users\[username]\mantidimaging...`, but we could look at making this more flexible

Check the background image displays in the welcome dialog on both Windows and Linux.
Check that there is no `?` button (next to the close button) in any dialog title bars on Windows. Check on Linux that the dialogs are still displaying as expected.

### Documentation

Issue number could be added to release notes after window resizing has been fixed.


mantidimaging/gui/windows/welcome_screen/view.py
```diff
@@ -11,9 +10,9 @@
 class WelcomeScreenView(BaseDialogView):
     def __init__(self, parent, presenter):
         super().__init__(parent, "gui/ui/welcome_screen_dialog.ui")
-        base_path = finder.ROOT_PATH
 
-        bg_image = os.path.join(base_path, "gui/ui/images/welcome_screen_background.png")
+        # The background image URL must use forward slashes on both Windows and Linux
+        bg_image = finder.ROOT_PATH.replace('\\', '/') + '/gui/ui/images/welcome_screen_background.png'
```
samtygier-stfc, line 15: I think it would be better to switch `finder.ROOT_PATH` to be a Path (from pathlib) rather than a string. Then the separators should be correct automatically. Would looks something like:

`bg_image = finder.ROOT_PATH / "gui" / "ui" ...`

(will probably need changes in a few other places)

samtygier-stfc (changes requested):

mantidimaging/gui/windows/welcome_screen/view.py
rbauststfc: Sorry, I think I'm getting a bit confused about the suggestion. `finder.ROOT_PATH` is outputting a Windows style path on Windows (i.e. using back slashes). The issue here was because the code uses the path to set the style `border-image:url({bg_image});` and it seemed that the url here needed to have forward slashes, regardless of platform.

In other circumstances where we're looking up a path (rather than using it to set a style sheet value) Windows can actually cope with backward, forward or a mix of slashes.

Let me know if I'm missing the point of the suggestion, though.

PR #1431, opened 2022-04-28
1420: Copy NeXus data without concatenate

### LINKED ISSUE #1420: NeXus Saving: Copy data without concatenate
### Desired Behaviour

Create an empty hdf5 Dataset and copy the different arrays rather than using `np.concatenate`

DolicaAkelloEgwel:
### Issue

Closes #1420

### Description

Creates an empty hdf5 Dataset and copies the `StrictDataset` data to it in chunks rather than using `np.concatenate`

### Testing 

The existing tests already cover the change. Also multiplied test arrays by 12 as 

### Acceptance Criteria 

Check that the tests pass.

### Documentation

Updated release notes.

### DIFF
--- a/mantidimaging/core/io/saver.py
+++ b/mantidimaging/core/io/saver.py
@@ -219,7 +219,7 @@ def _nexus_save(nexus_file: h5py.File, dataset: StrictDataset, sample_name: str)
     detector.create_dataset("data", shape=combined_data_shape, dtype="uint16")
     index = 0
     for arr in dataset.nexus_arrays:
-        detector["data"][index:index + arr.shape[0]] = arr.astype("uint16")
+        detector["data"][index:index + arr.shape[0]] = arr
         index += arr.shape[0]
     detector.create_dataset("image_key", data=dataset.image_keys)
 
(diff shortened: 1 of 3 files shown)


mantidimaging/core/io/saver.py
```diff
@@ -215,8 +215,12 @@ def _nexus_save(nexus_file: h5py.File, dataset: StrictDataset, sample_name: str)
     _set_nx_class(detector, "NXdetector")
 
     # instrument data
-    combined_data = np.concatenate(dataset.nexus_arrays).astype("uint16")
-    detector.create_dataset("data", data=combined_data)
+    combined_data_shape = (sum([len(arr) for arr in dataset.nexus_arrays]), ) + dataset.nexus_arrays[0].shape[1:]
+    detector.create_dataset("data", shape=combined_data_shape, dtype="uint16")
+    index = 0
+    for arr in dataset.nexus_arrays:
+        detector["data"][index:index + arr.shape[0]] = arr.astype("uint16")
```
samtygier-stfc, line 222: I think the astype is still generating a copy. If I save a large dataset (Olivin 12GB) I still a lot of memory use and mantidimaging gets OOM killed.

I tried with
`detector["data"][index:index + arr.shape[0]] = arr`
Which reduces memory use, but I have not checked it effects what is written to the file. I'd expect the conversion to still happen properly.

samtygier-stfc (approved):

Issue #1219, opened 2021-12-07
When loading files like NAME_001.tif NAME_002.tif don't include NAME_180deg.tif


samtygier-stfc:
### Summary

The IAEA_Cylinder_WhiteBeam dataset has a sample folder with files named

`IMAT00006388_PSI_cylinder_Sample_000.tif` -> `IMAT00006388_PSI_cylinder_Sample_943.tif`

plus
```
IMAT00006388_PSI_cylinder_Sample_180deg.tif
IMAT00006388_PSI_cylinder_Sample_360deg.tif
```
Currently the 180deg and 360deg get included in the main stack of projections

We should probably audit all the variations of how files are being named.

### Steps To Reproduce

Data set is at `\\isis\shares\IMAT\ExampleData\IAEA_Cylinder_WhiteBeam\Sample\`

or Spheres_1219.tar.xz in the onedrive example data folder.

Load the Tomo folder with Load Dataset

In the main windows scroll though the image stack, the issue causes the `Spheres_Tomo_180deg.tiff` to be included as the 180th projection in the main sample stack. So as you scroll passed z=180 the rotation will jump.

### Expected Behaviour

180 deg should be loaded as a 180. 360 should be ignored

### Current Behaviour

The 'deg' files are included in the main stack, and create jumps

### Context

Current main, not yet tested on older versions

PR #1440, opened 2022-05-04
Add FilenamePattern and FilenameGroup

samtygier-stfc:
### Issue
Work towards #1219

### Description
Add FilenamePattern and FilenameGroup.

These will be home to all the logic for finding and using groups of file names for stacks. They should also be a single point for change if we need to support different patterns of filenames.

### Testing 

Class includes some tests

### Acceptance Criteria 

Does not impact any other code yet

### Documentation

not yet

### DIFF
--- a/mantidimaging/core/io/filenames.py
+++ b/mantidimaging/core/io/filenames.py
@@ -4,6 +4,9 @@
 from pathlib import Path
 import re
 from typing import List, Iterator, Optional
+from logging import getLogger
+
+LOG = getLogger(__name__)
 
 
 class FilenamePattern:
@@ -29,7 +32,7 @@ class FilenamePattern:
             self.re_pattern = re.compile("^" + re.escape(prefix) + "([1-9]?[0-9]{" + str(digit_count) + "})" +
                                          re.escape(suffix) + "$")
 
-        self.re_pattern_metadata = re.compile("^" + re.escape(prefix.rstrip("_ ")) + ".json")
+        self.re_pattern_metadata = re.compile("^" + re.escape(prefix.rstrip("_ ")) + ".json$")
         self.template = prefix + "{:0" + str(digit_count) + "d}" + suffix
 
     @classmethod
@@ -99,6 +102,8 @@ class FilenameGroup:
                 self.all_indexes.append(self.pattern.get_index(filename.name))
 
             if self.pattern.match_metadata(filename.name):
+                if self.metadata_path is not None:
+                    LOG.warning(f"Multiple metadata files found: {filename}")
                 self.metadata_path = filename
         self.all_indexes.sort()
 


mantidimaging/core/io/filenames.py
```diff
@@ -0,0 +1,113 @@
...
+
+    @classmethod
+    def from_file(cls, path: Path) -> "FilenameGroup":
+        if path.is_dir():
+            raise ValueError(f"path is a directory: {path}")
+        directory = path.parent
+        name = path.name
+        pattern = FilenamePattern.from_name(name)
+        index = pattern.get_value(name)
+        new_filename_pattern = cls(directory, pattern, [index])
```
rbauststfc, line 88: This method seems to be returning a `FilenameGroup` - would it make sense to rename this variable to be `new_filename_group`?

mantidimaging/core/io/filenames.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def generate(self, index: int) -> str:
+        if self.digit_count == 0:
+            return self.prefix + self.suffix
+        return self.template.format(index)
+
+    def match(self, filename: str) -> bool:
+        return self.re_pattern.match(filename) is not None
+
+    def get_value(self, filename: str) -> int:
```
rbauststfc, line 60: Would this make sense to rename to `get_index`? As I was reading through it felt like it might be more consistent with the naming used elsewhere.

mantidimaging/core/io/filenames.py
```diff
@@ -0,0 +1,113 @@
...
+        for index in self.all_indexes:
+            yield self.directory / self.pattern.generate(index)
+
+    def find_all_files(self) -> None:
+        for filename in self.directory.iterdir():
+            if self.pattern.match(filename.name):
+                self.all_indexes.append(self.pattern.get_value(filename.name))
+
+            if self.pattern.match_metadata(filename.name):
+                self.metadata_path = filename
```
rbauststfc, line 102: Just to check, is the idea that if there were multiple metadata files in a directory then only the last one found would be used for `self.metadata_path`? Or are we assuming there will only ever be one metadata file?

mantidimaging/core/io/test/filenames_test.py
```diff
@@ -0,0 +1,164 @@
...
+    def test_pattern_match_different_digits(self):
+        # Allow cases where index has grown above padding
+        p1 = FilenamePattern.from_name("img_000.tif")
+        self.assertTrue(p1.match("img_000.tif"))
+        self.assertTrue(p1.match("img_001.tif"))
+        self.assertTrue(p1.match("img_999.tif"))
+        self.assertTrue(p1.match("img_1000.tif"))
+        self.assertFalse(p1.match("img_0000.tif"))
+
+    def test_pattern_get_value(self):
```
rbauststfc, line 59: Thanks for renaming those bits. I think this unit test needs updating to `test_pattern_get_index` and then everything looks good to go in.

rbauststfc (approved):

PR #1448, opened 2022-05-09
WelcomeScreenPresenterTest test in isolation

### LINKED ISSUE #1446: WelcomeScreenPresenterTest fails if run in isolation
### Summary

While investigating #1427 I tried running each test file in isolation.

I see a reproducible failure in WelcomeScreenPresenterTest::test_show_today

### Steps To Reproduce

`pytest -vs mantidimaging/gui/windows/welcome_screen/tests/presenter_test.py`

but it will pass if another file is added, e.g.

`pytest -vs mantidimaging/gui/windows/welcome_screen/tests/presenter_test.py mantidimaging/gui/windows/main/test/view_test.py`

### Expected Behaviour

pass

### Current Behaviour

Test fails

### Context

Current main

### Failure Logs

```
    def test_show_today(self):
        settings = QSettings()
        settings.setValue("welcome_screen/show_at_start", True)
        settings.setValue("welcome_screen/last_run_version", "")
        self.assertTrue(WelcomeScreenPresenter.show_today())
    
        settings.setValue("welcome_screen/show_at_start", False)
        settings.setValue("welcome_screen/last_run_version", "")
>       self.assertTrue(WelcomeScreenPresenter.show_today())
E       AssertionError: None is not true

```

samtygier-stfc:
### Issue

Closes #1446 

### Description

The version check has `versions._use_test_values()` which sets some standard version to make tests work. `WelcomeScreenPresenterTest` was missing a call to set this. Usually this was not a problem, because several other tests call it at import time. If the test was run in isolation then the real version checking functions would run causing the test to fail. Running tests with xdist may also cause the import not to happen in the process that runs the WelcomeScreenPresenterTest tests.

A better fix than adding `versions._use_test_values()`, is to stop depending on global state. Its a bit fiddly to set up the mock every time this is needed (every time the main window is used), so add a helper.

Try the pytest command on the #1446

Add a RuntimeError to top of `CheckVersion._retrieve_conda_installed_version()` and `CheckVersion._retrieve_conda_available_version()`. Run `make check`. Confirm that the only tests that fail are the tests in `TestCheckVersion` that have proper mocks for `subprocess.check_output` and `requests.get`. This confirms that none of the tests depend on external version numbers.

### Documentation

Not needed

### DIFF
--- a/mantidimaging/test_helpers/start_qapplication.py
+++ b/mantidimaging/test_helpers/start_qapplication.py
@@ -110,7 +110,7 @@ def mock_versions(cls):
     def tearDownClass():
         cls.mock_version_patch.stop()
         cls.mock_version_label_patch.stop()
-        cls.mock_version_available_patch.start()
+        cls.mock_version_available_patch.stop()
 
     return augment_test_setup_methods(cls, setup_class=setUpClass, teardown_class=tearDownClass)
 
--- a/mantidimaging/test_helpers/start_qapplication.py
+++ b/mantidimaging/test_helpers/start_qapplication.py
@@ -15,6 +15,7 @@ import gc
 import os
 import pytest
 import sys
+from unittest import mock
 
 import pyqtgraph
 from mantidimaging.core.parallel.manager import create_and_start_pool, end_pool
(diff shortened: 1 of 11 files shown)


mantidimaging/test_helpers/start_qapplication.py
```diff
@@ -89,6 +90,31 @@ def tearDownClass():
...
+
+        cls.mock_version_available_patch = mock.patch(
+            'mantidimaging.gui.windows.welcome_screen.presenter.versions.get_conda_available_version')
+        cls.mock_version_available = cls.mock_version_available_patch.start()
+        cls.mock_version_available.return_value = "1.0.0_1"
+
+    def tearDownClass():
+        cls.mock_version_patch.stop()
+        cls.mock_version_label_patch.stop()
+        cls.mock_version_available_patch.start()
```
rbauststfc, line 113: Is this intended to be `cls.mock_version_available_patch.stop()`?

rbauststfc (approved):

PR #1477, opened 2022-05-20
Fix scaling when saving to int

### LINKED ISSUE #1475: When saving a stack to int tiffs scaling is not consistent between slices
### Summary

The output of a reconstruction typically has values between -1 and 1 (usually only a small dip below zero). When saving as tiff in int16 these values must be scaled in the the 0-65535 range. The same scaling should be used for each image in the stack.

### Steps To Reproduce

Do a reconstruction.
Save the reconstructed stack as int16 tiffs

Inspect the values in the file, by loading back into MI with Load Images.

### Expected Behaviour

For a typical recon the air regions should have consistent values, and so the pixel values should be similar in each

### Current Behaviour

For an all zero slice, the values in the tiff are written out as 65535.

For other values there also differences

### Context

2.3.1 on IDAaaS

### Screenshot(s)

Taking an orthogonal slice through the saved tifs shows the issue

image

This issue does not occur for float tifs because no scaling is done

image

samtygier-stfc:
### Issue

Closes #1475

### Description

The scaling operation was using nanmax() of each slice in the scaling calculation. Change to to only depend on the function arguments, as this is what is really wanted.

Unify the 2 methods in the filter to reduce duplication.

Use numpy interp, because it does the scaling and clipping that we want in a single step

Note this does change the behaviour of the rescale operation, but in a way that makes it more sensible.

See bug report

### Documentation

Release_notes

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -25,6 +25,8 @@ Fixes
 - #1395 : Delete file when NeXus saving fails
 - #1397 : Only use real NeXus projection angles
 - #1473 : Tomopy COR finder use -log
+- #1475 : Fix scaling when saving to int
+
 
 Developer Changes
 -----------------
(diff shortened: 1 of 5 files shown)


StephenNneji (approved):
I tested on Linux and it works as described, int16 output is now scaled properly

PR #1485, opened 2022-06-06
Add line profile to reconstruction window

### LINKED ISSUE #1483: Recon line profile
Users want to be able to view line profiles of slices from the reconstruction preview to help with adjusting BHC coefficients or PDHG alpha when running a reconstruction. This will require a movable line to be added to the recon preview for selecting sections of the image to plot. Ideally this would be a line drawn between two movable points for maximum flexibility over the slices that can be selected.

The users have indicated that their preferred location for the plot would be under the reconstruction preview image in the recon window.

rbauststfc:
### Issue

Closes #1483

### Description

Line profile added to the reconstruction window to allow users to select a line segment from the recon preview to be plotted in the graph below.

Still to do towards this (under a separate issue) is re-sizing the plot so that it is smaller than the three image views in the grid layout. This will help keep the recon preview from getting too small. I've also noticed that if you switch between stacks from the drop-down on Linux, the sizing of the recon preview doesn't seem to be re-adjusting very well. This behaviour doesn't seem to occur on Windows. We should look at this as part of the stage 2 re-sizing work for this issue.

All tests pass (some eyes test baselines will need to be updated).
The ROI line can be used to select line sections from the image, which are correctly plotted on the graph below. These update correctly when the recon preview is updated and reset automatically when a new stack with different dimensions is selected from the drop-down.
The line profile plot is cleared when all data is deleted from the main window tree view.
The "Reset Profile Line" option appears on the image view right click menu for the recon preview. When clicked, this should reset the position of the ROI line and update the graph.
The ROI line handles can be used to select line segments at any angle.

### Documentation

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -17,6 +17,7 @@ New Features
 - #1430 : Provide an Anaconda package that is compatible with Windows
 - #1398 : Save recons to NeXus file
 - #1444 : CIL PDHG non-negativity constraint
+- #1483 : Add line profile to reconstruction window
 
 Fixes
 -----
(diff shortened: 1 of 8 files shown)


samtygier-stfc (approved):
Looks good.

PR #1488, opened 2022-06-08
Clear ops and recon window previews when there is no data to display

### LINKED ISSUE #1484: Segmentation fault if delete all loaded data with recon window open
### Summary

The application crashes with a segmentation fault if all data is deleted from the main window tree view while the reconstruction window is open. The crash doesn't occur straight away, but happens the next time the recon preview is triggered.

### Steps To Reproduce

1) Load a dataset
1) Open the reconstruction window
1) Minimise the recon window and delete the dataset from the main window
1) Maximise the recon window

### Expected Behaviour

The application should not crash. Need to decide what should happen instead - perhaps any open windows that require data should be closed when the tree view is empty?

### Current Behaviour

The application crashes with a segmentation fault.

### Context

On current main.

rbauststfc:
### Issue

Closes #1484

### Description

The source of the seg fault described on the issue was the `show_value` method in `MIMiniImageView` class. This is used to display the pixel value of the point that the mouse is hovering over on the image. In the reconstruction window, the projection and sinogram previews are backed by shared memory, so when the data was deleted this functionality would seg fault when the user hovered their mouse over either of these previews.

It feels a little bit confusing to be holding on to a view of data that the application no longer has available, so the solution in this PR is to clear the previews in the recon window when there is no data available to display. For consistency, I've also resolved this in the operations window, even though the previews there use copies of the data and so do not cause seg faults. I've also made some changes to make sure that bad data overlays are hidden when there is no data, and that status messages in the recon window are cleared.

The recon window seg fault should no longer occur - the updated steps to test this are in the comment on the issue (note the original steps given in the issue description turned out to be incorrect on further investigation).

Deleting all data from the tree view should clear all previews, bad data overlays and error messages from the operations and reconstruction windows, if they are open. When new data is loaded, these windows should update and function correctly.

### Documentation

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -26,6 +26,7 @@ Fixes
 - #1397 : Only use real NeXus projection angles
 - #1473 : Tomopy COR finder use -log
 - #1475 : Fix scaling when saving to int
+- #1484 : Clear image previews in open windows when there are no stacks to select
 
 
 Developer Changes
--- a/mantidimaging/gui/widgets/bad_data_overlay/bad_data_overlay.py
+++ b/mantidimaging/gui/widgets/bad_data_overlay/bad_data_overlay.py
@@ -43,9 +43,13 @@ class BadDataCheck:
         self.overlay.setZValue(11)
         self.overlay.setLevels([0, 1])
 
-    def clear(self):
+    def remove(self):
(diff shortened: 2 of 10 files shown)


samtygier-stfc (approved):
Looks good.

Sensible change. It was a bit confusing before that the previews stayed after the data had been deleted.

PR #1505, opened 2022-06-16
1504 disable windows publish

### LINKED ISSUE #1504: Stop windows publish step from running
### Summary

The windows github actions has a publish step, so that we know that producing packages on windows works.

It should be disabled because we only need to build a single noarch package and can do this faster on linux. If they both run then it can mess up the conda labels.

samtygier-stfc:
### Issue

Closes #1504

### Description

I'm not sure why `if false && ...` was not sufficient to stop it running. But hopefully `if: false` should.

`if: false` worked for other rules on ef5566f2b57395b36112dcf2691c339967299af2

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -122,9 +122,10 @@ jobs:
           python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov mantidimaging/eyes_tests --durations=10
         timeout-minutes: 15
 
-      # Label as 'windows-build-test' for testing purposes. To use, remove 'false &&' from the 'if' condition and change label to 'unstable' after testing
+      # Label as 'windows-build-test' for testing purposes.
+      # To re-enable, use the if rule from the conda workflow and change label to 'unstable' after testing
       - name: publish package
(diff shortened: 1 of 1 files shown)


samtygier-stfc:
> Possibly there were missing brackets

I guess so, but I feel that's an odd way around for it to work.

.github/workflows/windows.yml
```diff
@@ -124,7 +124,8 @@ jobs:
 
       # Label as 'windows-build-test' for testing purposes. To use, remove 'false &&' from the 'if' condition and change label to 'unstable' after testing
```
rbauststfc, line 125: Is it worth updating this comment now that the approach has changed?

Also, I was just wondering why we're using an `if` statement to stop this running rather than commenting it out in the workflow file? It seems a shame to have to evaluate the logic each time, is there an advantage to leaving it in that I'm not considering?

samtygier-stfc: Good spot, I have a bad habit of not spotting when comments need updating.

I'm not a fan of commented out code. Gives a higher risk of it not working when uncommented later. Actually now thinking that the commented `if ...` should be removed, and the advice should be to copy it back from the conda work flow if needed. That way it wont get out of date with any changes of when to build.

rbauststfc (approved):

PR #1517, opened 2022-06-22
SpectrumWidget

### LINKED ISSUE #1512: Add image and spectum views

samtygier-stfc:
### Issue

Closes #1512

### Description

Add SpectrumWidget with an image and plot region.

Simple display of the mean of the image values and the spectrum.

Tweaks to layout

There is no testing of show_new_sample(), because I expect that functionality will move into a model class soon.

See #1510 to enable the spectrum menu entry

Load a dataset. Open spectrum window. Select sample.

The image should show the image averaged through the 3rd axis. The spectrum view should show a line.

e.g. for the flower dataset it would look something like:
image
(Note this has angle as the third axis, so is not showing a spectrum)

### Documentation

Wait until whole feature is done

### DIFF
--- a/mantidimaging/gui/ui/spectrum_viewer.ui
+++ b/mantidimaging/gui/ui/spectrum_viewer.ui
@@ -14,168 +14,181 @@
    <string>Spectrum Viewer</string>
   </property>
   <widget class="QWidget" name="centralwidget">
-   <widget class="QSplitter" name="splitter">
-    <property name="geometry">
-     <rect>
-      <x>10</x>
-      <y>0</y>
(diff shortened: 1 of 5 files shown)


rbauststfc (approved):
The functionality is looking good! One thing I noticed when testing was that, although a sample is currently selected by default in the drop-down when the spectrum viewer dialog is first opened, no image or plot is drawn straight away. You have to change the sample selection to trigger drawing the image and plot. As part of the next stages it would be good to decide how the window should display when it's first opened, but I know this is still a work in progress, so very happy to approve as it is.

PR #1526, opened 2022-06-30
Add ToF range control to spectrum window

### LINKED ISSUE #1521: Spectrum viewer range control
Add a range control to the spectrum.

Update the main view with the selected subset

samtygier-stfc:
### Issue

Closes #1521 

### Description

Add a ToF range slider to the spectrum widget. Moving the slider updates the region that the main view is averaged over.

Also enables the Spectrum Viewer menu item.

### Testing 

- Open the 4_powders_tof_tiffs dataset (There is an `open` directory that can be loaded as the Flat Before)
- Workflow -> Spectrum Viewer
- Select the tbin_0 stack for the sample (you might need to select a different stack, then switch back)
- Adjust the scale so that you can see the powders
- Use the yellow bars on the spectrum to adjust the displayed range

### Acceptance Criteria 

Everything should work, no error messages or crashes

### Documentation

Not yet, will add to the release notes when feature complete

### DIFF
--- a/mantidimaging/gui/ui/main_window.ui
+++ b/mantidimaging/gui/ui/main_window.ui
@@ -225,9 +225,6 @@
    </property>
   </action>
   <action name="actionSpectrumViewer">
-   <property name="enabled">
-    <bool>false</bool>
-   </property>
    <property name="text">
     <string>Spectrum Viewer (Beta)</string>
    </property>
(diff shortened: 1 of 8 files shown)


rbauststfc (commented):
All working well. I just had a couple of suggestions for consideration, but looks good.

I noticed that when the window first opens it doesn't pick up the selected image until after we trigger `handle_sample_change` by manually toggling the drop-downs. Is it worth adding some relevant setup to the presenter `__init__` method (or elsewhere if more appropriate) so that the image is set as soon as the window opens?

rbauststfc (approved):

PR #1528, opened 2022-06-30
OperationsUserDoc record dependant files

### LINKED ISSUE #1527: Sphinx does not rebuild operations list on docstring changes
### Summary

`user_guide/operations/index.html` is built from `docs/user_guide/operations/index.rst` by `docs/ext/operations_user_doc.py` from docstrings in the operation classes.

But sphinx can't see the dependencies so will only rebuild the `html` page if the `rst` file is updated. So changes to docstrings are not seen unless the `docs/build` folder is deleted.

### Steps To Reproduce

Build the docs `python ./setup.py internal_doc`
Edit the class docstring in an operation, e.g. `mantidimaging/core/operations/arithmetic/arithmetic.py`
Rebuild the docs
Note that the change is not reflected in the output html.

### Expected Behaviour

The doc build should rebuild all html files that need updating

### Current Behaviour

Changes to docstrings not detected.

### Context
Current main

samtygier-stfc:
### Issue
Closes #1527 

### Description
Record files used to build the docs. This means that changes to them will trigger a rebuild of the appropriate page.

### Testing 

Follow steps on #1527

### Acceptance Criteria 

Edits to the operation class doc strings should cause the html page to be updated with sphinx is run again.

### Documentation
Not needed

### DIFF
--- a/docs/ext/operations_user_doc.py
+++ b/docs/ext/operations_user_doc.py
@@ -56,10 +56,14 @@ class OperationsUserDoc(Directive):
         except ImportError:
             raise ValueError("operations_user_doc could not import load_filter_packages")
 
+        env = self.state.document.settings.env
+        env.note_dependency(__file__)
+
         rst_lines = []
 
         operations = load_filter_packages()
         for op in operations:
+            env.note_dependency(inspect.getfile(op))
             # Title
             rst_lines += make_heading(op.filter_name, "-")
 


rbauststfc:
I'm afraid I can't seem to get this to work when I follow the steps on the issue. I'm using command `python ./setup.py internal_docs` to build the docs, I then make a change to the class doc for the ArithmeticFilter class and rebuild the docs using the same command. It doesn't seem to be rebuilding `mantidimaging/docs/build/html/user_guide/operations/index.html` though - am I missing a step somewhere?

samtygier-stfc:
It might need to build once to record the deps. Maybe something like this

On main:
- remove docs/build
- python ./setup.py internal_docs
- then edit a doc string
- python ./setup.py internal_docs (wont rebuild because its only check the rst file)
- html will not be updated

On this branch
- remove docs/build
- python ./setup.py internal_docs (this should record the dependencies in docs/build)
- then edit a doc string
- python ./setup.py internal_docs (should notice the change)
- html will be updated

PR #1531, opened 2022-07-05
Operations window lock scale and lock zoom settings selected by default

### LINKED ISSUE #1523: Ops window Lock Scale and Lock Zoom should be enabled by default
### Desired Behaviour

The Lock Scale and Lock Zoom tickboxes in the operations window should be enabled by default.

Request from users.

Note: will need new base lines for a lot of the screenshot tests.

rbauststfc:
### Issue

Closes #1523

### Description

Sets the lock scale and lock zoom checkboxes to be selected by default when the operations window opens.

Setting lock scale to be selected by default had an impact on the window first opening, as the histogram region to lock was being selected before the image had been drawn and was therefore so small that the image didn't appear. As a result, this PR  only starts applying the lock scale after the first previews have been drawn.

There didn't seem to be an issue with locking the zoom when the window first opened, however I thought we should handle the edge case where the user has deleted the tree view data with the ops window still open. In this case it seemed sensible for everything to refresh when the new data is loaded.

In testing I noticed that if you switch between stacks with a very different histogram range (i.e. between the flowers and phantom spheres) while the lock scale is on you can lose sight of the histogram range altogether. I experimented with `autoHistogramRange` to ensure the range was always in view while maintaining the locked region. However, I decided against using this because too many automatic adjustments could interfere with users that decide to zoom in to a region of the histogram, for example. There is a right click menu "View All" option for the histogram that brings the range back into view if needed.

I also noticed in testing that having lock scale on by default when we opened the window meant that the difference preview histogram scale could be very small for some operations e.g. with the flowers dataset when changing to a Gaussian filter after opening the window on Crop Co-ordinates. Some of the operations, like Crop Co-ordinates, don't provide a difference image so, to get around this, the lock is only applied to the difference image histogram when there was previously a difference image preview.

There are quite a lot of different ways the histograms can behave with lock scale on depending on the sequence of steps the users follow. I wouldn't recommend making this solution more complex until we have a clear steer from the users on the scenarios they encounter most frequently and how they need us to handle them.

The operations window should now open with the lock scale and lock zoom options checked by default. These should function in a useful way when switching between stacks and between operations.
If all data is deleted from the tree view with the window still open, the preview zoom and histogram scales should refresh when the first data is loaded again.

### Documentation


rbauststfc:
More information needed to figure out complications with lock scale - see details in the document linked to the issue.

PR #1535, opened 2022-07-08
Lock zoom by default in operations window

rbauststfc:
### Issue

Refs #1523

### Description

Selects the lock zoom checkbox by default when the operations window first opens. This is just one part of the referenced issue as further information is required to implement the lock scale side. I've set the previews to auto-range when data is first loaded as, although all the datasets I tested with had visible previews without auto-ranging, we seem to get some slight jumping around in size without it.

This PR also includes a change so that the operations window previews will auto-range (the zoom will reset) when the lock zoom checkbox is de-selected.

Operations window opens with lock zoom selected by default.
De-selecting lock zoom causes the zoom to reset (all three previews auto-range).

### Documentation

Lock zoom part of the issue is added to the release notes.

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -19,6 +19,7 @@ New Features
 - #1444 : CIL PDHG non-negativity constraint
 - #1483 : Add line profile to reconstruction window
 - #1480 : Automatic sinograms for sinogram operations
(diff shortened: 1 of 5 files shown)


samtygier-stfc (approved):
Works well.

PR #1538, opened 2022-07-12
Fix reconstruction double click error

### LINKED ISSUE #1351: Double clicking reconstruct buttons can cause RunTimeError
### Summary

When running the first reconstruction for a dataset, if the user double clicks either of the reconstruct buttons in the reconstruction window a RunTimeError occurs. I think this is because the double click creates two recon stacks in the dataset, but the code in method `add_recon_item_to_tree_view` (in main/presenter.py) only creates the recon group for the tree view when there is exactly one recon stack. If there is more than one the code tries to look up an existing recon group that doesn't exist, resulting in a RunTimeError.

### Steps To Reproduce

The behaviour occurs with either reconstruct button, but is easier to reproduce with the Reconstruct Volume button (you don't seem to need to double click quite as quickly):

1) Load a dataset and open the reconstruction window
2) Set the centre of rotation and go to the Reconstruct tab
3) The effect can be seen with any algorithm, but choosing gridrec will make the test quicker
4) Double click the Reconstruct Volume button

### Expected Behaviour

Ideally it should only be possible to click either of the reconstruct buttons once (i.e. they should disable immediately after the first click and only re-enable once the reconstruction has completed). If this isn't sufficient to stop it running through twice (maybe someone could double click quickly enough that the button hasn't disabled yet?) then the system should just display the two recon stacks in the tree view.

### Current Behaviour

Two recon stacks are created in the dataset, but the code that tries to display them in the tree view throws a RunTimeError. This error then continues when the user tries to run other reconstructions on the dataset, regardless of whether they double click the buttons. The dataset has to be deleted and re-loaded to be able to continue.

### Context

Current main

### Failure Logs

```
2022-03-03 09:27:57,280 [mantidimaging.gui.mvp_base.presenter:L24] ERROR: Presenter error: Unable to find recon group in dataset tree item for dataset 8b529d36-be18-414d-b97d-3686bf73f38c
  File "C:\Mantid\mantidimaging\mantidimaging\gui\windows\main\presenter.py", line 82, in notify
    self._add_recon_to_dataset(**baggage)
  File "C:\Mantid\mantidimaging\mantidimaging\gui\windows\main\presenter.py", line 592, in _add_recon_to_dataset
    self.add_recon_item_to_tree_view(parent_id, recon_data.id, len(self.model.datasets[parent_id].recons))
  File "C:\Mantid\mantidimaging\mantidimaging\gui\windows\main\presenter.py", line 540, in add_recon_item_to_tree_view
    recon_group = self.view.get_recon_group(dataset_item)
  File "C:\Mantid\mantidimaging\mantidimaging\gui\windows\main\view.py", line 546, in get_recon_group
    raise RuntimeError(f"Unable to find recon group in dataset tree item for dataset {dataset_item.id}")
RuntimeError: Unable to find recon group in dataset tree item for dataset 8b529d36-be18-414d-b97d-3686bf73f38c
```

rbauststfc:
### Issue

Closes #1351

### Description

Previously the system expected that if a dataset had multiple recon stacks then a recon tree view group would already have been created, and threw an error if it couldn't find one. This PR alters that behaviour to handle the edge case where a user double clicks the recon button and multiple stacks are then created before the tree view group. In that scenario, the system will now create the recon group and display all the stacks.

Clicking the reconstruction button a second time after the recon has started creates duplicate stacks that aren't needed and could cause a long-running reconstruction to unintentionally be run twice. Therefore this PR also implements enabling and disabling the reconstruction buttons so that only one reconstruction can be triggered at a time. This appears to work well on Linux, however on Windows it still seems possible to click the disabled button and end up with a duplicate for very short reconstructions. I've spent a long time looking into this and haven't managed to find a solution, so I will open a separate issue to document what I've found. It still seemed worth leaving the functionality in, as preventing duplicate runs of long reconstructions seems desirable. I suspect it's unlikely that Windows users will notice the problem very regularly, if at all.

See steps on issue for replicating the original problem. Double clicking either of the reconstruction buttons should now either display all recon data that was created or, depending on timings, may result in only one recon stack being created. There should be no errors.

Depending on the reconstruction length and the OS, you should see both reconstruction buttons disable when the reconstruction is running and then enable again when it has finished. For very fast reconstruction algorithms on Linux (i.e. gridrec) the effect may be barely noticeable.

### Documentation

### DIFF
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -50,7 +50,7 @@ class Notification(Enum):
 
 
 def _generate_recon_item_name(recon_no: int) -> str:
-    return f"Recon {recon_no}"
+    return "Recon" if recon_no == 1 else f"Recon {recon_no}"
 
 
 class MainWindowPresenter(BasePresenter):
@@ -524,12 +524,15 @@ class MainWindowPresenter(BasePresenter):
         :param recon_count: The number of the recon in the dataset. One indicates the first recon that has been added.
         """
         dataset_item = self.view.get_dataset_tree_view_item(parent_id)
-        if recon_count == 1:
+        recon_group = None if recon_count == 1 else self.view.get_recon_group(dataset_item)
+
+        if not recon_group:
             recon_group = self.view.add_recon_group(dataset_item, self.model.get_recon_list_id(parent_id))
-            name = "Recon"
+            recon_num = 1
         else:
-            recon_group = self.view.get_recon_group(dataset_item)
-            name = _generate_recon_item_name(recon_count)
+            recon_num = recon_count
+
+        name = _generate_recon_item_name(recon_num)
         self.view.create_child_tree_item(recon_group, child_id, name)
 
     def add_stack_to_dictionary(self, stack: StackVisualiserView) -> None:
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -49,10 +49,6 @@ class Notification(Enum):
     ADD_RECON = auto()
 
 
-def _generate_recon_item_name(recon_no: int) -> str:
-    return "Recon" if recon_no == 1 else f"Recon {recon_no}"
-
-
 class MainWindowPresenter(BasePresenter):
     LOAD_ERROR_STRING = "Failed to load stack. Error: {}"
     SAVE_ERROR_STRING = "Failed to save data. Error: {}"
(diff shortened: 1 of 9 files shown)


mantidimaging/gui/windows/main/presenter.py
```diff
@@ -524,12 +524,15 @@ def add_recon_item_to_tree_view(self, parent_id: uuid.UUID, child_id: uuid.UUID,
         :param recon_count: The number of the recon in the dataset. One indicates the first recon that has been added.
         """
         dataset_item = self.view.get_dataset_tree_view_item(parent_id)
-        if recon_count == 1:
+        recon_group = None if recon_count == 1 else self.view.get_recon_group(dataset_item)
```
samtygier-stfc, line 527: Could we simplify this by not passing `recon_count` as an argument, and just checking if there is an existing recon_group?

I think we could also avoid generating the name here, and use the stack name.
`self.model.get_images_by_uuid(child_id).name`
(I don't think it matters that name from the stack has an underscore in it).

samtygier-stfc (dismissed):
Looks good.

`add_recon_item_to_tree_view` feels a bit awkward, maybe it can be simplifies where we are here.

samtygier-stfc (approved):

PR #1539, opened 2022-07-12
Spectrum veiwer CSV output

samtygier-stfc:
### Issue

Closes 1537

### Description

### Testing 

### Acceptance Criteria 

### Documentation

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/view.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/view.py
@@ -1,8 +1,9 @@
 # Copyright (C) 2022 ISIS Rutherford Appleton Laboratory UKRI
 # SPDX - License - Identifier: GPL-3.0-or-later
+from pathlib import Path
 from typing import TYPE_CHECKING, Optional
 
-from PyQt5.QtWidgets import QCheckBox, QVBoxLayout
+from PyQt5.QtWidgets import QCheckBox, QVBoxLayout, QFileDialog, QPushButton
 
 from mantidimaging.gui.mvp_base import BaseMainWindowView
 from mantidimaging.gui.widgets.dataset_selector import DatasetSelectorWidgetView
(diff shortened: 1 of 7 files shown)


mantidimaging/gui/windows/spectrum_viewer/view.py
```diff
@@ -72,3 +76,10 @@ def try_to_select_relevant_normalise_stack(self, name: str) -> None:
 
     def get_normalise_stack(self) -> Optional['UUID']:
         return self.normaliseStackSelector.current()
+
+    def get_csv_filename(self) -> Optional[Path]:
```
rbauststfc, line 80: Just a thought, but if we ever wanted to re-use the CSV output functionality then it's likely we would need to replicate this method in other views. Would there be any benefit to making this a static method in the `CSVOutput` class instead?

rbauststfc (approved):
Works well and the code looks good. I've made a small suggestion but feel free to ignore it and merge as is if you prefer. 🙂

I noticed a couple of small things that would be good to add to the to do list for the spectrum viewer, but not needed for this PR:

1) Would it be worth providing a default CSV output name in the save dialog so the user doesn't haven't to provide their own if they don't want to?

1) Not related to CSV saving, but I noticed if I switched stacks between two images of different sizes the image view doesn't appear to be auto-ranging (I first selected the flowers 512 x 512 and then selected the powder dataset).

mantidimaging/gui/windows/spectrum_viewer/view.py
samtygier-stfc: I don't think it belongs in core because it use Qt.

If we started using it more places BaseMainWindowView might be a better place. There might be a bit more that can be shared between the windows.

rbauststfc: Yes, I wondered about that but then checked core and it looks like we're using Qt in some of the operations there. If the intention is actually to keep Qt out of core, though, then better off leaving this where it is.

samtygier-stfc:
I've added the other bits to the project board https://github.com/orgs/mantidproject/projects/2/views/1

PR #1549, opened 2022-07-19
Spectrum CSV normalised output

### LINKED ISSUE #1545: CSV output of normalised spectrum
CSV should have sample, open and normalised values.

samtygier-stfc:
### Issue

closes #1545

### Description

This updates `save_csv()` to output the open beam and normalized values. `get_spectrum` is now controlled by the arguments instead of state in the model. The normalisation mode is now stored in the presenter as an Enum.

I also merged getting the whole spectrum with getting the RoI spectrum, by having the model keep a dict of RoIs, and creating and `all` roi. This should make it simple to support additional RoIs in the future.

I had an intermediate version d30204437bc904a1f522acc6c641070162a1eada, where `normalised` was still a bool, just moved to the presenter, but deceided an Enum was clearer (and allows some flexibility for additional normalisation modes).

I'm not completely happy with how the presenter maps back and forth between the bool of the normalised checkbox and the spectrum_mode. I.e. I'm not a fan of having to do `self.model.save_csv(path, self.spectrum_mode == SpecType.SAMPLE_NORMED)`. If the model holds a bool, then it makes the `get_spectrum()` calls more complicated. If it holds a bool and spectrum mode, then that is 2 values that could get out of sync. I'm open to ideas here.

Load up the 4 powders sample and open stacks
Open the Spectrum Viewer
Resize the ROI
Export the CSV
Check that the CSV file opens in a spreadsheet

Enable the Normalisation checkbox, and select the open beam stack
Export the CSV
Check that the CSV file opens in a spreadsheet, there should now be `open` and `norm` columns for all and roi.

### Documentation
Once all spectrum work is done

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -10,6 +10,8 @@ from mantidimaging.core.data import ImageStack
 from mantidimaging.core.io.csv_output import CSVOutput
 from mantidimaging.core.utility.sensible_roi import SensibleROI
 
+ALL = "all"
+
 if TYPE_CHECKING:
     from mantidimaging.gui.windows.spectrum_viewer.presenter import SpectrumViewerWindowPresenter
 
@@ -34,7 +36,7 @@ class SpectrumViewerWindowModel:
         self._stack = stack
         self.tof_range = (0, stack.data.shape[0] - 1)
         height, width = self.get_image_shape()
-        self.set_roi("all", SensibleROI.from_list([0, 0, width, height]))
+        self.set_roi(ALL, SensibleROI.from_list([0, 0, width, height]))
         self.set_roi("roi", SensibleROI.from_list([0, 0, width, height]))
 
     def set_normalise_stack(self, normalise_stack: ImageStack) -> None:
@@ -90,7 +92,7 @@ class SpectrumViewerWindowModel:
         csv_output = CSVOutput()
         csv_output.add_column("tof_index", np.arange(self._stack.data.shape[0]))
 
-        for roi_name in ("all", "roi"):
+        for roi_name in (ALL, "roi"):
             csv_output.add_column(roi_name, self.get_spectrum(roi_name, SpecType.SAMPLE))
             if normalized:
                 if self._normalise_stack is None:
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -1,5 +1,6 @@
 # Copyright (C) 2022 ISIS Rutherford Appleton Laboratory UKRI
 # SPDX - License - Identifier: GPL-3.0-or-later
+from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, Optional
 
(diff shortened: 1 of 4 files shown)


rbauststfc:
It all seems to be working well. 🙂 I'm having a look through the code and just had a quick question at this stage - the code in the `get_spectrum` method is supporting an "OPEN" mode, but I can't see anywhere in the code that could actually request that mode. Is that something for the future and, if so, how might it work (just at a high-level)?

samtygier-stfc:
> the code in the `get_spectrum` method is supporting an "OPEN" mode, 

It is used in `SpectrumViewerWindowModel.save_csv()` so that the spectrum from the `_normalise_stack` can be written to the CSV.

rbauststfc:
> > the code in the `get_spectrum` method is supporting an "OPEN" mode,
> 
> It is used in `SpectrumViewerWindowModel.save_csv()` so that the spectrum from the `_normalise_stack` can be written to the CSV.

Ah yes, sorry, I think I spotted that on the first read through and then forgot about it after.

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -28,50 +34,69 @@ def set_stack(self, stack: ImageStack) -> None:
...
             return None
 
-    def get_spectrum(self) -> Optional['np.ndarray']:
+    @staticmethod
+    def get_stack_spectrum(stack: ImageStack, roi: SensibleROI):
+        left, top, right, bottom = roi
+        roi_data = stack.data[:, top:bottom, left:right]
+        return roi_data.mean(axis=(1, 2))
+
+    def get_spectrum(self, roi_name: str, mode: SpecType) -> Optional['np.ndarray']:
```
rbauststfc, line 62: A thought on the issue of mapping back and forth between a spectrum type, I wonder if we could resolve that by splitting up this method? So you would have `get_sample_spectrum`, `get_open_spectrum` and `get_sample_norm_spectrum`, keeping the `get_stack_spectrum` method to use within each of those. I think this would have the knock on effect that the presenter could just call the relevant method instead of having to pass in a type. In the places where the presenter needs to establish if we're normalising or not, could the view provide a method to get the status of the normalise checkbox and then the presenter could call that when it needs to find out if we're normalising instead of holding the state itself?

rbauststfc: Or the presenter could hold the boolean state that gets passed to `handle_enable_normalised` if we prefer not to keep looking it up.

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -28,50 +34,69 @@ def set_stack(self, stack: ImageStack) -> None:
         self._stack = stack
         self.tof_range = (0, stack.data.shape[0] - 1)
         height, width = self.get_image_shape()
-        self.roi_range = SensibleROI.from_list([0, 0, width, height])
+        self.set_roi("all", SensibleROI.from_list([0, 0, width, height]))
+        self.set_roi("roi", SensibleROI.from_list([0, 0, width, height]))
```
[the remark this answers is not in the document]
samtygier-stfc: Maybe for `"all"`, but for `"roi"` I think there will be multiple ROIs in the future, possible with user controllable names.

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: Maybe, though it might move more complexity into the presenter. I think I'll come back to this later.

rbauststfc (approved):

rbauststfc (approved):

PR #1553, opened 2022-07-21
Spectrum viewer: handle changing stacks

### LINKED ISSUE #1550: Spectrum viewer handle switching stacks
A bit more needs to be done when switching stacks.

Auto Zoom if sized changed

Auto levels?

Handle switching to None

Handle selected stack being closed in main window.

samtygier-stfc:
### Issue

Closes #1550 

### Description

Handle some cases for changing stacks

- Make `SpectrumViewerWindowModel.set_stack()` handle None properly (Happens when stacks are closed in main window)
- Invert some `is not None` into `is None` guards
- Clear the view when setting stack to None
- Handle case where the `normalise_uuid` is not an open stack. Can happen momentarily while stacks are closing.
- Don't do anything in `handle_sample_change` and `handle_normalise_stack_change` if the stack has not actually changed.
- Implement a warning icon if normalisation cannot be done.
- Return an empty array for a normalisation spectrum if it can't be performed.

For all these cases the user should not see any exception messages

1)
Open dataset, open spectrum viewer.
With spectrum viewer still open, close the dataset in the main window.

2)
Open dataset, open spectrum viewer.
With spectrum viewer still open, close just the selected stack in the main window.

3)
Open dataset, open spectrum viewer.
With spectrum viewer still open, close just the normalise (flat/open) stack in the main window.

4)
Open 2 datasets with different dimensions
Open spectrum viewer, try setting the the normalisation stack to a stack from the other dataset
A warning icon should show, with a tool tip explaining the issue

5)
Open 2 datasets of different sizes.
Open spectrum viewer
Switch back and forth between different stacks for the sample
Check that the zoom is reset when the new stack is selected
Check that zoom does not get reset by changing the ToF range selector

### Documentation

Not needed

### DIFF
--- a/mantidimaging/gui/ui/spectrum_viewer.ui
+++ b/mantidimaging/gui/ui/spectrum_viewer.ui
@@ -86,17 +86,40 @@
          </spacer>
         </item>
         <item>
-         <widget class="QCheckBox" name="normaliseCheckBox">
-          <property name="layoutDirection">
-           <enum>Qt::LeftToRight</enum>
-          </property>
-          <property name="text">
-           <string>Normalise to open beam</string>
-          </property>
-          <property name="checked">
-           <bool>false</bool>
-          </property>
-         </widget>
+         <layout class="QHBoxLayout" name="horizontalLayout">
+          <item>
+           <widget class="QCheckBox" name="normaliseCheckBox">
+            <property name="layoutDirection">
+             <enum>Qt::LeftToRight</enum>
+            </property>
+            <property name="text">
+             <string>Normalise to open beam</string>
+            </property>
+            <property name="checked">
+             <bool>false</bool>
+            </property>
+           </widget>
+          </item>
+          <item>
+           <widget class="QLabel" name="normaliseErrorIcon">
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
+            <property name="minimumSize">
+             <size>
+              <width>32</width>
+              <height>32</height>
+             </size>
+            </property>
+            <property name="text">
+             <string/>
+            </property>
+           </widget>
+          </item>
+         </layout>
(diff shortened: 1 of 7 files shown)


rbauststfc:
I've just had a play around with the functionality so far, but it looks good in terms of preventing exceptions. A couple of things I noticed:

1) When performing test 1, if you delete a dataset while an error is being displayed then the error icon isn't removed. Would it be worth doing that?

1) Should error icons only display if the normalise checkbox is selected? (this is made a bit more confusing by the bug on the project board about the normalise checkbox being deselected when the window first opens but the dropdown still being enabled, so the GUI appearance starts off out of sync)

1) When I was testing it initially took me a little while to get the tooltip to appear, and I had the advantage of knowing it was there to be found. There's a lot of space still available in the left hand panel, would it be worth displaying the error message somewhere there instead of in a tooltip?

samtygier-stfc:
1) I think that is ok, because normalisation can't be done without a stack

2) Yes, I like this idea

3) I think a tooltip is ok for now, it matches the indicators in the operations window that have a message on mouse over.

rbauststfc:
Zoom changes seem to be working well. I noticed that I get a RuntimeError if I load the powders dataset, select the normalise checkbox, then delete the dataset and reload it with the window still open. On reload, the following error is thrown:

```
2022-08-01 08:39:40,673 [mantidimaging.gui.windows.main.view:L453] ERROR: Traceback (most recent call last):
  File "\mantidimaging\mantidimaging\gui\windows\spectrum_viewer\presenter.py", line 57, in handle_sample_change
    self.show_new_sample()
  File "\mantidimaging\mantidimaging\gui\windows\spectrum_viewer\presenter.py", line 89, in show_new_sample
    self.view.set_spectrum(self.model.get_spectrum("roi", self.spectrum_mode), clear_all=True)
  File "\mantidimaging\mantidimaging\gui\windows\spectrum_viewer\model.py", line 84, in get_spectrum
    raise RuntimeError("No normalisation stack selected")
RuntimeError: No normalisation stack selected
```

rbauststfc (approved):

PR #1571, opened 2022-07-29
Prevent Reconstruction window previews updating while the window is closed

### LINKED ISSUE #1562: Reconstruction previews updated when new dataset is loaded with window closed
### Summary

If the Reconstruction window has been opened and then closed, deleting and then loading a new dataset still seems to trigger an update of the recon previews. It's not noticeable unless you have a reconstruction algorithm selected that takes long enough to display a progress dialog, like PDHG.

### Steps To Reproduce

1) Load a dataset.
1) Open the Reconstruction window and select PDHG as the recon algorithm. Set the iterations to something like 10 to ensure the preview generation takes long enough for this test.
1) Close the Reconstruction window.
1) Delete the dataset from the main window.
1) Load a dataset.

### Expected Behaviour

The dataset should load without the reconstruction previews being updated.

### Current Behaviour

The dataset loads and then a progress dialog for CIL PDHG appears. The dialog disappears once the reconstruction appears to have completed.

### Context

Current main, but has presumably been an issue since PDHG was added. The problem exists in version 2.3.

rbauststfc:
### Issue

Closes #1562

### Description

The problem was being caused because when a stack is loaded it causes a `model_changed` signal to be emitted. This triggers the DatasetSelectorWidget to reload it's dataset, which in turn emits a `stack_selected_uuid` signal. In the recon window, this triggers a call to method `set_stack_uuid`, which wasn't checking if the window was visible before refreshing the previews. This PR adds that check using the same approach as we already use when a stack is changed. If both the `stack_changed` and `model_changed` signals have been emitted while the reconstruction window was closed then it's the change in the selected stack (i.e. the `model_changed` signal) that will be applied.

See steps given on issue.
Check that the preview is still updating when needed.

### Documentation

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -34,6 +34,7 @@ Fixes
 - #1496 : OutliersFilter IndexError when using sinograms
 - #1515 : Unlink axis in recon window
 - #1351 : Double clicking reconstruct buttons can cause RunTimeError
+- #1562 : Stop reconstruction previews updating when window is closed
 
 
 Developer Changes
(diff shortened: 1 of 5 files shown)


samtygier-stfc (approved):
Thanks, fixes the issue.

samtygier-stfc (approved):

PR #1572, opened 2022-08-01
Update installation doc

### LINKED ISSUE #1563: Remove compiler (gcc) from installtion docs
### Desired Behaviour

The installation docs currently say
"gcc - needed for compiling some python modules"
~~This should be updated to say a C++ compiler. and recommend gcc on linux, and Microsoft Visual Studio Build Tools on windows.~~

Since #1565 a compiler is no longer needed for installation of mantidimaging or the dev environment on windows or linux. So this can be removed from the installation and dev guide pages

Dagonite:
### Issue

Closes #1563

### Description

Remove GCC as a required dependency for the code installation from the documentation.

### DIFF
--- a/docs/developer_guide/index.rst
+++ b/docs/developer_guide/index.rst
@@ -4,7 +4,7 @@ Developer Guide
 Getting Started
 ---------------
 
-You will need conda/mamba, CUDA and gcc installed, see the :ref:`Installation` instructions.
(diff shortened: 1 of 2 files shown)


samtygier-stfc:
There is also a mention of gcc in `docs/developer_guide/index.rst` that can be removed.

samtygier-stfc (approved):

PR #1584, opened 2022-08-12
Output zeros for missing rotation angle in NeXus save

### LINKED ISSUE #1556: Error when saving NeXus output
### Summary

The system raises a RuntimeError when trying to save the Flowers 512 x 512 dataset as a NeXus file.

### Steps To Reproduce

1) Load the Flowers 512 x 512 dataset
2) Click File -> Save as NeXus File

### Expected Behaviour

The file is saved successfully without an error.

### Current Behaviour

The system raises a RuntimeError, which is displayed in a dialog. A file seems to be output, although I haven't checked if the contents are correct.

### Context

Current main

### Failure Logs

```
File "/mantidimaging/gui/dialogs/async_task/task.py", line 49, in run
  File "/mantidimaging/core/utility/func_call.py", line 11, in call_with_known_parameters
    return func(**ka)
  File "/mantidimaging/gui/windows/main/model.py", line 90, in do_nexus_saving
    saver.nexus_save(self.datasets[dataset_id], path, sample_name)  # type: ignore
  File "/mantidimaging/core/io/saver.py", line 183, in nexus_save
    _nexus_save(nexus_file, dataset, sample_name)
  File "/mantidimaging/core/io/saver.py", line 233, in _nexus_save
    rotation_angle = sample_group.create_dataset("rotation_angle", data=np.concatenate(dataset.nexus_rotation_angles))
  File "/mantidimaging/core/data/dataset.py", line 138, in nexus_rotation_angles
    raise RuntimeError("Incomplete rotation angle data.")
RuntimeError: Incomplete rotation angle data.
2022-07-21 15:26:25,533 [mantidimaging.gui.windows.main.presenter:L174] ERROR: Failed to save data. Error: Incomplete rotation angle data.
2022-07-21 15:26:27,816 [mantidimaging.gui.mvp_base.presenter:L24] ERROR: Presenter error: Failed to save data. Error: Incomplete rotation angle data.
```

rbauststfc:
### Issue

Closes #1556

### Description

This is a short-term solution that it's likely we will want to re-work in future to allow more user choice. For now, though, we agreed with the users that we would output 0 in the NeXus file where rotation angles are missing. This also means we need to update the NeXus file loader to interpret a file with all zeros as not providing any rotation angles.

The way I've written the code handles this on a stack by stack basis, so if one of the stacks has missing projection angles then only that stack will have zeros written to the file, the rest of the stacks will retain any angle information. Similarly when loading, if a stack has all zero angles then it will have no projection angles set in MI, but other stacks created from the file could have angles if that information is present.

I'm not sure how likely it is that we would get a mixed file like this - it occurred to me because our test flowers dataset has angle information for the sample and flat, but not the dark images. If you loaded all the available information for that dataset into MI and then saved as NeXus, we would lose angle information if we didn't handle it stack by stack. The other approach would be to assume if any single stack doesn't have angle information then we should save zeros for all stacks - if this would be a more helpful assumption then I'm happy to refactor the changes in this PR.

I've tried to add sufficient unit test coverage for the different possibilities. I have also manually tested loading datasets with full, partial and no angle information and checked that a NeXus file is output that we're able to successfully re-load into MI.

I'm not sure how to view the contents of a NeXus file, so I've not been able to manually check the angle information being written to the file. I've done what I can with temporarily adding print/log statements to MI to check what is being output, however if there's a way to check the file contents then that would be helpful to know about.

### Documentation

Not required, fixes functionality introduced since the last release.

### DIFF
--- a/mantidimaging/core/data/dataset.py
+++ b/mantidimaging/core/data/dataset.py
@@ -139,6 +139,10 @@ class StrictDataset(BaseDataset):
         return [angles.value for angles in proj_angles]  # type: ignore
 
     @property
+    def num_nexus_angles_required(self) -> int:
+        return sum(image_stack.num_images for image_stack in self._nexus_stack_order)
+
+    @property
     def image_keys(self) -> List[int]:
         image_keys = []
         if self.dark_before is not None:
--- a/mantidimaging/core/data/dataset.py
+++ b/mantidimaging/core/data/dataset.py
@@ -133,14 +133,11 @@ class StrictDataset(BaseDataset):
 
     @property
     def nexus_rotation_angles(self) -> List[np.ndarray]:
-        proj_angles = [image_stack.real_projection_angles() for image_stack in self._nexus_stack_order]
-        if None in proj_angles:
-            raise RuntimeError("Incomplete rotation angle data.")
-        return [angles.value for angles in proj_angles]  # type: ignore
-
-    @property
-    def num_nexus_angles_required(self) -> int:
-        return sum(image_stack.num_images for image_stack in self._nexus_stack_order)
+        proj_angles = []
+        for image_stack in self._nexus_stack_order:
+            angles = image_stack.real_projection_angles()
+            proj_angles.append(angles.value if angles else np.zeros(image_stack.num_images))
+        return proj_angles
 
     @property
     def image_keys(self) -> List[int]:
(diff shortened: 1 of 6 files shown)


samtygier-stfc (approved):
Looks good. Thanks

Issue #1427, opened 2022-04-28
Pytest sometimes very slow


samtygier-stfc:
### Summary

The pytest step on git hub actions is usually below a minute, but sometimes is much slower, exceeding timeouts.

#1424 disabled the parallel running, so that we can see which test was being run when the timeout was reached, but this was not consistent. #1426 increases the timeouts to gather more debug information. Also I have added debug flags from https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging

Some slow runs
https://github.com/mantidproject/mantidimaging/runs/6207333156?check_suite_focus=true
https://github.com/mantidproject/mantidimaging/runs/6197532458?check_suite_focus=true

Slowest tasks were 
``` 
============================ slowest 10 durations =============================
26.83s teardown mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_collapse
23.50s call     mantidimaging/gui/windows/stack_choice/tests/view_test.py::StackChoiceViewTest::test_setup_stack_for_view
20.80s call     mantidimaging/gui/windows/stack_choice/tests/view_test.py::StackChoiceViewTest::test_connect_histogram_changes
19.70s call     mantidimaging/gui/windows/stack_choice/tests/view_test.py::StackChoiceViewTest::test_ensure_range_is_the_same_new_stack_min_original_stack_max
18.58s call     mantidimaging/gui/windows/stack_choice/tests/view_test.py::StackChoiceViewTest::test_sync_roi_plot_for_new_stack_with_old_stack
17.56s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_collapse
17.46s call     mantidimaging/gui/windows/stack_choice/tests/view_test.py::StackChoiceViewTest::test_closeEvent
16.61s call     mantidimaging/gui/windows/stack_choice/tests/view_test.py::StackChoiceViewTest::test_toggle_roi_show_true
16.37s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_on_auto_update_triggered_with_auto_not_selected
15.70s call     mantidimaging/gui/windows/stack_choice/tests/view_test.py::StackChoiceViewTest::test_images_are_synced
========= 871 passed, 104 skipped, 290 warnings in 471.92s (0:07:51) ==========
```
another slow one
```

============================ slowest 10 durations =============================
[...]

PR #1586, opened 2022-08-12
Update pytest and coverage versions

rbauststfc:
### Issue

Refs #1427

### Description

Increases pytest and coverage to the latest versions and installs them from Conda instead of pip.

I'm not closing the issue because we'll need to monitor the effect of these changes. Locally I've not been able to replicate a failure since updating to version 6.4 of coverage. Reading around, I found several reports of issues with freezing tests when using pytest-cov/coverage and multiprocessing. Digging further, there was mention of specific issues with coverage version 6.3. Because of the intermittent nature of the problem it will take some time to be able to confirm what effect these version changes have, if any, but it seemed worth a try.

I found that running our full GitHub Actions command with a specific seed (see command below) seemed to reproduce the freeze behaviour quite regularly locally, although still with some longish periods of successful runs:

`xvfb-run --auto-servernum python -m pytest --cov --cov-report=xml -n auto --ignore=mantidimaging/eyes_tests --durations=10 --randomly-seed=2998623717`

I was able to rule out the following as having fixed the issue:
- Simply updating the pytest versions
- Running without `-n auto`, regardless of pytest version. I also found that when running without `-n auto` there was no consistency in which test was the point of failure

The following haven't yet reproduced failures locally for me:
- Running with existing pytest and coverage versions but removing `--cov --cov-report=xml`
- Updating the pytest **and** coverage versions

### Documentation

I've added an entry to the release notes to document that the pytest and coverage versions have been increased.


samtygier-stfc (dismissed):

samtygier-stfc (approved):

samtygier-stfc:
I saw a Windows GUI system test timeout on the PR (this is possibly a different issue), but no timeouts in the pytest. A useful update even if it does not 100% solve the issue.

PR #1599, opened 2022-09-09
Improve usability of ROI selectors and default sizes

### LINKED ISSUE #1330: Initial ROI box is sometimes much larger than image
### Desired Behaviour

Starting box is the same size as/smaller than the image.

### Current Behaviour

You sometimes get things like this and have to drag it until the bottom right corner is visible in order to resize the box: 
image

rbauststfc:
### Issue

Closes #1330

### Description

Various fixes so that the default size of an ROI in the `MIImageView` should be equal to either the top left quadrant of the image or 20 x 20, whichever is larger. This is relevant for various stack visualisers through the application, including the main window.

For the ROI selectors on the operations window (used for Crop Co-ordinates and ROI Normalisation), the initial ROI is now set based on the value in the ROI field. If no valid value has been set then the selector uses the default for the `MIImageView`. After Crop Co-ordinates has been applied, the ROI field is reset to use the new image size and ensure the ROI size remains relevant.

I had a go at having the ROI selector dialog update in response to the user changing values in the ROI field while the dialog was open. This introduced some complexity to the code that seemed unnecessary given that I think it's reasonable to assume users will choose just one method of selecting the ROI (as they have to now). With this in mind, I've set the ROI field to be disabled when any ROI selector dialogs are opened and re-enabled when all are closed.

- Check the problems described on the issue.
- Additionally check that moving the ROI in a selector dialog updates the ROI field and also that moving the ROI updates the information at the bottom of the `MIImageView` image.
- Check that the ROI is adjusted correctly when a Crop Co-ordinates filter is applied, both in the Operations window and in the main window stack visualiser.

I'm expecting the Applitools ROI Normalisation test to fail because the value in the ROI field should now be set to the image size when the image size is less than 200 x 200.

### Documentation

Issue added to release notes.

### DIFF
--- a/mantidimaging/gui/windows/operations/view.py
+++ b/mantidimaging/gui/windows/operations/view.py
@@ -1,6 +1,6 @@
 # Copyright (C) 2022 ISIS Rutherford Appleton Laboratory UKRI
 # SPDX - License - Identifier: GPL-3.0-or-later
-
+import functools
 from typing import TYPE_CHECKING
 
 import numpy as np
@@ -65,6 +65,7 @@ class FiltersWindowView(BaseMainWindowView):
         self.presenter = FiltersWindowPresenter(self, main_window)
         self.roi_view = None
         self.roi_view_averaged = False
+        self._roi_selector_window_count = 0
         self.splitter.setSizes([200, 9999])
         self.splitter.setStretchFactor(0, 1)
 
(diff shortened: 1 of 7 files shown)


mantidimaging/gui/windows/operations/view.py
```diff
@@ -242,6 +243,8 @@ def roi_visualiser(self, roi_field):
             # Happens if nothing has been loaded, so do nothing as nothing can't be visualised
             return
 
+        self._roi_selector_window_count += 1
```
samtygier-stfc, line 246: We should probably prevent multiple ROI windows from opening, looks like it is a bit more work disable the button, so I'll make a new issue for this.

samtygier-stfc (approved):

mantidimaging/gui/windows/operations/view.py
samtygier-stfc: I have opened #1602

PR #1611, opened 2022-10-06
Update tomopy and remove work arounds

### LINKED ISSUE #1597: Remove work around in RingRemovalFilter
### Current Behaviour

In #1062 a work around was add to RingRemovalFilter to prevent a crash by raising a ValueError if the centre is closer to the edge of the image than theta_min. This crash is now fixed in tomopy.

### Desired Behaviour

Remove the check from RingRemovalFilter as it is now done in tomopy.

The range of theta min in the gui should also be changed to 0-179

samtygier-stfc:
### Issue
Closes #1597 

### Description

Update tomopy
Remove workaround
Limit Theta spinbox

With `tilt_simple_recon.tar.xz` from the Onedrive example data.
Open the Ring removal, use the values from the screenshot. There should be some reduction in ring artefacts (its a bit subtle).

image

If you change Center of Rotation to manual, and put in some small values (e.g. 2,2) then there should no longer be an error message.

The Theta input should now be limited to 0-179

### Documentation

Not needed

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -25,7 +25,7 @@ requirements:
     - scikit-image=0.19.*
     - numpy=1.20.*
     - algotom=1.0.*
-    - tomopy=1.10.*
+    - tomopy=1.12.*
     - cudatoolkit=10.2*
     - cupy=10.2.*
     - astra-toolbox=2.0.*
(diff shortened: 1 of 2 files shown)


samtygier-stfc:
There is a small screenshot change in the ring removal operation due to the spin box being narrower for the smaller range

rbauststfc (approved):
Changes look good. The screenshot tests are currently failing because another PR has added a new baseline image. I'll rebase and merge this in after the other PR has been merged.

PR #1628, opened 2022-11-02
1625: Add stack or replace stack in existing datasets

### LINKED ISSUE #1625: Allow adding extra stacks to a dataset
### Desired Behaviour

It should be possible to add extra stacks to a dataset that it doesn't already have.

DolicaAkelloEgwel:
### Issue

Closes #1625 

### Description

Allows adding a stack to a mixed/strict dataset. Stacks can also be replaced in the case of the strict dataset.

### Testing 

Have created tests for the new presenter + view. and tests for the changes made to other files.

### Acceptance Criteria 

Check that the tests pass. Check that when an image stack is added the tree view is updated and a new stack visualiser is created. Check that when a stack is replaced the new stack visualiser is created and the old one is destroyed, and that the entry on the tree view becomes linked to the new stack visualiser.

### Documentation

Updated release notes. Adding stuff about add / replace / delete data in the documentation.


samtygier-stfc (changes requested):
Just a small change on the assert in the tests. Otherwise looks good and works well.

samtygier-stfc (approved):

PR #1629, opened 2022-11-03
GUI work for sPDHG

samtygier-stfc:
### Issue

Closes #1508 

### Description
This adds the input fields for stochastic PDHG, ready for them to be used in `CILRecon`.

This also contains some changes to the tests. The individual commit names should explain the changes.

The "Stochastic" and "Subsets" field should show in the recon window only for CIL: PHDG-TV.

Subsets should be disabled unless Stochastic is checked.

Changing either should trigger a new preview, unless Auto Update is unchecked. (though the values have no effect on the recon yet).

The values should be printed to the terminal on preview. e.g.
`SPDHG params: recon_params.stochastic=True recon_params.subsets=2`

### Documentation
Will add release notes once the CIL side is implemented.

### DIFF
--- a/mantidimaging/core/reconstruct/cil_recon.py
+++ b/mantidimaging/core/reconstruct/cil_recon.py
@@ -203,4 +203,4 @@ class CILRecon(BaseRecon):
 
 
 def allowed_recon_kwargs() -> dict:
-    return {'CIL: PDHG-TV': ['alpha', 'num_iter', 'non_negative']}
+    return {'CIL: PDHG-TV': ['alpha', 'num_iter', 'non_negative', 'stochastic', 'subsets']}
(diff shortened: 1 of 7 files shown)


mantidimaging/core/reconstruct/cil_recon.py
```diff
@@ -69,6 +69,8 @@ def single_sino(sino: np.ndarray,
         Should return a numpy array,
         """
 
+        print(f"SPDHG params: {recon_params.stochastic=} {recon_params.subsets=}")
```
DolicaAkelloEgwel, line 72: It's a pretty minor thing but I wonder if it should show the subsets value when stochastic is false?

samtygier-stfc: This is just here so the CIL devs can see how to get the parameters. It will be removed from the final version.

DolicaAkelloEgwel (approved):

PR #1639, opened 2022-11-24
Dont pass loggers as arguments

### LINKED ISSUE #1638: Don't pass loggers around
### Summary
Some functions in core/io receive Logger objects as arguments. It looks like this might have originally been to decide if messages should be logged or not. This is better expressed with log levels.

It also makes log messages confusing, as they have the name of the module they were passed from. 

### Steps To Reproduce

Load some data by drag and drop or the --path argument

### Current Behaviour
In the terminal you will see some messages like
`[mantidimaging.gui.windows.main.presenter:L164] INFO: Could not find tif files in /data/imaging/IMAT00010675_128/Tomo/../Flat Before`

This message actually comes from `mantidimaging/core/io/utility.py` Line 164, not `mantidimaging/gui/windows/main/presenter.py`

### Expected Behaviour

Correct log messages

### Context

Current main

samtygier-stfc:
### Issue

Closes #1638

### Description

Remove all the places where loggers are passed into functions.

### Testing 

Can check that log message mentioned in the issue now shows with the correct module.

Automated testing should hit everything

### Documentation

Not needed

### DIFF
--- a/mantidimaging/core/io/loader/loader.py
+++ b/mantidimaging/core/io/loader/loader.py
@@ -237,7 +237,7 @@ def create_loading_parameters_for_file_path(file_path: str,
                                                         format=image_format)
 
     # 180 Degree projection
-    proj_180deg = find_180deg_proj(Path(sample_directory), image_format, logger)
+    proj_180deg = find_180deg_proj(Path(sample_directory), image_format)
     if proj_180deg != "":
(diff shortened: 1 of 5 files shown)


JackEAllen (approved):
These changes look good! Functionally the path printed when a file is not found now correctly tracks back to `[mantidimaging.core.io.utility:L162]` and automated test pass so happy to approve. 🥇

PR #1636, opened 2022-11-21
Use pyfakefs and cleanup log loading

samtygier-stfc:
### Issue
Progresses #1219

### Description

This improves some of the tests around file finding and loading by using pyfakefs. I have made the tests a bit more rigorous by having reducing mocking and having them act on the fake filesystem. This exposed some issues around how paths are handled on windows (auto finding log files probably never worked). Switching some path handling to use pathlib required a cascade of small changes through a lot of files.

### Testing & Acceptance Criteria 

Failures are expected on the docker tests as the docker images wont have pyfakefs in yet. Other tests should pass.

From the Load Dataset dialog, select the flower dataset sample. Check that the log files get correctly detected:
image
When loading check that there are no
'Could not find a log file for' messages in the terminal.

### Documentation

Not needed yet

PR #1689, opened 2023-02-06
Fix finding log in Load Dataset

### LINKED ISSUE #1649: NoneType error when loading phantom spheres dataset
### Summary

Trying to load the phantom spheres data as a dataset throws an error. Loading the flowers dataset still seems to work OK.

### Steps To Reproduce

1) Click File -> Load dataset
2) Attempt to load the phantom spheres test dataset

### Expected Behaviour

Load dialog is populated without error and the data can be loaded. If it's no longer supposed to be possible to load phantom spheres as a dataset due to the data structure then a helpful error message should be shown.

### Current Behaviour

An error message pops up with:

> The object passed as path for this field is not a string. Instead got <class 'NoneType'>

After clicking OK twice to dismiss the error, the OK button in the Load dialog is still disabled so it's not possible to load the dataset.

### Context

Current main

### Failure Logs

```
2022-12-16 14:37:20,983 [mantidimaging.gui.mvp_base.presenter:L24] ERROR: Presenter error: The object passed as path for this field is not a string. Instead got <class 'NoneType'>
  File "mantidimaging\mantidimaging\gui\windows\image_load_dialog\presenter.py", line 105, in do_update_sample
    self.set_sample_log(self.view.fields["Sample Log"], self.last_file_info.filenames)
  File "mantidimaging\mantidimaging\gui\windows\image_load_dialog\presenter.py", line 204, in set_sample_log
    sample_log.path = sample_log_filepath
  File "mantidimaging\mantidimaging\gui\windows\image_load_dialog\field.py", line 72, in path
    raise RuntimeError(f"The object passed as path for this field is not a string. Instead got {type(value)}")
RuntimeError: The object passed as path for this field is not a string. Instead got <class 'NoneType'>
  File "mantidimaging\mantidimaging\gui\windows\image_load_dialog\presenter.py", line 105, in do_update_sample
    self.set_sample_log(self.view.fields["Sample Log"], self.last_file_info.filenames)
  File "mantidimaging\mantidimaging\gui\windows\image_load_dialog\presenter.py", line 204, in set_sample_log
    sample_log.path = sample_log_filepath
  File "mantidimaging\mantidimaging\gui\windows\image_load_dialog\field.py", line 72, in path
    raise RuntimeError(f"The object passed as path for this field is not a string. Instead got {type(value)}")
RuntimeError: The object passed as path for this field is not a string. Instead got <class 'NoneType'>
2022-12-16 14:37:20,986 [mantidimaging.core.io.utility:L171] INFO: Could not find a log file for \phantom_spheres_float32\output\flat_0000.tiff
```

samtygier-stfc:
### Issue

Close #1649

### Description

Finish up some missing pathlib parts from #1636 . Also need checks in `do_update_sample()` because `find_log_for_image()` can return None.

Plus some extra type annotations that would have caught this issue.

Steps on #1649 should work without error

### Documentation

Not needed

### DIFF
--- a/mantidimaging/gui/windows/image_load_dialog/presenter.py
+++ b/mantidimaging/gui/windows/image_load_dialog/presenter.py
@@ -50,7 +50,7 @@ class LoadPresenter:
         self.last_file_info: Optional[FileInformation] = None
         self.dtype = '32'
 
-    def do_update_field(self, field: Field):
+    def do_update_field(self, field: Field) -> None:
         if field.file_info.mode == "sample":
             self.do_update_sample()
         elif field.file_info.mode == "images":
@@ -60,14 +60,14 @@ class LoadPresenter:
         elif field.file_info.mode in ["log", "180"]:
             self.do_update_single_file(field)
 
-    def do_update_sample(self):
+    def do_update_sample(self) -> None:
         """
         Updates the memory usage and the indices in the dialog.
         """
         selected_file = self.view.select_file("Sample")
         if not selected_file:
             self.view.ok_button.setEnabled(False)
-            return False
+            return
 
         self.view.sample.path = selected_file
         self.view.sample.widget.setExpanded(True)
@@ -110,8 +110,12 @@ class LoadPresenter:
         self.view.fields["Sample Log"].use = False
 
         for pos in ["Before", "After"]:
-            self.view.fields[f"Flat {pos} Log"].path = find_log_for_image(self.view.fields[f"Flat {pos}"].path_text())
-            self.view.fields[f"Flat {pos} Log"].use = False
+            flat_path = self.view.fields[f"Flat {pos}"].path_text()
+            if flat_path:
+                log_path = find_log_for_image(Path(flat_path))
+                if log_path:
+                    self.view.fields[f"Flat {pos} Log"].path = log_path
+                    self.view.fields[f"Flat {pos} Log"].use = False
 
         self.view.images_are_sinograms.setChecked(self.last_file_info.sinograms)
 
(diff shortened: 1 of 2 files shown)


mantidimaging/gui/windows/image_load_dialog/presenter.py
```diff
@@ -110,8 +110,12 @@ def do_update_sample(self):
         self.view.fields["Sample Log"].use = False
 
         for pos in ["Before", "After"]:
-            self.view.fields[f"Flat {pos} Log"].path = find_log_for_image(self.view.fields[f"Flat {pos}"].path_text())
```
JackEAllen, line 113: NITPICK: Given this is already a pretty long method and we are touching it anyway, what are you're thoughts on breaking this out into a separate method to be called called something like: `flat_path_logs` and break into a separate test in `presenter_test.py`?

JackEAllen (changes requested):
Looks pretty good! Thank you for adding all the type checks Sam!

From manual testing, comparing between `main` and `1649-load-nonetype` follwing instructions left on #1689 loading the phantom spheres dataset no longer produces previously experienced errors relating to unexpected path types.

I have left a minor nitick I wanted to get your opinion on to seize an opportunity for a small refactor to shrink the size of `do_update_sample()` before approving. Once I have your opinion on this, happy to merge 👍

mantidimaging/gui/windows/image_load_dialog/presenter.py
samtygier-stfc: I'm a bit torn on this. I want to refactor, but I'll do it in future PRs

JackEAllen: That's fine by me! It's why I wanted to get your opinion on this first. I'll approve this and merge in. We should create and issue to refactor and link to this PR for reference.

JackEAllen (approved):
LGTM! 👍

PR #1699, opened 2023-02-14
Drag and drop loading dialog and cleanup

samtygier-stfc:
### Issue
Cleanups for #1219 

### Description

When loading through drag and drop go though the Load Dataset dialog. This is useful as it allows the user to check settings, and avoids using a different code path.

I was hope to remove the need for `create_loading_parameters_for_file_path()` but it is still needed when using `--path` on the CLI. So not as big a cleanup as hoped.

This updates a system test to check that loading a dataset from the dialog actually works.

Moving calls to `select_file()` into `do_update_field` simplifies the `do_update...()` and reduces the amount of mocking needed in tests.

(I recommend view changes a commit at a time).

Check that a dataset can be loaded by drag and drop

### Documentation

Not needed

### DIFF
--- a/mantidimaging/gui/windows/image_load_dialog/presenter.py
+++ b/mantidimaging/gui/windows/image_load_dialog/presenter.py
@@ -51,20 +51,23 @@ class LoadPresenter:
         self.dtype = '32'
 
     def do_update_field(self, field: Field) -> None:
-        if field.file_info.mode == "sample":
-            self.do_update_sample()
-        elif field.file_info.mode == "images":
-            self.do_update_flat_or_dark(field)
-        elif field.file_info.name == "Sample Log":
-            self.do_update_sample_log(field)
-        elif field.file_info.mode in ["log", "180"]:
-            self.do_update_single_file(field)
+        name = field.file_info.name
+        is_image_file = field.file_info.mode != "log"
+        selected_file = self.view.select_file(name, is_image_file)
 
-    def do_update_sample(self) -> None:
+        if field.file_info.mode == "sample":
+            self.do_update_sample(selected_file)
+        elif field.file_info.mode == "images":
+            self.do_update_flat_or_dark(field, selected_file)
+        elif field.file_info.name == "Sample Log":
+            self.do_update_sample_log(field, selected_file)
+        elif field.file_info.mode in ["log", "180"]:
+            self.do_update_single_file(field, selected_file)
+
+    def do_update_sample(self, selected_file: Optional[str]) -> None:
         """
         Updates the memory usage and the indices in the dialog.
         """
-        selected_file = self.view.select_file("Sample")
         if not selected_file:
             self.view.ok_button.setEnabled(False)
             return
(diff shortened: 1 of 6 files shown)


mantidimaging/gui/windows/image_load_dialog/presenter.py
```diff
@@ -51,20 +51,23 @@ def __init__(self, view: 'ImageLoadDialog'):
         self.dtype = '32'
 
     def do_update_field(self, field: Field) -> None:
+        name = field.file_info.name
+        is_image_file = field.file_info.mode != "log"
+        selected_file = self.view.select_file(name, is_image_file)
```
JackEAllen, line 56: NITPICK: We could add type checking here if you feel it would be helpful

mantidimaging/gui/windows/image_load_dialog/test/presenter_test.py
```diff
@@ -23,13 +23,23 @@ def setUp(self):
         self.p = LoadPresenter(self.v)
 
     def test_do_update_sample_with_no_selected_file(self):
-        self.v.select_file.return_value = None
+        self.p.do_update_sample(None)
```
JackEAllen, line 26: NITPICK: though it is bordering being outside the scope of the work done within this PR, I feel that `self.v` and `self.p` are not particularly descriptive names. How would you feel about updating them as part of your changes or would you prefer we do this as a separate flash review if you agree?

mantidimaging/gui/windows/image_load_dialog/test/presenter_test.py
```diff
@@ -23,13 +23,23 @@ def setUp(self):
...
-
-        self.v.select_file.assert_called_once_with("Sample")
         self.v.ok_button.setEnabled.assert_called_once_with(False)
 
+    @mock.patch("mantidimaging.gui.windows.image_load_dialog.presenter.LoadPresenter.do_update_flat_or_dark")
+    def test_do_update_field(self, mock_do_update_flat_or_dark):
+        selected_file = "SelectedFile"
+        self.v.select_file.return_value = selected_file
+        name = "Name"
+        suffix = "Apples"
```
JackEAllen, line 35: NITPICK: As there is only one suffix, should we rename this from Apples 🍎 🍏 to Apple 🍎 ? 😉

mantidimaging/gui/windows/main/view.py
```diff
@@ -226,6 +227,14 @@ def show_image_load_dialog(self):
         self.image_load_dialog = ImageLoadDialog(self)
         self.image_load_dialog.show()
 
+    def show_image_load_dialog_with_path(self, file_path: str) -> bool:
```
JackEAllen, line 230: Can we add a docstring here?

JackEAllen (changes requested):
Looks good overall from a first pass through! 👍  
Functionally behaves as expected and it's nice to have to option to play around with file selection when dragging and dropping a dataset into mantid imaging.
Once I have your opinion on a few nitpicks I have added and I give this another once over, happy to approve and merge this in.

mantidimaging/gui/windows/image_load_dialog/test/presenter_test.py
samtygier-stfc: Its a shorthand we use in lots of tests, so would be a big change if we did it. Maybe a cleanup to consider in the future.

mantidimaging/gui/windows/image_load_dialog/test/presenter_test.py
samtygier-stfc: These go away in the next PR.

JackEAllen: I'll miss the Apple suffix 😢

JackEAllen (approved):
Thanks for adding in the docstring. I've given your changes another pass through and happy for this to be merged in.

JackEAllen:
> That looks like a timeout in githubs queuing system maybe. Do you want to try queuing it again.

Yeah, maybe... - I have retried it so lets see if it works this time round. One thing i'm not keen on is how long it takes to move from the queue to actually running the build checks even if it's the only PR in the queue.

Issue #1641, opened 2022-11-25
Timeouts in windows GUI system tests


samtygier-stfc:
### Summary

These tests are timing out often enough to be annoying.

Successes typically take between 7 and 12 minutes. Timeout is set to 30. (If this can't be solved then the timeout should be reduced to 15 mins, to fail faster).

![image](https://user-images.githubusercontent.com/74248560/203983121-a0adf581-079e-4065-8779-9b4164976022.png)

This got left over from the last round of timeout testing as it was less common than the other issues.

A possible cause is windows being left open in some of the tests.

### Steps To Reproduce
Seen on several recent actions runs
https://github.com/mantidproject/mantidimaging/actions/workflows/windows.yml?query=is%3Afailure

### Expected Behaviour

Tests should pass

### Current Behaviour
Failures

### Context
All recent PRs

samtygier-stfc:
Its possible to get timeouts on actions running just gui_system_reconstruction_test.py https://github.com/mantidproject/mantidimaging/actions/runs/3565159791/jobs/5989970716

samtygier-stfc:
Running with just a single test from gui_system_reconstruction_test.py, I don't see any timeouts. https://github.com/mantidproject/mantidimaging/actions/runs/3566484458

samtygier-stfc:
With exclusions. I do see timeouts running gui_system_reconstruction_test.py, with a test excluded.
test_correlate
test_selecting_auto_update_triggers_preview
test_clicking_update_now_btn_triggers_preview

samtygier-stfc:
With latest test PR the minimal trigger seems to be running just 
`python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov --run-system-tests --durations=10 mantidimaging/gui/test/gui_system_reconstruction_test.py -k test_refine_stress`

When this succeeds it takes ~ 12 seconds.

In the single test case we see:
`D:\a\_temp\196debb1-9287-4d0d-be42-b42ba56806e3.sh: line 1:  1461 Segmentation fault      python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov --run-system-tests --durations=10 mantidimaging/gui/test/gui_system_reconstruction_test.py -k test_refine_stress`

Instead of a timeout.

samtygier-stfc:
With `python -X faulthandler` to only extra output is.
```
 Windows fatal exception: access violation

[...]

PR #1709, opened 2023-02-20
Prevent timeouts in gui system tests

samtygier-stfc:
### Issue

Partly solves #1641

### Description

This hopefully turns all current and many possible hangs into visible error messages.

Enables live logging, so that log messages a visible straight away.

Unifies calls to show `QMessageBox` for errors/warnings, and makes sure that a log message is sent first.
(Note this drops the custom title used in a few places. Not sure it was really needed).

Mock out `show_error_dialog` in `TestGuiSystemReconstruction` to prevent the test getting stuck at a message box.

Tests might still fail, but they should fail fast (no timeouts) and should have some error message

### Documentation
release notes


JackEAllen (dismissed):
Looks good! 
I think we should probably place some more formal documentation somewhere about best practices maybe under a GitHub Wiki or modify the pre-commit to check for references to `QMessageBox.warning` over `self.show_warning_dialog` to make sure we use `self.show_warning_dialog` going forward and that `QMessageBox.warning` isn't missed/flagged up during review

JackEAllen (approved):

JackEAllen (approved):

PR #1726, opened 2023-03-01
Use FileNameGroups for load dataset

### LINKED ISSUE #1219: When loading files like NAME_001.tif NAME_002.tif don't include NAME_180deg.tif

samtygier-stfc:
### Issue

Nearly there on #1219 

### Description

Move Load Dataset to new code. This covers the loading side, finding files in ImageLoadDialog still uses old methods.

Remove unused code

Check that you can load  with Load Dataset

### Documentation

Not yet

### DIFF
--- a/mantidimaging/core/io/test/filenames_test.py
+++ b/mantidimaging/core/io/test/filenames_test.py
@@ -57,6 +57,13 @@ class FilenamePatternTest(unittest.TestCase):
         self.assertTrue(p1.match("img_1000.tif"))
         self.assertFalse(p1.match("img_0000.tif"))
 
+    def test_pattern_match_different_digits_no_zeros(self):
+        # Allow cases where index has grown above padding
+        p1 = FilenamePattern.from_name("img_1.tif")
+        self.assertTrue(p1.match("img_10.tif"))
+        self.assertTrue(p1.match("img_1234.tif"))
+        self.assertFalse(p1.match("img_0000.tif"))
+
     def test_pattern_get_index(self):
         p1 = FilenamePattern.from_name("IMAT_Flower_Tomo_000007.tif")
         self.assertEqual(p1.get_index("IMAT_Flower_Tomo_000023.tif"), 23)
--- a/mantidimaging/gui/windows/image_load_dialog/test/presenter_test.py
+++ b/mantidimaging/gui/windows/image_load_dialog/test/presenter_test.py
@@ -233,77 +233,68 @@ class ImageLoadDialogPresenterTest(unittest.TestCase):
         mock_load_log.assert_not_called()
         mock_log.raise_if_angle_missing.assert_not_called()
 
-    @mock.patch("mantidimaging.gui.windows.image_load_dialog.presenter.get_prefix", return_value="/path")
-    def test_get_parameters(self, get_prefix):
-        path_text = "/path/text"
-        sample_input_path = "/sample/input/path"
-        image_format = ".tif"
-        sample_indices = [1, 1]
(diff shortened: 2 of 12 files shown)


JackEAllen (changes requested):
From an initial pass through, looks really good and changes work as expected! Thank you for the changes! Definitely on the larger side of what I'm happy to consider as one PR. If it was much larger I would suggest breaking into multiple PRs. I have left a few nitpicks here and there that I would like to get your opinion on. I'll give this another pass through to make sure I'm not missing anything before approving.

mantidimaging/core/io/test/filenames_test.py
```diff
@@ -57,6 +57,13 @@ def test_pattern_match_different_digits(self):
         self.assertTrue(p1.match("img_1000.tif"))
         self.assertFalse(p1.match("img_0000.tif"))
 
+    def test_pattern_match_different_digits_no_zeros(self):
```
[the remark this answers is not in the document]
samtygier-stfc: I think its simpler to do it in a single test here, as there is not lots of duplicated setup.

mantidimaging/gui/windows/image_load_dialog/test/presenter_test.py
```diff
@@ -233,77 +233,68 @@ def test_ensure_sample_log_consistency_exits_when_none_or_empty_str(self, mock_l
...
+        sample_path = "/sample/tomo/tomo_0001.tiff"
+        sample_file = "tomo_0001.tiff"
+        sample_log_path = "/sample/tomo.log"
+
+        self.fields["Sample"].path_text.return_value = sample_path
+        self.fields["Sample"].file.return_value = sample_file
+        self.fields["Sample"].use.isChecked.return_value = True
+        self.fields["Sample Log"].path_text.return_value = sample_log_path
+
+        # sample_indices = [1, 1] #TODO
```
JackEAllen, line 247: Did you mean to leave this in?

samtygier-stfc: It gets fixed in the next PR #1728 ( ba2b7bb6d83889c53a268534796fc9aef215a19e )

JackEAllen (approved):
Thanks for making the small changes to add docstrings. Though they may be simple descriptions, they are enough to provide context for future usage should anyone else ever need to touch this. Happy to approve this 👍

PR #1737, opened 2023-03-17
Remove grad norm calculation

### LINKED ISSUE #1724: SPDHG Remove grad norm calc
### Desired Behaviour

#1647 added calculation for the norm of the Grad and ProjectionOperators. The Grad part is now done upstream in CIL, so can be removed from mantid imaging.

https://github.com/TomographicImaging/CIL/commit/5ebebbacea08491fe731e73858d9cc63512ced5b

(The calc for ProjectionOperator is still needed as that can't be moved into CIL)

samtygier-stfc:
### Issue

Closes #1724

### Description
Remove no longer needed grad norm calculation

### Acceptance Criteria 

Check we are no longer using set_norm for the Grad operator.

Manually check that reconstruction with CIL still works.

### Documentation

Not needed.

### DIFF
--- a/mantidimaging/core/reconstruct/cil_recon.py
+++ b/mantidimaging/core/reconstruct/cil_recon.py
@@ -57,11 +57,6 @@ class CILRecon(BaseRecon):
         alpha = recon_params.alpha
         Grad = GradientOperator(image_geometry)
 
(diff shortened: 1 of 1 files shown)


JackEAllen (approved):
LGTM! 👍

PR #1764, opened 2023-04-18
NSIS installer for windows

### LINKED ISSUE #1762: Windows exe not starting ImportError
We have several reports of the windows exe from https://github.com/mantidproject/mantidimaging/releases/tag/2.5.0a2.post262-20230310 not starting on some windows system.

image

One user has 2 machines with Microsoft Windows 10 Enterprise Version 10.0.19044 Build 19044, one works and one does not. Possibly they have different GPU hardware.

samtygier-stfc:
### Issue

Closes #1762

### Description

The onefile mode of pyInstaller seems to be vulrable to picking up the wrong DLL for astra. The onedir is more robust because the DLLs sit next to the executable (without the unpacking to temp trick). But an installer is needed to avoid the user having to manually unpack and find the correct executable.

An installer gives several other advantages, menu entries, uninstall support, control over compression.

### Acceptance Criteria 

To build the installer you need makensis from `mamba install nsis`.
In the packaging directory run
`python PackageWithPyInstaller.py`
`makensis nsis_settings.nsi`
this should produce a setup file in dist.

Try installing from this file. Check that Mantid Imaging is added to the start menu. Check that it launches and runs. Check that it uninstalls.


samtygier-stfc:
https://github.com/mantidproject/mantidimaging/releases/tag/2.5.0a4 has been produced with this branch, and tested on a number of systems.

packaging/nsis_settings.nsi
JackEAllen, line 1: NITPICK: Can we add our copyright statement to this?

JackEAllen (approved):
Changes look good and have tested on Windows 11 without CUDA and changes seem to resolve issue with correct DLL pickup so happy to approve. Only nitpick it to add our copyright statement to the `.nsi` file.

packaging/nsis_settings.nsi
samtygier-stfc: I don't think I'll have time today to fire up a vm to test a change, so lets leave as is.

packaging/nsis_settings.nsi
```diff
@@ -0,0 +1,47 @@
...
+	WriteRegStr SHCTX "Software\Microsoft\Windows\CurrentVersion\Uninstall\${PRODUCT}" "UninstallString" '"$INSTDIR\uninstall.exe"'
+	WriteRegDWORD SHCTX "Software\Microsoft\Windows\CurrentVersion\Uninstall\${PRODUCT}" "NoModify" 1
+	WriteRegDWORD SHCTX "Software\Microsoft\Windows\CurrentVersion\Uninstall\${PRODUCT}" "NoRepair" 1
+SectionEnd
+
+Section "Uninstall"
+	SetShellVarContext all
+	Delete "$SMPROGRAMS\${PRODUCT}\${PRODUCT}.lnk"
+	Delete $INSTDIR\uninstall.exe
+	RMDir /r "$INSTDIR"
```
thomashampson, line 43: `$INSTDIR` here is just the directory from which you run uninstall.exe, so it's not very safe to do this, e.g. someone may copy the uninstaller somewhere else before running it (this has actually happened to me before, when a user moved an uninstaller to their desktop before running it and it deleted their desktop folder).
See the warning here:
https://nsis.sourceforge.io/Reference/RMDir

samtygier-stfc: Filed as #1772

PR #1767, opened 2023-04-19
RebinFilter: correct docstring

### LINKED ISSUE #1406: RebinFilter modes in docs does not match allowed modes
### Summary

The docs say

```
        :param mode: Interpolation to use for re-sizing
                     ('nearest', 'lanczos', 'bilinear', 'bicubic' or 'cubic').
```
But the allowed modes are
```
def modes():
    return ["constant", "edge", "wrap", "reflect", "symmetric"]
```

Might need to add options to choose between edge mode and interpolation mode.

### Context
Current main

samtygier-stfc:
### Issue
Closes #1406

### Description
Fix docsting to describe what mode does.

### Acceptance Criteria 

Check docs locally with
`python ./setup.py internal_docs`

### DIFF
--- a/mantidimaging/core/operations/rebin/rebin.py
+++ b/mantidimaging/core/operations/rebin/rebin.py
@@ -39,8 +39,8 @@ class RebinFilter(BaseFilter):
                             int - Percentage of current size.
                             float - Fraction of current size.
                             tuple - Size of the output image (x, y).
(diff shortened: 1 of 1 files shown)


JackEAllen (approved):
Changes look good and docs correctly updated when built

PR #1769, opened 2023-04-20
Hide float64 option when loading images

### LINKED ISSUE #1761: Hide float64 option in load dataset
Based on discussion with IMAT scientists remove the float64 option when loading a dataset.

If possible this should be done in a way that would be easy to add back in. Possibly just hiding the GUI widgets.

samtygier-stfc:
### Issue

Closes #1761

### Description

Some parts of Mantid Imaging don't handle float64. There is not much benefit to the extra precision, only a large memory code. So hide the option, and probably remove it completely in the future.

### Acceptance Criteria 

Load Dataset and Load Nexus should no longer show the bit depth option.

### Documentation
release notes

### DIFF
--- a/docs/release_notes/next.rst
+++ b/docs/release_notes/next.rst
@@ -51,6 +51,7 @@ Fixes
 - #1721 : For SPDHG show projections per subset
 - #1720 : Update memory requirements for SPDHG
 - #1742 : Restore native dialogs on Linux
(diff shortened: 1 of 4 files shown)


JackEAllen (approved):
Happy with these changes having discussed alternatives outside this thread so will approve. 👍

PR #1797, opened 2023-05-17
MyPy Pre-commit

### LINKED ISSUE #1796: Add MyPy Checks to Pre-Commit
### Desired Behaviour

Add MyPy checks to pre-commit hook so that we can catch issues before pushing to remote.

### Current Behaviour

MyPy checks do not run as part of pre-commit hook currently. MyPy is only run once in PR by Git Workflows.

JackEAllen:
### Issue

Closes #1796

### Description

Add MyPy Checks as Pre-Commit checks as extra validation that developers are producing MyPy compliant code and reducing the risk of type errors being introduced into the project.

### Testing 

- [ ] Checkout branch this branch locally and create some compliant MyPy code and try to commit it. Pre-commit should block the commit.

### Where
* `.pre-commit-config.yaml`

### Acceptance Criteria 

- [ ] Only MyPy compliant code can be committed locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,3 +21,12 @@ repos:
         entry: python -m CheckLicensesInFiles
         language: python
         types: [python]
+-   repo: local
+    hooks:
+    -   id: MyPy Checks
+        name: MyPy Checks
+        description: Checks that all files have correct type annotations
+        entry: p  mypy --ignore-missing-imports --no-site-packages mantidimaging
+        language: python
+        types: [python]
+
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,16 +21,9 @@ repos:
         entry: python -m CheckLicensesInFiles
         language: python
         types: [python]
--   repo: local
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v0.910'
     hooks:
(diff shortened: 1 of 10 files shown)


.pre-commit-config.yaml
```diff
@@ -21,3 +21,12 @@ repos:
         entry: python -m CheckLicensesInFiles
         language: python
         types: [python]
+-   repo: local
+    hooks:
+    -   id: MyPy Checks
+        name: MyPy Checks
+        description: Checks that all files have correct type annotations
+        entry: p  mypy --ignore-missing-imports --no-site-packages mantidimaging
```
samtygier-stfc, line 29: I think this should be
`entry: mypy --ignore-missing-imports --no-site-packages`

The `p` is giving me command not found

And the path is not needed as the changed files get passed in by pre-commit. (otherwise I get `error: Duplicate module` when I have a changed file).

samtygier-stfc (changes requested):
I had to tweak the command to get it working here

.pre-commit-config.yaml
JackEAllen: Should be fixed now, I hadn't realised I introduced a typo just before committing. 😅  Should by python instead of p now. I've also updated the path.  p worked for me as I have it aliased to python as i don't trust my ability to type python all the time

PR #1809, opened 2023-05-25
Update to python 3.10

### LINKED ISSUE #1792: Update python version
### Desired Behaviour

Update to a newer python version.

Python 3.10 gives us Pattern matching, some typing improvements, improved error messages. Also performance improvements

3.10 is the newest version that CIL is packaged for.

samtygier-stfc:
### Issue

Closes #1792 

### Description
Updates python to 3.10, and also numpy to 1.21 (needed for the python update)
This is the most up to date supported by CIL

Also fixes a typo in the tests, that new python picked up. I have switched here to using autospec, to reduce the risk of typos here.

### Testing 

Existing tests should all still pass

### Acceptance Criteria 

Create a new dev environment 
`python ./setup.py create_dev_env`

Check that MI still starts

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -15,12 +15,12 @@ requirements:
     - python=3.9.*
     - setuptools=62.*
   run:
-    - python=3.9.*
+    - python=3.10.*
     - pip
     - astropy=5.0.*
     - scipy=1.7.*
     - scikit-image=0.19.*
-    - numpy=1.20.*
+    - numpy=1.21.*
     - numexpr=2.8.*
     - algotom=1.0.*
     - tomopy=1.12.*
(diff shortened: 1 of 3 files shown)


JackEAllen (approved):
Happy to approve this as changes look good, tests pass and Mantid Imaging runs as expected. I did notice that I needed to install Mamba inside the dev env as I experienced some visibility issues with `TestCheckVersion.test_conda_update_message` not finding conda, but finding mamba only on installing from within the dev env. falling over here:
```python 
    @staticmethod
    def find_conda_executable() -> str:
        if shutil.which("mamba"):
            return "mamba"
        if shutil.which("conda"):
            return "conda"
>       raise FileNotFoundError
E       FileNotFoundError
```
This doesn't seem to be related to this PR though as I can replicate on main.

PR #1817, opened 2023-05-31
Move release notes to one change per file

### LINKED ISSUE #1803: File per change release notes
### Desired Behaviour

The release notes file is a common source on merge conflicts, because most PRs touch it. Git will always raise a conflict if two commits append a line to the same point in file.

We should move to each release note entry being its own file, and the merge them at release time.
Mantid did this in https://github.com/mantidproject/mantid/issues/33593 though a simpler solution may work for us.

Worth having a quick look at the solutions used by mantid and numpy ( https://numpy.org/doc/stable/dev/releasing.html ). But it might be easiest to do them in a way where they can just be `cat`'ed into the final file.

samtygier-stfc:
### Issue
Closes #1803

### Description
Use individual files for release note changes.

A sphinx extension to combine them at doc build time.
A setup.py rule to join them for release

`python3 ./setup.py internal_docs`
and check that the release note page is built correctly

`python3 ./setup.py release_notes`
And check that the release notes are shown correctly

### Documentation

Updated the release notes and documentation

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -52,6 +52,7 @@ extensions = [
 # Add custom extensions
 sys.path.append(os.path.abspath("./ext"))
 extensions.append("operations_user_doc")
+extensions.append("release_notes")
 
 # Add any paths that contain templates here, relative to this directory.
 templates_path = ['_templates']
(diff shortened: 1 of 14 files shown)


JackEAllen (approved):
Changes look good and generating docs and release notes works well!

PR #1826, opened 2023-06-02
Skip mantidimaging in create_dev_env even when it has as version

### LINKED ISSUE #1822: mantidimaging package being installed in dev enviroment
### Summary

`python3 ./setup.py create_dev_env` should exclude the mantidimaging package as it adds dependencies from the meta.yaml file to a temporary environment file.

f85799909107dfbdd2b47d3249b0f5000957deea broke the check in `setup.py` to ignore that line

`if line.strip() == "- mantidimaging":`

this needs adjusting to catch `mantidimaging>=2.5.0`

### Steps To Reproduce
`python3 ./setup.py create_dev_env`
notice that mantidimaging gets installed.

You can check if mantid imaging is in an environment by running
`mamba list mantidimaging`

### Expected Behaviour

mantidimaging not installed in dev env.

### Context
Current main. Since #1799

samtygier-stfc:
### Issue

Closes #1822

### Description

Skip mantidimaging in create_dev_env even when it has as version

Check that mantidimaging is no longer being installed in the dev environment (see bug report)

### Documentation

Not needed


JackEAllen (approved):
Looks good! Can't see mantidimaging in a new dev env

PR #1836, opened 2023-06-07
1833 rotate stack dropdown

### LINKED ISSUE #1805: Mantid Imaging - Rotating after cropping causes loss of data
### Summary

**Raised by user Hamish_C** https://forum.mantidproject.org/t/bug-mantid-imaging-rotating-after-cropping-causes-loss-of-data/857

Had a quick search of issues, doesn't look like this is a duplicate.

### Steps To Reproduce

"Open an image in Mantid Imaging. Go to Workflow → Operations. Perform a crop to some non-square aspect ratio. Perform a 90 degree rotation."

### Expected Behaviour

"When performing operations on a stack, if you choose to crop the stack to a non-square aspect ratio and then rotate the stack, you would expect that the “aspect ratio” also “rotates”, e.g. if you have a stack that has been cropped to have a width of 2 and a height of 1, and you rotate it by 90 degrees, you would expect the width to become 1 and the height to become 2, and all of the original data to still be represented."

### Current Behaviour

"The aspect ratio of the final image (after rotation) is kept identical to before rotation, combined with a loss of data and addition of black bars to fill the space."

### Context

"I am using this on IDAaaS at the moment, I believe the same happens in either the stable or unstable builds that are present on IDAaaS right now."

### Screenshot(s)

image

JackEAllen:
### Issue

Closes #1805 

### Description

Create Dropdown of default and custom angles of rotation within Rotate Stacks Operation window.
Angles: 0, 90, 180, 270, custom
Selecting custom will enable the custom angle spinbox so a user can select a more precise angle.

### Acceptance Criteria 

- [ ] A dropdown exists within the Operations window Rotate Stack selection containing the values: 0, 90, 180, 270, custom
- [ ] If any value within the dropdown is selected other than "custom" then the custom angle text entry spinbox is disabled.
- [ ] If a default value is selected from the dropdown, the spinbox value below is updated to show this angle

### Documentation

See: `docs/release_notes/next/fix-1833-rotate_stack_dropdown` : [972259e](https://github.com/mantidproject/mantidimaging/pull/1836/commits/972259ebcc54cc38f525fbf18571581354c98f40)

### DIFF
--- a/docs/release_notes/next/fix-1833-rotate_stack_dropdown
+++ b/docs/release_notes/next/fix-1833-rotate_stack_dropdown
@@ -1,0 +1,1 @@
+#1833 : Create dropdown menu of default degrees of rotation within operations rotate stack 
--- a/mantidimaging/core/operations/rotate_stack/rotate_stack.py
+++ b/mantidimaging/core/operations/rotate_stack/rotate_stack.py
@@ -6,6 +6,7 @@ from functools import partial
 from typing import TYPE_CHECKING
 
 from skimage.transform import rotate
+from PyQt5.QtWidgets import QComboBox
 
 from mantidimaging import helper as h
 from mantidimaging.core.operations.base_filter import BaseFilter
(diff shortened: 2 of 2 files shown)


samtygier-stfc (approved):
Looks good

PR #1852, opened 2023-07-03
1805 rotate stack aspect ratio

### LINKED ISSUE #1805: Mantid Imaging - Rotating after cropping causes loss of data

JackEAllen:
### Issue

Closes #1805 

### Description

- Allow for rotating of image in-place and not in-place.
- If image is rotated using default dropdown cardinal angles, image and image aspect ration is rotated to corresponding cardinal angle.
- If image is rotated to a custom general angle, the aspect ratio is snapped to the closest cardinal angle rounding down i.e. 45 would snap to 0 rotation and 46 would snap to 90.
- Allow to rotating in clockwise and counter-clockwise directions.

### Testing 

Added basic unit test to check that aspect ratio is correctly reshaped if out or place rotations and stays the same for in_place rotations, testing that for a given general angle, the correct cardinal angle is returned.

### Acceptance Criteria 

- [ ] Can rotate using dropdown menu to a default cardinal angle
- [ ] If rotated to a default cardinal angle, then select custom, angle starts from angle last set.
- [ ] Can rotate custom angle
- [ ] If image angle is <=45, aspect ration is 0. if image angle is 46, aspect ration is rotated by 90
- [ ] If image angle is <=134, aspect ration is 90, if image angle is >=136, aspect ration is rotated by 180
- [ ] If image angle is <=225, aspect ration is 180, if image angle is >=226, aspect ration is rotated by 270
- [ ] If image angle is <=315, aspect ration is 270, if image angle is >=316, aspect ration is rotated by 0
- [ ] If rotating image into negative values, image rotates clockwise following similar behaviour if rotated using positive values.

### Documentation

### DIFF
--- a/mantidimaging/core/operations/rotate_stack/rotate_stack.py
+++ b/mantidimaging/core/operations/rotate_stack/rotate_stack.py
@@ -10,6 +10,7 @@ from PyQt5.QtWidgets import QComboBox
 from mantidimaging import helper as h
 from mantidimaging.core.operations.base_filter import BaseFilter
 from mantidimaging.core.parallel import shared as ps
+from mantidimaging.core.utility.progress_reporting import Progress
 from mantidimaging.core.parallel import utility as pu
 from mantidimaging.gui.utility.qt_helpers import Type
 
--- a/mantidimaging/core/operations/rotate_stack/rotate_stack.py
+++ b/mantidimaging/core/operations/rotate_stack/rotate_stack.py
@@ -4,15 +4,16 @@ from __future__ import annotations
 
 from functools import partial
 from typing import TYPE_CHECKING
-
 from skimage.transform import rotate
 from PyQt5.QtWidgets import QComboBox
 
 from mantidimaging import helper as h
 from mantidimaging.core.operations.base_filter import BaseFilter
 from mantidimaging.core.parallel import shared as ps
-from mantidimaging.core.utility.progress_reporting import Progress
+from mantidimaging.core.parallel import utility as pu
 from mantidimaging.gui.utility.qt_helpers import Type
+
+import numpy as np
 
 if TYPE_CHECKING:
     from mantidimaging.core.data import ImageStack
@@ -37,20 +38,16 @@ class RotateFilter(BaseFilter):
         """
         Rotates images by an arbitrary degree.
 
-        :param data: stack of sample images
-        :param angle: The rotation to be performed, in degrees
+        param data: stack of sample images
+        param angle: The rotation to be performed, in degrees
 
-        :return: The rotated images
+        return: The rotated images
         """
         h.check_data_stack(data)
-
         if angle is None:
             raise ValueError('Value must be provided for angle parameter')
-
-        # No need to run the filter for an angle of 0 as it won't have any effect
-        if not angle == 0:
-            _execute(data, angle, progress)
-
+        progress = Progress.ensure_instance(progress, task_name='Rotate Stack')
+        _do_rotation(data, round(angle, 3), progress)
         return data
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


mantidimaging/core/operations/rotate_stack/rotate_stack.py
```diff
@@ -96,15 +95,94 @@ def _update_angle(angle, dropdown):
...
+    param data: current image data array
+    param angle: angle of rotation
+    return: new_data: rotated image data array
+    """
+    angle += 360 if angle < 0 else 0
+    if angle == _get_cardinal_angle(angle):
+        data.shared_array = _cardinal_rotation_per_slice(data, angle, progress)
+    else:
+        data.shared_array = _cardinal_rotation_per_slice(data, _get_cardinal_angle(angle), progress)
+        data.shared_array = _inplace_rotation(data.data.shape[0], data, angle, _get_cardinal_angle(angle), progress)
```
samtygier-stfc, line 127: Could we do the subtraction here instead of in the function, i.e. pass `angle - _get_cardinal_angle(angle)`.

mantidimaging/core/operations/rotate_stack/rotate_stack.py
```diff
@@ -96,15 +95,94 @@ def _update_angle(angle, dropdown):
...
+    """
+    Handle rotation of image slice by angle. This includes rotation of aspect
+    ratio and in place rotation depending on angle.
+
+    param array_index: index of array in array_list
+    param data: current image data array
+    param angle: angle of rotation
+    return: new_data: rotated image data array
+    """
+    angle += 360 if angle < 0 else 0
```
samtygier-stfc, line 122: `angle = angle % 360` should do the same and work for numbers outside -360 to 360.

mantidimaging/core/operations/rotate_stack/rotate_stack.py
```diff
@@ -96,15 +95,94 @@ def _update_angle(angle, dropdown):
...
+    param: array_index_z: index of array in array_list
+    param: data: current image data array
+    param: angle: angle of rotation
+    param: angle_offset: angle offset
+    param: progress: progress bar
+    return: new_data: rotated image data array
+    """
+    new_data = pu.create_array(data.data.shape, data.data.dtype)
+    ps.run_compute_func(_compute_rotation_per_slice_inplace, array_index_z, [data.shared_array, new_data],
+                        {"angle": angle - angle_offset}, progress)
```
samtygier-stfc, line 166: We should not need to create a new array here. Just need to pass in the existing data, and have the compute function write into the same function.

mantidimaging/core/operations/rotate_stack/rotate_stack.py
```diff
@@ -96,15 +95,94 @@ def _update_angle(angle, dropdown):
...
+    rotated by.
+
+    param: value: angle of rotation
+    return: angle: cardinal angle of rotation
+    """
+    angle = ((value % 360 + 45) // 90 * 90 - (90 if (value % 360) % 90 == 45 else 0))
+    return angle % 360
+
+
+def _compute_rotation(data, angle, progress) -> np.ndarray:
```
samtygier-stfc, line 112: I'd rather we preserved the name `compute` for the functions passed to `run_compute_func`.

It could either be `do_rotation`, or content could be moved into `filter_func`

samtygier-stfc (commented):
Looking good, but a few comments.

mantidimaging/core/operations/rotate_stack/rotate_stack.py
```diff
@@ -96,15 +95,94 @@ def _update_angle(angle, dropdown):
...
+    """
+    Calculates the angle based on the input value, by adding 45 to the value modulo 360,
+    dividing it by 90, multiplying by 90, and then subtracting 90 if the value modulo 360 modulo
+    90 is equal to 45, otherwise subtracting 0 to get the cardinal angle the aspect ratio should be
+    rotated by.
+
+    param: value: angle of rotation
+    return: angle: cardinal angle of rotation
+    """
+    angle = ((value % 360 + 45) // 90 * 90 - (90 if (value % 360) % 90 == 45 else 0))
```
samtygier-stfc, line 108: If we have already normalised the range to 0--360 in `filter_func` would `round(x/90)*90` do the same?

JackEAllen: I think i'm going to stay with the existing solution such that cardinal angle rounds down if exactly at the half way point like discussed originally so if value is 45deg, 0 is returned, but if 45.001, 90 is returned. If using `round(<45> / 90) * 90`, I would get 90 instead of 0.

samtygier-stfc: For me `round(45/90)` is giving `0`. The docs say it uses round-half-even, so it should prefer 0 and 180 when on the boundary. That means avoiding the aspect ratio change unless we are closer to 90 or 270 which seems reasonable.

JackEAllen: Using `round(value/90) * 90` fails for cases 135 and 315 for me, setting aspect ratios to the next aspect ratio, 180 and 0 respectively where value is set to 135 will change aspect ratio to 180 instead of leaving at 90 and 315 to 0 instead of leaving at 270.

samtygier-stfc (changes requested):
Working well. Just a few last bits

samtygier-stfc (approved):
Working well.

Issue #1828, opened 2023-06-02
Performance of Spectrum Viewer Window Opening


JackEAllen:
### Desired Behaviour

**As a user**, I would like to be able to open large datasets within the spectrum viewer where the spectrum viewer window opens immediately, even if it take time for the data to be drawn in the GUI so I know that Mantid Imaging is doing something and has not crashed.

**As a developer**, I would like to be able to open the Spectrum Viewer GUI with no data as a prerequisite to loading of large data. This will allow for the creation of a system test which views the spectrum viewer before data is loaded and drawn.

## Optional
Load an empty image into the spectrum viewer with text stating data is loading when waiting for larger datasets to be drawn **OR** Create a progress indicator to indicate data is loading into the spectrum viewer.

### Current Behaviour

Spectrum Viewer window only loads once all data is ready to be drawn. This gives the perceived affect that nothing is happening and that Mantid Imaging has crashed.

PR #1860, opened 2023-07-14
Spec viewer performance part 1

samtygier-stfc:
### Issue

Progress #1828 

### Description

First steps on the issue. Removes some redundant work, and bits that should never happen.

Reduces call to `get_spectrum()` from 4 to 3. saving about 25% of window start time.

There is still more to do, but may as well get this merged.

### Testing 

Starting the spectrum viewer and using it should still work.

### Acceptance Criteria 

Enable performance logging. Check time with this vs on main

### Documentation
Not needed yet.


JackEAllen (approved):
Changes look good and slight speedup compared to main 👍

PR #1862, opened 2023-07-14
Slow mode for live view test script

samtygier-stfc:
### Issue

No issue

### Description

Add a slow mode that copies each file a chunk at at time.

For me this triggers an issue when the live view tries to read a partial file:

```
2023-07-14 15:59:41,157 [mantidimaging.gui.windows.main.view:L507] ERROR: Traceback (most recent call last):
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/live_viewer/model.py", line 34, in _handle_image_changed
    self.presenter.update_image(image_file)
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/live_viewer/presenter.py", line 46, in update_image
    image_data = tif.asarray()
  File "/home/sam/mambaforge/envs/mantidimaging-dev/lib/python3.10/site-packages/tifffile/tifffile.py", line 4265, in asarray
    result = page0.asarray(out=out, maxworkers=maxworkers)
  File "/home/sam/mambaforge/envs/mantidimaging-dev/lib/python3.10/site-packages/tifffile/tifffile.py", line 8923, in asarray
    for _ in self.segments(
  File "/home/sam/mambaforge/envs/mantidimaging-dev/lib/python3.10/site-packages/tifffile/tifffile.py", line 8724, in segments
    yield decode(segment)
  File "/home/sam/mambaforge/envs/mantidimaging-dev/lib/python3.10/site-packages/tifffile/tifffile.py", line 8712, in decode
    return func(decode(*args, **decodeargs))  # type: ignore
  File "/home/sam/mambaforge/envs/mantidimaging-dev/lib/python3.10/site-packages/tifffile/tifffile.py", line 8641, in decode_other
    data = decompress(data, out=size * dtype.itemsize)
  File "imagecodecs/_deflate.pyx", line 236, in imagecodecs._deflate.deflate_decode
imagecodecs._deflate.DeflateError: libdeflate_zlib_decompress returned LIBDEFLATE_BAD_DATA
```

### DIFF
--- a/scripts/simulate_live_data.py
+++ b/scripts/simulate_live_data.py
@@ -8,7 +8,7 @@ from tqdm import tqdm
 import time
 
 
-def copy_dataset(source_dir, dest_dir, rate):
+def copy_dataset(source_dir, dest_dir, rate, slow_copy_mode):
     """
     Copy data from a source directory to a destination directory.
     Files are copied one image at a time at a specified rate in seconds.
@@ -16,6 +16,7 @@ def copy_dataset(source_dir, dest_dir, rate):
     param source_dir: Source directory containing dataset to copy
     param dest_dir: Destination directory to copy dataset to
     param rate: Rate in seconds at which to copy files from dataset i.e. 1, 1.5, 2...
+    param slow_copy_mode: copy the file slowly in chunks
     """
     if not os.path.exists(dest_dir):
         os.makedirs(dest_dir)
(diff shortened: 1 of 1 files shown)


scripts/simulate_live_data.py
```diff
@@ -24,13 +25,32 @@ def copy_dataset(source_dir, dest_dir, rate):
...
+
+    if os.path.exists(dest):
+        os.remove(dest)
+
+    with open(dest, "wb") as dest_file:
+        writen = 0
+        while writen < data_size - 1000:
+            start = writen
+            stop = start + data_size // 10
+            writen += dest_file.write(data[start:stop])
```
JackEAllen, line 51: If I understand this correctly (I may not have), to calculate the start and stop, Instead of `data_size // 10`, should it be:
```Python
with open(dest, "wb") as dest_file:
        written = 0
        chunk_size = data_size // 10
        while written < data_size:
            start = written
            stop = written + chunk_size
            dest_file.write(data[start:stop])
            written += chunk_size
```
to determine the stop position for each chunk?

JackEAllen (changes requested):
Looks good. Have left a small suggestion if i understand correctly that the intention is copying the file in chunks of 10%. as I think what's happening here is that while loop will iterate until writen is less than data_size - 1000, which means it will stop when writen is very close to data_size?

scripts/simulate_live_data.py
samtygier-stfc: Oops, that ` - 1000` should not have been committed. I was trying to force some errors by skipping the last block.

That should do roughly the same. Though the last chunk will probably be a different size from the others (if the file size is not a multiple of 10), so the file value of `written` would be wrong. This would not matter much is would be larger than data_size and would stop the loop.

`write()` returns the number of bytes written, so the value should be correct. I'll admit it is a bit C-ish, but I think its more correct.

samtygier-stfc:
And I forgot to include the sleep :-)

JackEAllen (approved):
Looks good! thanks for updating this

JackEAllen:
> And I forgot to include the sleep :-)

Not as part of this PR, but should we find the need in the future, might be worth us passing in a delay time for how long to sleep to control the speed of the slow copy

PR #1863, opened 2023-07-14
Spec view refactoring

samtygier-stfc:
### Issue
More enabling work for #1828

### Description

This does some refactoring to simplify things when there is no stack loaded. It removes some code to treat the first ROI specially.

This causes quite a lot of churn in the tests, as the first ROI is now added in the presenter.

This will be useful as it solves some issues I hit trying to get the spectrum window to display before calculating spectrums.

### Testing 

Updates to many tests

### Acceptance Criteria 

Check that spectrum viewer works. Check adding and removing ROIs.

Its now possible to remove all the ROI, as I don't think there is much value to preventing this.

Check that things are sensible when all the images stacks are removed from the main window.

You can also test opening the spectrum window without any datasets loaded. By changing `mantidimaging/gui/windows/main/view.py` line 218 to always enable the `menuWorkflow`.

### Documentation

I'll update the release notes when #1828 is done

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -163,7 +163,7 @@ class SpectrumViewerWindowModel:
         else:
             return 0, 0
 
-    def can_export(self) -> bool:
+    def has_stack(self) -> bool:
         """
         Check if data is available to export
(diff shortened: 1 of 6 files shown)


samtygier-stfc:
image

Small visual change, the first ROI is selected at the start now.

PR #1868, opened 2023-07-26
1859 select log if found

### LINKED ISSUE #1859: Load logs by default
Currently in the load dataset dialog the log files do not get ticked to load when they are find automatically. They should be.

Steps to reproduce.
* File -> Load Dataset
* Select a sample with a well structure directory (the standard example should work)
* Note that even when the log file is detected, it is not enabled

image

Should be a quick fix in `LoadPresenter.update_field_with_filegroup()`

JackEAllen:
### Issue

Closes #1859 

### Description

If sample log is found, it is selected by default now in load dialog when loading a new dataset into MI

### Testing 

Tried Loading a dataset where sample log is found

### Acceptance Criteria 

- Compare against main branch.
- Load a dataset within MI using the menu.
- [ ] If the sample log file is found, the sample log is selected by default.
- [ ] Flat and Dark log files if detected are not selected by default.

### Documentation

`docs/release_notes/next/feature-1859-select_sample_log`

### DIFF
--- a/mantidimaging/gui/test/gui_system_loading_test.py
+++ b/mantidimaging/gui/test/gui_system_loading_test.py
@@ -108,11 +108,11 @@ class TestGuiSystemLoading(GuiSystemBase):
 
         self.assertEqual(len(self.main_window.presenter.datasets), 1)
         sample = list(self.main_window.presenter.datasets)[0].sample
-        self.assertNotIn("log_file", sample.metadata)
+        self.assertIn("log_file", sample.metadata)
 
-        # Initial angles are just evenly spaced from 0 to 2*pi
-        stack_len = sample.num_images
-        self.assertAlmostEqual(sample.projection_angles().value[1], 2 * math.pi / (stack_len - 1), 12)
+        # Assert initial angles are the same as log_file angles
+        log_angle = math.radians(self._get_log_angle(log_path))
+        self.assertAlmostEqual(sample.projection_angles().value[1], log_angle, 12)
 
         # Load sample log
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/test/gui_system_loading_test.py
```diff
@@ -108,11 +108,11 @@ def test_load_log(self, mocked_select_file):
...
         sample = list(self.main_window.presenter.datasets)[0].sample
-        self.assertNotIn("log_file", sample.metadata)
+        self.assertIn("log_file", sample.metadata)
 
-        # Initial angles are just evenly spaced from 0 to 2*pi
-        stack_len = sample.num_images
-        self.assertAlmostEqual(sample.projection_angles().value[1], 2 * math.pi / (stack_len - 1), 12)
+        # Assert initial angles are the same as log_file angles
+        log_angle = math.radians(self._get_log_angle(log_path))
+        self.assertAlmostEqual(sample.projection_angles().value[1], log_angle, 12)
```
samtygier-stfc, line 115: I think it might work better here to remove the existing log, so that we can check that `actionSampleLoadLog` does something. Otherwise the test still passes if I comment out lines 118 to 120.

I'd roll back the changes, and instead put after line 110
```
        sample.log_file = None
        sample._projection_angles = None
```

And the same in `test_load_angles()`

samtygier-stfc (changes requested):
Just a small change to the testing.

samtygier-stfc (approved):
Looks good. Thanks.

PR #1877, opened 2023-07-28
Find flat and dark when names differ

### LINKED ISSUE #1865: User Bug Report: Load Data Doesn't Find Flat and Dark Folders and Filenames
### Summary

Mantid Imaging no longer finds flat and dark folders and filenames after selection of sample projections.

### Steps To Reproduce

* Open Mantid Imaging
* Load a Sample selecting sample projections
* Notice that Flat and Dark folders are not loaded

### Expected Behaviour

Flat and dark folders should be loaded automatically.

### Current Behaviour

Flat and dark folders are not loaded automatically.

### Context

Latest Nightly build.

samtygier-stfc:
### Issue

Closes #1865

### Description
Make finding the flat and dark files less strict.

Now just look for the first image in a folder with the expected name.

### Testing 

### Acceptance Criteria 

### Documentation

### DIFF
--- a/mantidimaging/core/io/test/filenames_test.py
+++ b/mantidimaging/core/io/test/filenames_test.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 from pathlib import Path
 import unittest
 
-import pytest
 from parameterized import parameterized
 
 from mantidimaging.test_helpers.unit_test_helper import FakeFSTestCase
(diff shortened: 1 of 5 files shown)


mantidimaging/core/io/test/filenames_test.py
```diff
@@ -184,6 +197,8 @@ def test_find_log_best(self):
         ("/a/Tomo/foo_Tomo_%06d.tif", "/a/Flat_Before/foo_Flat_Before_%06d.tif"),
```
JackEAllen, line 197: NITPICK: We should probably have a test following this structure to test invalid naming as well.

JackEAllen (approved):
Changes look good and work as expected, thanks for putting this fix in. 👍 
I have left a few minor comments about test format and possibly additional tests to check invalid naming. I am neutral on both points so will approve. If you don't agree, feel free to merge in otherwise I can re-review once changes are made :)

mantidimaging/core/io/test/filenames_test.py
samtygier-stfc: Might need to wait until we have an example of something that we should not have found.

PR #1879, opened 2023-08-01
1875 file validation

### LINKED ISSUE #1875: LIVE Viewer: File Error Validation Handling for Partial Files
### Desired Behaviour

Provide error validation handling for slow copies of files arriving in a directly within the Live viewer

### Current Behaviour

The Live Viewer will return errors if a file is being copied into the selected directly slowly such that the file arrives in the directly before it has been fully written. 

This behaviour can be reproduced using the file generator script (`scripts/simulate_live_data.py`)

JackEAllen:
### Issue

Closes #1875 

### Description

* Add additional error handling when trying to read a file that may not have been fully written to directory being watched by Live Viewer.
* Check file size before read to only read files once past the ~minimum size a valid tif file can be (at least according to [this stackoverflow thread](https://stackoverflow.com/questions/42787235/smallest-possible-valid-tif-file))
* Fix minor issue with missed argument in `scripts/simulate_live_data.py`

### Testing 

Tested slow copy and quick copies of files into directory being watched by live viewer. 

### Acceptance Criteria 

* Slow copies work (run in debug mode to view file validation check status's)
* Fast copy still works
* Tests pass

### Documentation

* `docs/release_notes/next/fix-1875-live_viewer_file_state_validation`

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/model.py
+++ b/mantidimaging/gui/windows/live_viewer/model.py
@@ -73,14 +73,25 @@ class ImageWatcher(QObject):
         self.find_images()
         self.find_last_modified_image()
 
+    def _validate_file(self, file_path) -> bool:
+        """
+        Check if a file is valid.
+        """
+        return file_path.is_file() and self._is_image_file(file_path.name)
+
     def _get_image_files(self):
         image_files = []
         for file_path in Path(self.directory).iterdir():
-            if file_path.is_file() and self._is_image_file(file_path.name):
+            file_size = file_path.stat().st_size
+            if file_size > 45 and self._validate_file(file_path):
+                LOG.debug(f'VALID FILE: {file_path} is an image file and is not empty')
                 image_files.append(file_path)
+            else:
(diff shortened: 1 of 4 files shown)


mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -87,14 +91,37 @@ def _handle_file_change(self, file):
...
 
+    def _try_to_read_file(self, file_path) -> bool:
+        """Try and read file, return True if successful and False otherwise"""
+        try:
+            with tifffile.TiffFile(file_path) as tif:
+                tif.asarray()
+                tif.close()
+        except Exception:
+            return False
+        return True
```
samtygier-stfc, line 102: Would it work to drop this, and just let it fail in the presenter? It would save doing the load twice.

JackEAllen: should do!

samtygier-stfc (approved):
Looks good

PR #1883, opened 2023-08-02
1882 file signal emit efficiency

### LINKED ISSUE #1882: Live Viewer: Explore File Emit Signals - Improve Efficiency of Image_changed signal Emits
### Summary

Image_changed may be emitted too often, emitting for every image in the watched directory, Reduce this if proven to to be the case to only emit once for each new image added/modified in the watched directory.

`_handle_file_change()` may not be needed. remove if redundant and rely only on `_handle_directory_change()`

Hangs are also present on windows which should be explored.

### Steps To Reproduce

Add logs to view signal emits when directory changed and when images are loaded. 

### Expected Behaviour

Only the minimum number of signal emits should be sent when new images arrive in the watched directory.

### Current Behaviour

Potentially more than the minimum number of signals are emitted on changed to the watched directory which could slow down the live viewer. 

### Context

This issue needs to be completed to allow for #1879 to be tested fully.

JackEAllen:
### Issue

Closes #1882 

### Description

* Additional Logging
* Reduce file_change signal emits

### Testing 

Add logs proposed in PR changes to main and compare emit frequency (note that the additional emits for directory changes will be reduced in #1879)

### Acceptance Criteria 

* Compare number of file change emits to main.

### Documentation

docs/release_notes/next/fix-1882-reduce_file_update_signal_emits

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/presenter.py
+++ b/mantidimaging/gui/windows/live_viewer/presenter.py
@@ -42,6 +42,7 @@ class LiveViewerWindowPresenter(BasePresenter):
         if not Path(image_path).exists():
             return
         try:
+            logger.debug("Showing image in presenter: %s", image_path)
             with tifffile.TiffFile(image_path) as tif:
                 image_data = tif.asarray()
                 self.view.show_image(image_data)
--- a/docs/release_notes/next/fix-1882-reduce_file_update_signal_emits
+++ b/docs/release_notes/next/fix-1882-reduce_file_update_signal_emits
@@ -1,0 +1,1 @@
+#1882 : Additional Logging of file change emit signals and reduction in file change emit signals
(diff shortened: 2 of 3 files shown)


mantidimaging/gui/windows/live_viewer/presenter.py
```diff
@@ -45,6 +45,7 @@ def update_image(self, image_path: str):
             with tifffile.TiffFile(image_path) as tif:
                 image_data = tif.asarray()
                 self.view.show_image(image_data)
+                logger.debug("Showing image in presenter: %s", image_path)
```
samtygier-stfc, line 48: Small thing: I'd put the log message before the `with`, so that we see the message even if any of the open, load, show steps fail.

samtygier-stfc (approved):

PR #1895, opened 2023-08-07
Fix python 3.10 walk_packages needing a str not a path

### LINKED ISSUE #1894: Pyinstaller wont build on linux
### Summary

We hit https://github.com/python/cpython/issues/88227 in PackageWithPyInstaller.py

In python 3.10 `pkgutil.walk_packages()` fails if the path is a `pathlib.Path` on linux

### Steps To Reproduce

Run `PackageWithPyInstaller.py`

### Expected Behaviour

Builds package

### Current Behaviour

`AttributeError: 'PosixPath' object has no attribute 'startswith'`

### Context

Current main. Probably since switching to python 3.10

samtygier-stfc:
### Issue

Closes #1894

### Description

In python 3.10 pkgutil.walk_packages() fails if the path is a pathlib.Path on linux.

Convert it to a string, and add a `COMPAT` message so the workaround can be removed in the future

### Testing 
Not needed

### Acceptance Criteria 

Check that PackageWithPyInstaller.py runs on Linux

### Documentation

release notes

### DIFF
--- a/docs/release_notes/next/fix-1894-fix-pyinstaller-path
+++ b/docs/release_notes/next/fix-1894-fix-pyinstaller-path
@@ -1,0 +1,1 @@
+#1894 : Fix for path in PackageWithPyInstaller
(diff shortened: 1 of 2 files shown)


JackEAllen (approved):
Looks good! Thanks for the quick fix

PR #1896, opened 2023-08-09
Ignore files starting with dot when finding flat/dark

### LINKED ISSUE #1888: Avoid matching names like ._IMAT_Flower_Flat_Before_000000.tif
### Summary

We have a dataset that has some garbage files with names like `._IMAT_Flower_Flat_Before_000000.tif`. These are possibly caused during an interrupted file copy.

#1878 made the file pattern matching a lot less strict, so we now pick up these files as candidates for the flat and dark. We should probably ignore any file starting with a `.` as those are very unlikely to be actual valid data files.

### Steps To Reproduce

Make a copy of a dataset, eg IMAT00010675
In the `Flat_Before` make some problem files
`for f in *.tif ; do touch ._${f} ; done`
In MI do Load Dataset
Select the sample from the Tomo directory
Note that `._IMAT_Flower_Flat_Before_000000.tif` is picked up as the Flat Before
Load and note the error message

### Expected Behaviour

Files starting with a `.` should not be picked automatically

### Context

Probably introduced by #1878

samtygier-stfc:
### Issue
Closes #1888

### Description

When searching for releated files, filter out files starting with a `.` these are unlikely to be actual image files.

### Testing 

Added some stronger tests

### Acceptance Criteria 

Follow instructions on the bug report. Check that the actual files get selected and opened.

### Documentation
release notes

### DIFF
--- a/docs/release_notes/next/fix-1888-ignore-dotfiles
+++ b/docs/release_notes/next/fix-1888-ignore-dotfiles
@@ -1,0 +1,1 @@
+#1888 : Don't return files beginning with a dot when finding related
--- a/mantidimaging/core/io/filenames.py
+++ b/mantidimaging/core/io/filenames.py
@@ -157,6 +157,8 @@ class FilenameGroup:
         for file in sorted(path.iterdir()):
             if file.suffix.lstrip(".") not in IMAGE_FORMAT_EXTENSIONS:
                 continue
+            if file.name.startswith("."):
+                continue
             fg = cls.from_file(file)
             return fg
 
--- a/mantidimaging/core/io/filenames.py
+++ b/mantidimaging/core/io/filenames.py
@@ -154,13 +154,20 @@ class FilenameGroup:
         if not path.is_dir():
             raise ValueError(f"path is a file: {path}")
 
-        files = (f for f in path.iterdir() if f.name[0] != '.' and f.suffix[1:] in IMAGE_FORMAT_EXTENSIONS)
+        files = (f for f in path.iterdir() if cls.valid_image_filename(f))
 
         try:
             first_file = min(files)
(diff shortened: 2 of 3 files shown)


docs/release_notes/next/fix-1888-ignore-dotfiles
```diff
@@ -0,0 +1 @@
+#1888 : Don't return files beggining with a dot when finding related
```
JackEAllen, line 1: MINOR: beggining -> beginning

mantidimaging/core/io/filenames.py
```diff
@@ -154,13 +154,13 @@ def from_directory(cls, path: Union[Path, str]) -> FilenameGroup | None:
         if not path.is_dir():
             raise ValueError(f"path is a file: {path}")
 
-        for file in sorted(path.iterdir()):
-            if file.suffix.lstrip(".") not in IMAGE_FORMAT_EXTENSIONS:
-                continue
-            fg = cls.from_file(file)
-            return fg
+        files = (f for f in path.iterdir() if f.name[0] != '.' and f.suffix[1:] in IMAGE_FORMAT_EXTENSIONS)
```
JackEAllen, line 157: We should probably add the class method to a class docstring so when we forget why 
```python
f for f in path.iterdir() if f.name[0] != '.'
```
exists so we don't remove it as trivial code and update tests to not test this case.

JackEAllen: - At least if i was new dev in the team an no-one else in the team had long standing knowledge of the project, I'd consider removing it if i had not prior context of this issue.

JackEAllen (approved):
Looks good! 👍  Thanks for change.

PR #1908, opened 2023-08-21
Create Image_Data Class and Store All Images in Array

### LINKED ISSUE #1904: image scroller - in modified order
* scrollbar
* tracking imgs in dir
*

JackEAllen:
### Issue

Closes #1904 

### Description

- Create DataClass to store Image data
- Store list of Image objects to allow for re-ordering of images and displaying of all images

### Testing 

Manual testing to check new images are loaded in.
If images already exist, in directory, they are loaded into list
If images are deleted (one at a time or all), Live viewer list of images updated to reflect change. 

### Acceptance Criteria 
Behaviour is the same as main branch:
- live viewer collects new images as they are added to directory
- live viewer handles deleting of files fro dir
- live viewer can load dir if images are already in dir
- live viewer only loads in complete files that are valid `.tiff` or `.tif`

### Documentation

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/model.py
+++ b/mantidimaging/gui/windows/live_viewer/model.py
@@ -12,12 +12,96 @@ if TYPE_CHECKING:
 LOG = getLogger(__name__)
 
 
+class Image_Data:
+    """
+    Image Data Class to store represent image data.
+
+    ...
+
+    Attributes
+    ----------
+    image_path : str
+        path to image file
+    image_name : str
+        name of image file
+    image_size : int
+        size of image file
+    image_modified_time : int
+        last modified time of image file
+    """
+
+    def __init__(self, image_path: Path):
+        """
+        Constructor for Image_Data class.
+
+        Parameters
+        ----------
+        image_path : str
+            path to image file
+        """
+        self.image_path: Path = image_path
+        self.image_name: str = image_path.name
+        self._image_size: int
+        self._image_modified_time: float
+        self._set_file_info()
+
+    def _set_file_info(self):
+        self.image_size = self.image_path
+        self.image_modified_time = self.image_path
+
+    @property
+    def image_size(self):
+        """Return the image size"""
+        return self._image_size
+
+    @image_size.setter
+    def image_size(self, image_path: str):
+        """
+        assume image may be deleted before objet is fully created
+        """
+        try:
+            self._image_size = Path(image_path).stat().st_size
+        except FileNotFoundError:
+            self._image_size = 0
+
+    @property
+    def image_modified_time(self):
+        """Return the image modified time"""
+        return self._image_modified_time
+
+    @image_modified_time.setter
+    def image_modified_time(self, image_path: str):
+        """ Assume image may be deleted before object is fully created"""
+        try:
+            self._image_modified_time = Path(image_path).stat().st_mtime
+        except FileNotFoundError:
+            self._image_modified_time = 0
+
+
 class LiveViewerWindowModel:
     """
     The model for the spectrum viewer window.
+
+    ...
+
+    Attributes
+    ----------
+    presenter : LiveViewerWindowPresenter
+        presenter for the spectrum viewer window
+    path : str
+        path to dataset
     """
 
     def __init__(self, presenter: 'LiveViewerWindowPresenter'):
+        """
+        Constructor for LiveViewerWindowModel class.
+
+        Parameters
+        ----------
+        presenter : LiveViewerWindowPresenter
+            presenter for the spectrum viewer window
+        """
+
         self.presenter = presenter
         self._dataset_path = None
 
@@ -33,7 +117,14 @@ class LiveViewerWindowModel:
         self.image_watcher.find_images()
         self.image_watcher.find_last_modified_image()
 
-    def _handle_image_changed(self, image_file):
+    def _handle_image_changed(self, image_file) -> None:
+        """
+        Handle an image changed event. Update the image in the view.
+        This method is called when the image_watcher detects a change
+        which could be a new image, edited image or deleted image.
+
+        :param image_file: path to image file
+        """
         if image_file == '':
             self.presenter.handle_deleted()
         self.presenter.update_image(image_file)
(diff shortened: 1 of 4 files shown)


mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -12,12 +12,100 @@
...
+        Parameters
+        ----------
+        image_path : str
+            path to image file
+        """
+        self.image_path: Path = image_path
+        self.image_name: str = image_path.name
+        self._image_size: int
+        self._image_modified_time: float
+        self._set_file_info()
```
samtygier-stfc, line 46: Can I suggest that only the path and result of `stat()` are stored as attributes, and then properties are used to get the other values.

```python
    self.image_path = image_path
    self.stat = image_path.stat()
```

Then you can drop the setters. And have

```python
    @property
    def image_size(self):
        """Return the image size"""
        return self.stat.st_size
```
This would simplify this class, and mean that `stat()` only gets called once. And it saves the setters having a different input than the value.

Maybe let the exception bubble up and catch it in _get_image_files() if needed.

Also, you should not need to type annotate `self.image_path` because it is set from an annotated argument to `__init__`.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -29,85 +117,147 @@ def path(self):
...
         self.image_watcher = ImageWatcher(self.path)
-        self.image_watcher.image_changed.connect(self._handle_image_changed)
+        self.image_watcher.image_changed.connect(self._handle_image_changed_in_list)
         self.image_watcher.find_images()
-        self.image_watcher.find_last_modified_image()
+        self.image_watcher.get_images()
 
-    def _handle_image_changed(self, image_file):
-        if image_file == '':
+    def _handle_image_changed_in_list(self, image_file: list[Image_Data]) -> None:
```
samtygier-stfc, line 124: `image_file` -> `image_files`

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -29,85 +117,147 @@ def path(self):
...
         for file_path in Path(self.directory).iterdir():
-            if not self._validate_file(file_path):
-                LOG.debug(f'INVALID FILE: {file_path} is not valid an image file')
-                continue
             file_size = file_path.stat().st_size
             if file_size > 45 and self._validate_file(file_path):
-                image_files.append(file_path)
+                image_files.append(Image_Data(file_path))
             else:
                 LOG.debug(f'INVALID FILE: {file_path} is not valid an image file or is empty')
```
samtygier-stfc, line 249: I think the checks can be simplified here, so that it does not need to check size again.

Maybe check the filename looks like an image, then try to create the `Image_Data`, if that works, and its size is `>45` then append it to the list.

mantidimaging/gui/windows/live_viewer/presenter.py
```diff
@@ -46,20 +45,16 @@ def handle_deleted(self):
         self.view.remove_image()
         self.clear_label()
 
-    def update_image(self, image_path: str):
+    def update_image(self, images_list: list):
         """Update the image in the view."""
-        if not Path(image_path).exists():
-            return
-        if image_path == '':
+        if not images_list and len(images_list) < 1:
```
samtygier-stfc, line 50: I don't think the `not images_list`. We have a type annotation, so `images_list` must be a list, so `len` should be safe. Or could just check `not images_list`, as that will be false for an empty list.

mantidimaging/gui/windows/live_viewer/view.py
```diff
@@ -36,9 +36,9 @@ def __init__(self, main_window: 'MainWindowView', live_dir_path: Path) -> None:
         # reposition to the right of the main window to be visible when launched from cli
         self.move(self.main_window.x() + self.main_window.width(), self.main_window.y())
 
-    def show_image(self, image: np.ndarray) -> None:
+    def show_most_recent_image(self, image: np.ndarray) -> None:
```
samtygier-stfc, line 39: Best to leave the name as `show_image()` because once the user can move the slider this will be used to show other images.

samtygier-stfc (changes requested):

mantidimaging/gui/windows/live_viewer/model.py
JackEAllen: I mostly agree with this, I will still need to wrap stat in a try catch as stat may not exist when trying to set initially if the file is removed or in a partial state during initialisation.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -29,85 +100,148 @@ def path(self):
...
         for file_path in Path(self.directory).iterdir():
-            if not self._validate_file(file_path):
-                LOG.debug(f'INVALID FILE: {file_path} is not valid an image file')
-                continue
-            file_size = file_path.stat().st_size
-            if file_size > 45 and self._validate_file(file_path):
-                image_files.append(file_path)
+            image_obj = Image_Data(file_path)
+            if self._validate_file(file_path) and image_obj.image_size > 45:
+                image_files.append(image_obj)
```
samtygier-stfc, line 230:
```
for file_path in Path(self.directory).iterdir():
            if not _is_image_file(file_path): continue
            try:
                image_obj = Image_Data(file_path)
            execpt FileNotFound:
                continue
            if  image_obj.image_size > 45:
                image_files.append(image_obj)
```

JackEAllen: I've refactored this into a generator.
Another alternative solution could be:
```Python
image_files = []
for file_path in Path(self.directory).iterdir():
     if self._is_image_file(file_path.name):
         image_obj = Image_Data(file_path)
         if image_obj.image_size > 45: image_files.append(image_obj)
 sorted_images = self.sort_images_by_modified_time(image_files)
 return sorted_images
```

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -29,85 +93,125 @@ def path(self):
...
-
-            self.image_changed.emit(str(last_modified_image))
-
-    def _handle_directory_change(self, directory):
-        self.find_images()
-        if self.images and self._validate_images(self.images):
-            self.find_last_modified_image()
-        else:
-            self.image_changed.emit('')
+                return []
```
samtygier-stfc, line 184: I don't think this can fail. `self.images` is just a list that might be empty. Think we can either just return the list, or let the caller use the attribute directly.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -29,85 +93,125 @@ def path(self):
...
-            file_size = file_path.stat().st_size
-            if file_size > 45 and self._validate_file(file_path):
-                image_files.append(file_path)
-            else:
-                LOG.debug(f'INVALID FILE: {file_path} is not valid an image file or is empty')
-        return image_files
+        image_files = [
+            image_obj for file_path in Path(self.directory).iterdir()
+            if self._is_image_file(file_path.name) and (image_obj := Image_Data(file_path)).image_size > 45
+        ]
```
samtygier-stfc, line 204: I think the try/except need to be inside the loop. Otherwise one bad file causing an exception in `Image_Data(file_path)` will jump us out of ` _get_image_files()` so will result in no images being found.

samtygier-stfc (changes requested):
Looking much better, just a couple of thoughts on error handling.

mantidimaging/gui/windows/live_viewer/model.py
samtygier-stfc: The diff has shown in a strange way. The comment is for the lines
```python
    def get_images(self):
        """Return the sorted images"""
        if self.images:
            try:
                return self.images
            except FileNotFoundError:
                return []
```

JackEAllen: I'll keep it in it's own method for the time being in case we choose to modify to include additional logic later on.

samtygier-stfc (approved):
Looking good.

PR #1921, opened 2023-09-07
1889 main window priority cli

### LINKED ISSUE #1889: Windows opened from the CLI should be on top of main window
### Summary
When using the command line arguments to open operations, recon or live viewer windows, they should be opened on top of the main window.

### Steps To Reproduce

`python -X faulthandler -W default  -m mantidimaging --log-level DEBUG --path ~/mantidimaging-data/ISIS/IMAT/IMAT00010675/Tomo --operations`

### Expected Behaviour

Operation window should be visible

### Current Behaviour

Operations window is behind the main window

### Context

Current main, and since forever

JackEAllen:
### Issue

Closes #1889 

### Description

When launching mantid imaging from the CLI with flags to launch windows such as: operations window; reconstruction window; live viewer - the specified window now stay on-top of the main window. 

Previously, the main window would display over the top of additional windows when launched from the CLI.

### Testing 
Checkout this branch locally and launch mantid imaging from the CLI with additional flags to open one of the following windows:
 * Operations window
 * Reconstruction window
 * Live viewer WIndow. 

Verify that the selected window from the CLI opens above the main window. 
 
##### **Operations Window:**
```bash
python -m mantidimaging --path <DATASET_DIR> --operation crop-coordinates
```

##### **Reconstruction Window:**
```bash
python -m mantidimaging --path <DATASET_DIR> --recon
```

##### **Live Viewer Window:**
```bash
python -m mantidimaging -lv <LIVE_OUTPUT_DIR>
```

### Acceptance Criteria 

- Additional windows opened when mantid imaging is launched from the CLI open on top of main window
    - [ ] Operations Window
    - [ ] Reconstruction Window
    - [ ] Live View Window

### Documentation

- `docs/release_notes/next/dev-1889-cli_launched_window_focus`

### DIFF
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -168,12 +168,25 @@ class MainWindowPresenter(BasePresenter):
         else:
             self._handle_task_error(self.LOAD_ERROR_STRING, task)
 
+    def _open_window_if_not_open(self) -> None:
+        """
+        Launches windows that requires loaded data if the CLI flags are set.
+        Resets args after window has opened.
+        """
+        if self.view.args.operation() != "" and self.view.filters is None:
+            self.show_operation(self.view.args.operation())
+            self.view.args.clear_window_args()
+        if self.view.args.recon() and self.view.recon is None:
+            self.view.show_recon_window()
+            self.view.args.clear_window_args()
+
     def _on_dataset_load_done(self, task: 'TaskWorkerThread') -> None:
 
         if task.was_successful():
             self._add_strict_dataset_to_view(task.result)
             self.view.model_changed.emit()
             task.result = None
+            self._open_window_if_not_open()
         else:
             self._handle_task_error(self.LOAD_ERROR_STRING, task)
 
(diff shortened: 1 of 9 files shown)


mantidimaging/gui/windows/main/presenter.py
```diff
@@ -168,12 +168,22 @@ def _on_stack_load_done(self, task: 'TaskWorkerThread') -> None:
...
+        if self.view.args.recon():
+            self.view.show_recon_window()
+
     def _on_dataset_load_done(self, task: 'TaskWorkerThread') -> None:
 
         if task.was_successful():
             self._add_strict_dataset_to_view(task.result)
             self.view.model_changed.emit()
             task.result = None
+            self._launch_window_if_cli_flag_set()
```
samtygier-stfc, line 186: Will this cause the operation or recon window to reopen if the user loads another dataset?

Maybe `load_stacks_from_folder()` could create a function like
```python
    def on_load_and_finish_cli_processing(task):
        self._on_dataset_load_done(task)
        self._finish_processing_cli_args()
```
And pass that to `start_async_task_view()`

JackEAllen: Nice catch. Doesn't seem to reopen as the widow is already open, but it does change the focus to place the recon window over the main window again as the active window so current implementation will need altered slighty to avoid this.

PR #1923, opened 2023-09-12
1922 live view emit dir structure on load

### LINKED ISSUE #1922: Live Viewer: Load Existing Images From Selected Directory
### Summary

When launching the live viewer to watch a selected directory where that directory is already populated with Images no longer works. The live viewer displays a blank window.

This is likely because the live viewer no longer receives an event to emit a signal until a change occurs in the directory.

### Steps To Reproduce

* Populate a directory (`LIVE_OUTPUT_DIR/`) with some tomo images.
* Open the live viewer to watch the directory: `python -m mantidimaging -lv <LIVE_OUTPUT_DIR/>`
* Verify that the live viewer only displays a blank window that does not include images from dir until there is a change in the directory - i.e. adding a new image to the directory.

### Expected Behaviour

Live viewer displays any images already present within the directory being watched if present before launching live viewer window.

### Current Behaviour

Live viewer displays a blank window instead of images present in the watched directory if populated before launching the live viewer. 

### Context

* Bug found when testing on ubuntu machine
* Bug was likely introduced by being missed in testing to to a quick rollout of multiple new changes touching the same code.

### Screenshot(s)

Screenshot 2023-09-07 at 14 06 13

JackEAllen:
### Issue

Closes #1922 

### Description

View makes a call on initialisation to the model to emit a signal to load images so that if images are already present in the selected directory, they are loaded into the live viewer. Previously Images if present were not loaded into live viewer on initialisation. 

### Testing 

Try loading the image viewer with and without images already populated inside the selected directory. 
1. Create a empty directory
2. Load live viewer using CLI flag `python -W default -m mantidimaging --log-level DEBUG -lv <LIVE_DIR_OUT/>` or from the menu: `File>Live Viewer` select a folder.
3. Verify that the live viewer window opens with no data loaded successfully
4. Begin copying image files over to the live viewer using the utility script `scripts/simulate_live_data.py`: `python simulate_live_data.py -s ../<LIVE_DIR_IN/>/ -d <../LIVE_DIR_OUT/>/ --faulty`
and verify images display in the live viewer.
5. Close Mantid Imaging and repeat steps with the populated directory verifying that the live viewer shows loaded data on opening. 

### Acceptance Criteria 
* Live viewer opens successfully when selected with an empty directory.
* Live viewer displays images present in a selected directory that already contains images.

### Documentation

`docs/release_notes/next/fix-1922-live_view_image_load_on_init`

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -32,14 +32,13 @@ class LiveViewerWindowView(BaseMainWindowView):
         self.presenter = LiveViewerWindowPresenter(self, main_window)
         self.live_viewer = LiveViewWidget()
         self.imageLayout.addWidget(self.live_viewer)
+        self.live_viewer.z_slider.valueChanged.connect(self.presenter.select_image)
 
     def show(self) -> None:
         """Show the window"""
         super().show()
         self.activateWindow()
         self.watch_directory()
-
-        self.live_viewer.z_slider.valueChanged.connect(self.presenter.select_image)
 
     def show_most_recent_image(self, image: np.ndarray) -> None:
         """
--- a/docs/release_notes/next/fix-1922-live_view_image_load_on_init
+++ b/docs/release_notes/next/fix-1922-live_view_image_load_on_init
@@ -1,0 +1,1 @@
+#1922 : Load images into live viewer on initialisation if images already present in selected directory.
--- a/mantidimaging/gui/windows/live_viewer/model.py
+++ b/mantidimaging/gui/windows/live_viewer/model.py
@@ -100,7 +100,7 @@ class LiveViewerWindowModel:
         self._dataset_path = path
         self.image_watcher = ImageWatcher(path)
         self.image_watcher.image_changed.connect(self._handle_image_changed_in_list)
-        self.image_watcher.find_images()
+        self.image_watcher._handle_directory_change("")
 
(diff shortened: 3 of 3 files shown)


mantidimaging/gui/windows/live_viewer/view.py
```diff
@@ -52,6 +53,10 @@ def watch_directory(self) -> None:
         """Show the most recent image arrived in the selected directory"""
         self.presenter.set_dataset_path(self.path)
 
+    def load_exisitng_images(self) -> None:
+        """Load the existing images in the selected directory if present"""
+        self.presenter.model.get_images()
+
```
samtygier-stfc, line 59: This is not used. I'd say its fine to call the `get_images()` directly than to have a pass through function.

samtygier-stfc (changes requested):
Fine with the unused function removed.

Wondering if this could just be added directly to `ImageWatcher`, so that it calls `self._handle_directory_change()` in its constructor. Is there a time we'd want to create and ImageWatcher and not an immediate file list?

mantidimaging/gui/windows/live_viewer/view.py
JackEAllen: I also considered adding directly to ImageWatcher, but couldn't find a nice way to call `self._handle_directory_changes()` that didn't cause `TypeError: native Qt signal is not callable` when called from within the same class and add duplication of `_handle_image_changed_in_list` as `self.images` is needed from `LiveViewerWindowModel` to include the latest images.

PR #1934, opened 2023-09-20
Fix showing most recent image

### LINKED ISSUE #1933: Live viewer not keeping up with latest image
### Summary

When copying images slowly to the live directory, the displayed images can be 2 behind the most recent image

### Steps To Reproduce

Use the simulation script in slow mode with a big enough delay to see what is happening.

`IMAT00010675_128_label.tar.xz` in the shared folder can be useful for checking what is being shown.

### Expected Behaviour

Most recent image get shown to user

### Current Behaviour

First image not shown until 3rd has started copying

### Context

Current main

samtygier-stfc:
### Issue

Closes #1933

### Description
This resolves 2 issues that prevent the most recent image being shown.
- A signalling issue, where a new image was only shown if its arrival had caused the position on the z-slider to change
- Lack of notification when an image file is modified. So the loading was being trigged at the start of writing the file, so we try to load it before it is complete.

I have also added an indicator when the image cannot be loaded. We get this situation a bit more now, because we don't filter out sub 45 byte files.

Should also fix the snap folder issue, as we keep a watch on the most recent file.

Use the `simulate_live_data.py` script with `--slow`.
Now the image should be shown as soon as writing is complete.

Note, that during the slow write, we display a blank image, as the most recent image is not loadable.

### Documentation

Not needed

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from pathlib import Path
 from typing import TYPE_CHECKING
 
+from PyQt5.QtCore import QSignalBlocker
 from PyQt5.QtWidgets import QVBoxLayout
 
 from mantidimaging.gui.mvp_base import BaseMainWindowView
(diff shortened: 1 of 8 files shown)


mantidimaging/gui/windows/live_viewer/view.py
```diff
@@ -55,8 +56,14 @@ def remove_image(self) -> None:
...
 
+    def set_image_range(self, index_range: tuple[int, int]) -> None:
+        with QSignalBlocker(self.live_viewer.z_slider):
+            self.live_viewer.z_slider.set_range(*index_range)
+
     def set_image_index(self, index: int) -> None:
-        self.live_viewer.z_slider.set_value(index)
+        with QSignalBlocker(self.live_viewer.z_slider):
+            self.live_viewer.z_slider.set_value(index)
+        self.live_viewer.z_slider.valueChanged.emit(index)
```
JackEAllen, line 66: NITPICK: I would probably want docstring on both of these to explain why we need to use `QSignalBlocker` so there's no possibility of removing later down the line when the context for them being added is forgotten.

JackEAllen (approved):
Looks good! Only a slight nitpick to add docstrings for future reference to describe context for `QSignalBlocker` usage, but other than that, happy to approve. 👍

PR #1942, opened 2023-10-05
Spectrum viewer tab for exporting to RITS

### LINKED ISSUE #1941: Spectrum Viewer: RITS Front-End GUI
### Desired Behaviour

Create a basic front-end GUI for exporting to RITS format within the spectrum viewer.

### Criteria
* RITS export should be accessible via a tab in the spectrum viewer window as denoted in the wireframe design below.
* One ROI, the same size as the FoV should be initially loaded similarly to default spectrum viewer state.
* It should be possible for the ROI to be resized and repositioned within the bounds of the FoV.
* A numeric entry box should be provided to allow users to set the binning size.
* A separate box should also be included to set the moving average.
* The export RITS button should export the ROI spectrum data in RITS format.

image

### Current Behaviour

There is currently no front-end GUI for exporting to RITS format within Mantid Imaging.

samtygier-stfc:
### Issue

Closes #1941

### Description

Start of the tab for RITS export.
Hooked up to existing export code

TODO
ROI hiding

### Testing 

Tests around enabling export buttons were updated and improved

### Acceptance Criteria 

Check that the export button triggers an export

### Documentation

Release notes

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -1,6 +1,8 @@
 # Copyright (C) 2023 ISIS Rutherford Appleton Laboratory UKRI
 # SPDX - License - Identifier: GPL-3.0-or-later
 from __future__ import annotations
+
+from enum import Enum
 from typing import TYPE_CHECKING, Optional
 
 from logging import getLogger
@@ -17,6 +19,12 @@ if TYPE_CHECKING:
 LOG = getLogger(__name__)
 
 
+class ExportMode(Enum):
+    # Needs to match GUI tab order
+    ROI_MODE = 0
+    IMAGE_MODE = 1
+
+
 class SpectrumViewerWindowPresenter(BasePresenter):
     """
     The presenter for the spectrum viewer window.
@@ -29,6 +37,7 @@ class SpectrumViewerWindowPresenter(BasePresenter):
     spectrum_mode: SpecType = SpecType.SAMPLE
     current_stack_uuid: Optional['UUID'] = None
     current_norm_stack_uuid: Optional['UUID'] = None
+    export_mode: ExportMode
 
     def __init__(self, view: 'SpectrumViewerWindowView', main_window: 'MainWindowView'):
         super().__init__(view)
@@ -36,6 +45,7 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         self.view = view
         self.main_window = main_window
         self.model = SpectrumViewerWindowModel(self)
+        self.export_mode = ExportMode.ROI_MODE
 
     def handle_sample_change(self, uuid: Optional['UUID']) -> None:
         if uuid == self.current_stack_uuid:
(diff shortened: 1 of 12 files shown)


mantidimaging/gui/windows/spectrum_viewer/view.py
```diff
@@ -142,6 +138,22 @@ def cleanup(self):
...
+            if self.presenter.export_mode == ExportMode.ROI_MODE:
+                if self.roi_table_model.row_data(roi_item)[2] is False:
+                    self.set_roi_alpha(0, roi_item)
+                else:
+                    self.set_roi_alpha(255, roi_item)
+                    self.presenter.redraw_spectrum(self.roi_table_model.row_data(roi_item)[0])
+            else:
+                self.set_roi_alpha(0, roi_item)
+
+        return
```
JackEAllen, line 155: NITPICK, although the efficiency gain is likely negligible, we could reduce the number of calls to `self.roi_table_model.row_data(roi_item)` in the loop by instead placing the row data in a variable (`row_data = self.roi_table_model.row_data(roi_item)`) before the loop to be used within the loop.

JackEAllen (changes requested):
Looks good and toggles ROI visibility when swapping between ROI to Image view. I like the use of an Enum 👍  Only a few minor nitpicks.

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -17,6 +19,12 @@
 LOG = getLogger(__name__)
 
 
+class ExportMode(Enum):
+    # Needs to match GUI tab order
```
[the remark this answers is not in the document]
samtygier-stfc: I don't think its much use to show in the docs. It just there for future developers if another tab is ever added.

mantidimaging/gui/windows/spectrum_viewer/view.py
samtygier-stfc: It still needs to be in the loop, because `roi_item` get incremented. But I can be done once and put in names to improve readability.

JackEAllen: Yes, that's correct, apologies, I didn't word this very well. 
I was thinking something like this:
```Python
def on_visibility_change(self) -> None:
    """
    When the visibility of an ROI is changed, update the visibility of the ROI in the spectrum widget
    """
    for roi_item in range(self.roi_table_model.rowCount()):
        row_data = self.roi_table_model.row_data(roi_item)
        if self.presenter.export_mode == ExportMode.ROI_MODE:
            if row_data[2] is False:
                self.set_roi_alpha(0, roi_item)
            else:
                self.set_roi_alpha(255, roi_item)
                self.presenter.redraw_spectrum(row_data[0])
        else:
            self.set_roi_alpha(0, roi_item)
```

mantidimaging/gui/windows/spectrum_viewer/presenter.py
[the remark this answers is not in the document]
JackEAllen: Happy to leave as is. This is more of a personal preference of mine to have as a docstring as it looks cleaner in my opinion.

JackEAllen (approved):

PR #1946, opened 2023-10-05
Fix loading a subset of a sample when there is a log with angles

### LINKED ISSUE #1940: Failed to load stack: when using a log file and start/stop/step
### Summary

When MI loads a log file it check thats the number of log entries matches the number of images in the stack. If the user uses the the start, stop or step controls in the load dialog, then the number of images will be lower and the load will fail.

### Steps To Reproduce

In Load Dataset, select a sample and sample log.
Confirm that it loads correctly
Repeat, but change the start, stop or step

### Expected Behaviour

The appropriate lines from the log are used

### Current Behaviour

`Failed to load stack. Error: The number of angles does not match the number of images. Num angles 100 and num images 99`

### Context

Current main

### Failure Logs

```
  File "/home/sam/git/mantidimaging/mantidimaging/gui/dialogs/async_task/task.py", line 50, in run
  File "/home/sam/git/mantidimaging/mantidimaging/core/utility/func_call.py", line 12, in call_with_known_parameters
    return func(**ka)
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/main/model.py", line 44, in do_load_dataset
    sample = load(parameters.image_stacks[FILE_TYPES.SAMPLE])
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/main/model.py", line 42, in load
    return loader.load_stack_from_image_params(im_param, progress, dtype=parameters.dtype)
  File "/home/sam/git/mantidimaging/mantidimaging/core/io/loader/loader.py", line 106, in load_stack_from_image_params
    return load(filename_group=image_params.file_group,
  File "/home/sam/git/mantidimaging/mantidimaging/core/io/loader/loader.py", line 148, in load
    image_stack.set_projection_angles(ProjectionAngles(angles[angle_order]))
  File "/home/sam/git/mantidimaging/mantidimaging/core/data/imagestack.py", line 302, in set_projection_angles
    raise RuntimeError("The number of angles does not match the number of images. "
```

samtygier-stfc:
### Issue

Closes #1940

### Description

When a user has set start, stop or increment in the load dataset dialog then only a subset of images are loaded. Previously we then set the angles, with the full list of angles in the log file. With this change the indices are applied to the angles as well.

### Testing 

Added a new system test with a range of cases

### Acceptance Criteria 

Follow test steps on issue

### Documentation

Release notes

### DIFF
--- a/docs/release_notes/next/fix-1940-load-subset-with-angles
+++ b/docs/release_notes/next/fix-1940-load-subset-with-angles
@@ -1,0 +1,1 @@
+#1940 : Fix loading a subset of a sample when there is a log with angles
--- a/mantidimaging/core/io/loader/loader.py
+++ b/mantidimaging/core/io/loader/loader.py
@@ -139,13 +139,15 @@ def load(filename_group: FilenameGroup,
         log_data = load_log(log_file)
         angles = log_data.projection_angles().value
         angle_order = np.argsort(angles)
+        angles = angles[angle_order]
         file_names = [file_names[i] for i in angle_order]
 
     image_stack = img_loader.execute(load_func, file_names, in_format, dtype, indices, progress)
 
     if log_file is not None:
         image_stack.log_file = log_data
-        image_stack.set_projection_angles(ProjectionAngles(angles[angle_order]))
+        angles = angles[indices[0]:indices[1]:indices[2]] if indices else angles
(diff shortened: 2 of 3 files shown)


JackEAllen (approved):
Looks good!

PR #1973, opened 2023-11-03
brightness percentiles set in live viewer images

### LINKED ISSUE #1964: Use percentiles rather than min max for auto brightness levels in live viewer
### Desired Behaviour

When displaying a new image in the live viewer, use the 5%-95% percentiles to set the brightness range rather than the min and max. This would prevent a small number of outlier pixels skewing the levls.

For this to be useful it needs to not be significantly slower that existing min/max method. An approximate algorithm would probably fine.
Hopefully there is a useful method in https://numpy.org/doc/stable/reference/generated/numpy.percentile.html

The code for displaying images in the live viewer is in:
~~mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py~~ `mantidimaging/gui/windows/live_viewer/live_view_widget.py`
and the MIMiniImageView widget (note this is used in other places)
`mantidimaging/gui/widgets/mi_mini_image_view/view.py`

Will also be useful to look at how we use the pyqtgraph `ImageItem`
https://pyqtgraph.readthedocs.io/en/latest/api_reference/graphicsItems/imageitem.html

MikeSullivan7:
### Issue

Closes #1964 

### Description

In the Live Viewer, the brightness of the image is now set using the 5th and 95th percentiles rather than min/max. This removes outliers and makes the brightness of the images consistent and easier to compare as data is read in.
Percentiles can be changed via `self.image.use_brightness_percentiles(image, low, high)` in `mantidimaging/gui/windows/live_viewer/live_view_widget.py` where `low` and `high` are the lower and higher percentiles respectively. The brightness levels are then passed though `self.image.setImage(image, levels=self.image.brightLevels)` in the Live Viewer only put are left as min/max elsewhere.

No slowdown has been noticed due to the percentile calculations.

### Testing 

Data was fed through the Live Viewer to verify the outliers where left out and the image brightness was consistent. This auto levelling is not done in normal operations outside of the Live Viewer.

### Acceptance Criteria 

Check standard tests and check brightness levels are correct in the Live Viewer.

### DIFF
--- a/mantidimaging/gui/widgets/mi_mini_image_view/view.py
+++ b/mantidimaging/gui/widgets/mi_mini_image_view/view.py
@@ -15,8 +15,9 @@ from mantidimaging.gui.utility.qt_helpers import BlockQtSignals
 from mantidimaging.gui.widgets.auto_colour_menu.auto_color_menu import AutoColorMenu
 from mantidimaging.gui.widgets.bad_data_overlay.bad_data_overlay import BadDataOverlay
 
+import numpy as np
+
 if TYPE_CHECKING:
-    import numpy as np
     from PyQt5.QtWidgets import QWidget
 
 graveyard = []
@@ -32,6 +33,9 @@ def pairwise(iterable):
 
 
 class MIMiniImageView(GraphicsLayout, BadDataOverlay, AutoColorMenu):
+    brightLevels: None | list = None
+    lowerLevel: float
+    higherLevel: float
 
     def __init__(self, name: str = "MIMiniImageView", parent: 'Optional[QWidget]' = None, recon_mode: bool = False):
         super().__init__()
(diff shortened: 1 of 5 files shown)


mantidimaging/gui/widgets/mi_mini_image_view/view.py
```diff
@@ -166,3 +170,8 @@ def update_sibling_histograms(self):
         for img_view in self.histogram_siblings:
             with BlockQtSignals(img_view.hist):
                 img_view.hist.setLevels(*hist_range)
+
+    def use_brightness_percentiles(self, image: np.ndarray, percent_low: int, percent_high: int, *args, **kwargs):
```
samtygier-stfc, line 174: We don't need the `*args, **kwargs` part here, as its not being passed through in the this method.

mantidimaging/gui/windows/live_viewer/live_view_widget.py
```diff
@@ -35,7 +35,8 @@ def show_image(self, image: np.ndarray) -> None:
         Show the image in the image view.
         @param image: The image to show
         """
-        self.image.setImage(image)
+        self.image.use_brightness_percentiles(image, 5, 95)
+        self.image.setImage(image, levels=self.image.brightLevels)
```
samtygier-stfc, line 39: Can `MIMiniImageView.setImage()` do a check to see if the percentiles are being used, so that levels does not need to be passed in from outside. So this part would be 

```
self.image.use_brightness_percentiles(image, 5, 95)
self.image.setImage(image)
```

samtygier-stfc (changes requested):
Looking good. A couple of comments.

* Could you add a test

samtygier-stfc (approved):
Thanks. Looks good.

Issue #1965, opened 2023-11-01
Improve developer exerience on windows


samtygier-stfc:
### Desired Behaviour

`make` should be installed as a dev tool

PR #1976, opened 2023-11-16
-n auto added to make test in Makefile

MikeSullivan7:
### Issue

Continues #1965 

### Description

Added "-n auto" to the "make test" command in the Makefile. This reduces the time to run pytest on windows by a factor of ~5-10,

### Testing 

Check "make test" in Windows and Linux

### Acceptance Criteria 

See if the make test still runs in an acceptable amount of time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ install-dev-requirements:
 	python ./setup.py create_dev_env
 
 test:
-	python -m pytest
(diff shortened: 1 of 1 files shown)


samtygier-stfc (approved):
Thanks. Works fine on Linux. We have also been using this flag on github actions, so should be safe.

Issue #1968, opened 2023-11-02
Windows GUI layout problems on ultrawide monitor


MikeSullivan7:
### Summary

When using Mantid Imaging on Windows on an ultrawide monitor (2560 x 1080), the GUI layout does not display correctly.

### Steps To Reproduce

Check how Qt widget sizing dynamically changes depending on monitor resolution and make the layout resolution-independent.

### Expected Behaviour

GUI should be consistent across different monitor resolutions

### Current Behaviour

Some parts of the GUI are not in the correct position or size when displayed on an ultrawide monitor e.g. sliders, scale, buttons.

### Context

main branch

### Screenshot(s)

Standard monitor:
![image](https://github.com/mantidproject/mantidimaging/assets/30868085/53a06377-6600-433c-a913-deedc8254edf)

Ultrawide monitor:
![image](https://github.com/mantidproject/mantidimaging/assets/30868085/9f6096ac-373d-4400-beae-8b090d4c4274)

samtygier-stfc:
Symptoms look a lot like https://github.com/pyqtgraph/pyqtgraph/issues/1250 , but that should have been fixed in pyqtgraph. Can you check which version you have installed `mamba list pyqtgraph`. Otherwise that issue is probably a good place to start researching.

MikeSullivan7:
I'm running pyqtgraph 0.13.3 which is the latest version but still having issues. This will take more investigation...

JackEAllen:
May be worth renaming this to Windows GUI resolution layout problems as MI looks fine on windows on my ultra-wide with scaled down resolution.

PR #1978, opened 2023-11-20
Added environment settings and allowed variable scaling

### LINKED ISSUE #1975: Reconstruction window does not display correctly on 4K monitor
### Summary

In the recon window, the buttons, tables, etc do not display correctly and the text on the buttons are not readable. These issues are not present in the other windows: Operations, Live Viewer, etc.

### Steps To Reproduce

Use Mantid Imaging on Windows on a 4k monitor or nonstandard resolution.

### Expected Behaviour

The reconstruction window should display correctly regardless of what the resolution of the users screen is.

### Current Behaviour

Buttons and entry widgets in the reconstruction windows are squashed and cannot be read or seen easily.

### Context

Main branch on windows

### Screenshot(s)

image

MikeSullivan7:
### Issue

closes #1975 and #1968 

### Description

During setup, environment variables are set to allow high dpi scaling and auto screen scaling. The font is also fixed, which then seems to scale well when changing monitors. The odd scaling on ultrawide monitors also now displays correctly.

### Testing 

Open on a HD and 4K monitor on Windows and Linux. If using multiple monitors with different resolutions, try changing your "main" or "primary" monitor in windows settings to see if the GUI still scales correctly.

### Acceptance Criteria 

Check that everything is readable when changing monitor resolutions/dragging across monitors.

### DIFF
--- a/mantidimaging/main.py
+++ b/mantidimaging/main.py
@@ -55,9 +55,6 @@ def parse_args() -> argparse.Namespace:
 
 
 def setup_application() -> QApplication:
-    os.environ["QT_ENABLE_HIGHDPI_SCALING"] = "1"
-    os.environ["QT_AUTO_SCREEN_SCALE_FACTOR"] = "1"
-    os.environ["QT_SCALE_FACTOR"] = "1"
     QGuiApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
     q_application = QApplication(sys.argv)
     q_application.setStyle('Fusion')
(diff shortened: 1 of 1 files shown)


MikeSullivan7:
Ah yes, I can fix it by setting the attribute in QGuiApplication and it still works, I'll make the changes

MikeSullivan7:
Now added a bit which makes the font changes only on windows and also a scaling factor which slightly adjusts the font size when using a higher dpi monitor to make the text more readable. An assumption is made that a "standard" monitor has a dpi of around 90 but this can be tweaked.

MikeSullivan7:
All scaling and font changes are now Windows only

samtygier-stfc (approved):
Looks good. Solves issues on windows.

Issue #1951, opened 2023-10-09
Pytest on Windows is slow


samtygier-stfc:
### Summary

Running pytest on windows takes over 15 minutes on my laptop. Compare to under 100 seconds on Linux.

On github CI servers it is 3:30 vs 1 minute. So some of the slowness might be anti-malware processes running locally.

This needs some investigation. The slowest tests seem to be in the views.

```
214.47s teardown mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_on_auto_update_triggered_with_auto_selected
102.47s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_on_auto_update_triggered_with_auto_not_selected
57.59s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_on_auto_update_triggered_with_auto_selected
50.53s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_lock_zoom_changed_selected
49.13s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_lock_scale_changed_deselected
43.08s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_lock_zoom_changed_deselected
31.78s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_lock_scale_changed_selected_flat_fielding
26.54s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_lock_scale_changed_selected_not_flat_fielding
24.57s call     mantidimaging/gui/windows/recon/test/view_test.py::ReconstructWindowViewTest::test_update_recon_preview_and_hist
22.26s call     mantidimaging/gui/windows/operations/test/view_test.py::OperationsWindowsViewTest::test_collapse
```

### Steps To Reproduce

`python -m pytest --durations=10`

### Expected Behaviour

Tests on windows take a similar time to on Linux

### Current Behaviour
Windows much slower

### Context

On physical hardware

MikeSullivan7:
Timing of pytest on my laptop is inconsistent as to what exactly is taking the most time to run. 
Two consecutive runs of `python -m pytest --durations=10` give the following results:
```
======================================= slowest 10 durations ========================================
[...]

Issue #902, opened 2021-03-26
Investigate if fix in #894 still needed


samtygier-stfc:
### Summary

#894 prevented a segfault by preventing MIImageView from ever being garbage collected. A future Qt or Pyqtgraph update might resolve the issue
 
### Steps To Reproduce

Remove the `graveyard.append(self)` line and see if a segfault happens while running through a standard work flow.

PR #1979, opened 2023-11-20
Cleanup image views

samtygier-stfc:
### Issue
Work on #1951

### Description

Some changes to remove some reference cycles in the `MIMiniImageView` and classes that use them. The aim is reduce the number of objects that can't be delete, which may be slowing down tests. Its also possible that some of the tangle is responsible for #902 .

Avoid a lambdas that hold a reference to `self` in `auto_colour_menu` and `im.hoverEvent`.

Delete some references to the image views in the operations and recon windows

Better mocking in `ReconstructWindowViewTest`. Previously the `ReconImagesView` was created and then hidden with a mock. Now mock the creation.

On Linux this reduces single threaded pytest runtime from about 2:00 to 1:30. Also it eliminates about a 1 minute wait from pytest completing and output its results to returning to the prompt.

Checkout the commit `Add leak tracking fixture to conftest` which has the diagnostics but not the fixes
`git checkout b1b2166ea8483767f8301b9ef0e239f46d1db0f4`
In `conftest.py` set `autouse = True`  for `leak_test_stats`
In the constructor for `MIMiniImageView` add
`leak_tracker.add(self, msg=name)`
You will need the import `from mantidimaging.core.utility.leak_tracker import leak_tracker`
This will print the number of tracked objects that have not been garbage collected. You will see something like:
```
Leaked item stats:
<class 'numpy.ndarray'>: 36
<class 'mantidimaging.core.parallel.utility.SharedArray'>: 10
<class 'mantidimaging.gui.widgets.mi_mini_image_view.view.MIMiniImageView'>: 211
```
With the same changes on the head of this branch, there should be no leaked `MIMiniImageView` objects.

### Documentation

release notes

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -2,7 +2,11 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 from __future__ import annotations
 
+from collections import Counter
+
 import pytest
+
+from mantidimaging.core.utility.leak_tracker import leak_tracker
 
 
 def _test_gui_system_filename_match(basename: str) -> bool:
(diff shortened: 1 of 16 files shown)


MikeSullivan7 (approved):
Everything looks good! 

Tested without the fix with the following result:
image

Tested with the fix (ignore the execution_timer_test error):
image
image

Now without multithreading, pytest in Windows takes around 3 minutes which is on par if not faster than with multithreading (as setting up the threads can take a bit of time).

PR #1980, opened 2023-11-21
Fix pytest running with -vs

samtygier-stfc:
### Issue
Quick fix for issue encounter while working on #1951

### Description

Mock out save_roi_coords. The path calculation for the coords file goes odd when the main path is a StringIO

Allow the wrapped StringIO to be closed, capturing the output makes tests better

Fix the class name

Run verbose tests, previously this failed with an error `OSError: [Errno 9] Bad file descriptor`. Now it should pass.
```make test-verbose```
or 
```python -m pytest -vs -o log_cli=true```

(Note some tests do log error s that a visible in verbose mode)

### Documentation

Not needed


MikeSullivan7 (approved):
Looks good, no failures when running test-verbose. All errors are those that are intentional from the unit tests.

Issue #1961, opened 2023-10-27
Update sphinx


samtygier-stfc:
### Desired Behaviour

We are running several versions behind.

Check that docs still build, as there is potential for some breaking changes.

PR #1983, opened 2023-11-27
Updated sphinx to 6.2

MikeSullivan7:
### Issue

Progresses #1961 

### Description

sphinx version has been changed to sphinx==6.2.* in environment-dev.yml.

Note: upgrading to the latest version of 7 of Sphinx causes the documentation building in the setup.py file to break. This is because Sphinx has stopping supporting `BuildDoc ` in `from sphinx.setup_command import BuildDoc` and have removed it.

### Testing 

Check that Documentation still builds


samtygier-stfc (approved):
Looks good.

PR #1989, opened 2023-12-06
main_window no longer passed to BaseMainWindowView

### LINKED ISSUE #1429: Highlighting main window doesn't bring it to front on Windows
### Desired Behaviour

When other windows are open (i.e. operations and/or reconstruction), clicking the main window should activate it and bring it to the front, as it does on Linux. We should confirm first if this is the way other Windows application behave, though, as it may not be desirable to change it if it is.

### Current Behaviour

When the operations and/or reconstruction windows are open, they must either be minimised or closed to be able to bring the main window to the front. Clicking on the main window seems to activate it and you can take actions within it, however it continues to sit behind any other opened MI windows. In contrast, the operations and reconstruction windows will activate and move to the front if they are clicked when there are multiple MI windows open.

MikeSullivan7:
### Issue

Closes #1429.

### Description

In the Views for the additional windows (operations, spectrum viewer, reconstruct, etc), `self.main_window` is no longer passed to the `BaseMainWindowView` through `super().__init__()` but is still passed to the view for access. This stops the parent/child relationship between the two QMainWindows.
A side effect of this is that the additional windows do not close with the MainWindow is this is now done manually in the Main views `closeEvent()`.

### Acceptance Criteria 

Open branch in Windows, load in data and then open the Operations, Live Viewer, Spectrum Viewer, and Reconstruction windows and check that the Main window can now sit on top when clicked. All interactions between the additional windows and the main windows should work the same as before.

Check that the program still behaves as it did before in Linux.

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -26,7 +26,7 @@ class LiveViewerWindowView(BaseMainWindowView):
     imageLayout: QVBoxLayout
 
     def __init__(self, main_window: 'MainWindowView', live_dir_path: Path) -> None:
-        super().__init__(main_window, 'gui/ui/live_viewer_window.ui')
+        super().__init__(None, 'gui/ui/live_viewer_window.ui')
         self.setWindowTitle("Mantid Imaging - Live Viewer")
         self.main_window = main_window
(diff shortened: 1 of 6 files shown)


samtygier-stfc (approved):
Looks good. Tested on Linux and Windows

Issue #1992, opened 2023-12-07
Spectrum Viewer: Change ROI Colour Through Right Click Dialog Menu


JackEAllen:
### Desired Behaviour

From the image view, spectrum line plot or table of ROIs, provide a right click dialog menu allowing a user to set a new colour for a given ROI. Only colours provided by `mantidimaging.gui.windows.spectrum_viewer.spectrum_widget.colour_generator` should be made available. 

### Current Behaviour
It it not possible to change an ROIs colour once added. 

The current work around for this is to delete the ROI and add a replacement ROI which will be given the next colour in the list from the previous colour. 
The list of colours can be found here: `mantidimaging.gui.windows.spectrum_viewer.spectrum_widget.colour_generator`
The list tries to use colours that are visible for users who may have varying levels of colour blindness.

samtygier-stfc:
I think this should also work by clicking the colour on the ROI table.

We could show a QColorDialog and let the user pick any colour.

PR #1993, opened 2023-12-07
Fix ROI Colour Clash

JackEAllen:
### Issue

Avoids ROI colour clashes if one ROI is added after another. Previously the first two ROI colours were hard to distinguish between each other.

### Description

* Refactor ROI colour generator method name to be more descriptive
* Moved second ROI colour (12, 123, 220) to another location in array where it does not clash with other ROI colours

### Testing 

* manually tested adding ROIs and check they were distinguishable from each other in image view and spectrum line plot.
* Executed unit tests. 

### Acceptance Criteria 

Check that ROIs added are distinguishable from one another (in most cases, usually 2-3 ROIs are added) in the image view and spectrum line plot when moved around.

### Documentation

N/A - Update documentation as part of #1992 which will enable right clicking of an ROI to change the colour of the ROI.

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
@@ -114,7 +114,7 @@ class SpectrumWidget(GraphicsLayoutWidget):
         r_min, r_max = self.range_control.getRegion()
         return int(r_min), int(r_max)
 
-    def random_colour_generator(self) -> tuple[int, int, int, int]:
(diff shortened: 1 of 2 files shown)


MikeSullivan7 (approved):
Looks good to me, ROIs look distinguishable to me in both the image view and the spectrum. 
image

Only improvement I can think of is maybe increasing the thickness of the ROI boxes in the  imageview as I could imagine some of the lighter colours getting lost in busy images. That is only personal preference though and I'm happy to approve as is.

PR #1996, opened 2023-12-11
added fits compatibility for live viewer

### LINKED ISSUE #1995: LIVE VIEWER: FITS Compatibility
### Desired Behaviour

The Live Viewer should be able to load .fits files as it does .tif and .tiff files.

### Current Behaviour

Currently, the Live Viewer only accepts .tif and .tiff files

MikeSullivan7:
### Issue

Closes #1995 

### Description

.fits files can now be loaded into the Live Viewer

### Testing 

The brass dataset at (\\isis\shares\IMAT\ExampleData\Brass\Corrected_Sample_PH20) was loaded into the Live Viewer and displayed correctly.

### Acceptance Criteria 

Check that .fits files can be loaded into the live viewer either directly or by simulating live data.

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/model.py
+++ b/mantidimaging/gui/windows/live_viewer/model.py
@@ -259,7 +259,7 @@ class ImageWatcher(QObject):
         :param file_name: name of file
         :return: True if file is an image file
         """
-        image_extensions = ['.tif', '.tiff']
+        image_extensions = ['.tif', '.tiff', '.fits']
         file_names = any(file_name.lower().endswith(ext) for ext in image_extensions)
         return file_names
 
--- a/mantidimaging/gui/windows/live_viewer/model.py
+++ b/mantidimaging/gui/windows/live_viewer/model.py
@@ -259,9 +259,8 @@ class ImageWatcher(QObject):
         :param file_name: name of file
         :return: True if file is an image file
         """
-        image_extensions = ['.tif', '.tiff', '.fits']
-        file_names = any(file_name.lower().endswith(ext) for ext in image_extensions)
-        return file_names
+        image_extensions = ('tif', 'tiff', 'fits')
+        return file_name.rpartition(".")[2].lower() in image_extensions
 
     def remove_path(self):
         """
--- a/mantidimaging/gui/windows/live_viewer/presenter.py
+++ b/mantidimaging/gui/windows/live_viewer/presenter.py
@@ -78,10 +78,10 @@ class LiveViewerWindowPresenter(BasePresenter):
 
     def load_and_display_image(self, image_path: Path):
         try:
-            if image_path.__str__().lower().endswith('.tif') or image_path.__str__().lower().endswith('.tiff'):
+            if image_path.suffix.lower() in [".tif", ".tiff"]:
                 with tifffile.TiffFile(image_path) as tif:
                     image_data = tif.asarray()
-            elif image_path.__str__().lower().endswith('.fits'):
+            elif image_path.suffix.lower() == ".fits":
                 with fits.open(image_path.__str__()) as fit:
                     image_data = fit[0].data
         except (IOError, KeyError, ValueError, TiffFileError, DeflateError) as error:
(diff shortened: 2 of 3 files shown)


mantidimaging/gui/windows/live_viewer/presenter.py
```diff
@@ -77,8 +78,12 @@ def select_image(self, index: int) -> None:
...
     def load_and_display_image(self, image_path: Path):
         try:
-            with tifffile.TiffFile(image_path) as tif:
-                image_data = tif.asarray()
+            if image_path.__str__().lower().endswith('.tif') or image_path.__str__().lower().endswith('.tiff'):
+                with tifffile.TiffFile(image_path) as tif:
+                    image_data = tif.asarray()
+            elif image_path.__str__().lower().endswith('.fits'):
+                with fits.open(image_path.__str__()) as fit:
+                    image_data = fit[0].data
```
JackEAllen, line 86: `image_path` is of type `Path`, so we can make use of the `suffix` property to get the extension of a file.
I would recommend using the `suffix` property to simplify the if else logic cases like this:
```suggestion
            if image_path.suffix.lower() in [".tif", ".tiff"]:
                with tifffile.TiffFile(image_path) as tif:
                    image_data = tif.asarray()
            elif image_path.suffix.lower() == ".fits":
                with fits.open(image_path.__str__()) as fit:
                    image_data = fit[0].data
```

JackEAllen (changes requested):
Looks good! I can see `.fits` files within the live viewer. Just a small suggestion to make use of the `suffix` property available through `image_path` and happy to approve. 👍

JackEAllen:
Could you please also add release notes for your changes here `docs/release_notes/next` as part of this PR? 
You can copy one of the existing files and rename to something like `feature-1996-fits-compatibility` and inside it, add the PR number and brief description of changes following the same format as the other release notes.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -259,7 +259,7 @@ def _is_image_file(file_name: str) -> bool:
         :param file_name: name of file
         :return: True if file is an image file
         """
-        image_extensions = ['.tif', '.tiff']
+        image_extensions = ['.tif', '.tiff', '.fits']
         file_names = any(file_name.lower().endswith(ext) for ext in image_extensions)
```
samtygier-stfc, line 263: A micro optimisation
```
image_extensions = ('tif', 'tiff', 'fits')
return file_name.rpartition(".")[2].lower() in image_extensions
```
would avoid calling lower and ends with multiple times.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -259,9 +259,8 @@ def _is_image_file(file_name: str) -> bool:
         :param file_name: name of file
         :return: True if file is an image file
         """
-        image_extensions = ['.tif', '.tiff']
-        file_names = any(file_name.lower().endswith(ext) for ext in image_extensions)
-        return file_names
+        image_extensions = ('tif', 'tiff', 'fits')
+        return file_name.rpartition(".")[2].lower() in image_extensions
```
JackEAllen, line 263: Just need to remove `.lower()` from here to meet Sam's micro optimisation suggestion now we are using `.lower` in `presenter.load_and_display_image` and this is good to go 👍

JackEAllen (changes requested):

mantidimaging/gui/windows/live_viewer/model.py
samtygier-stfc: This is called from `find_images()` which is passing the original names, so it does need the `lower()`

samtygier-stfc: Though as its only called there, I would be tempted to inline it into line 188 as
`if file_path.suffix.lower() in (".tif", ".tiff", ".fits")`
and remove the `_is_image_file()` method.

samtygier-stfc (approved):

JackEAllen (approved):

PR #1998, opened 2023-12-14
Tiff Unpacker Utility Script

JackEAllen:
### Issue

Currently it is not possible to unpack a tiff container directly into MI I have create the issue #1990 which will be later absorbed into an EPIC to refactor how MI loads data.

IMAT are using a new camera which produces TIFF containers therefore there is a need to be able to make these compatible with mantid imaging.

As a temporary work around, I have created a quick utility script to allow users to unpack tiffs into a directory.

### Testing 

Manually tested the script using argument flags in different combinations.

### Acceptance Criteria 

* Script can be used to unpack a tiff container into a new directory.
* A user can select a range in which to copy images within the stack.
* A user can set the naming convention for the images unpacked.
* unpacked tiff container can be loaded into MI and viewed within the Spectrum Viewer.

### Documentation

NA as this is a temporary script which should be removed once MI's loading has been refactored to handle tiff containers.

### DIFF
--- a/scripts/tiff_unpacker.py
+++ b/scripts/tiff_unpacker.py
@@ -1,0 +1,125 @@
+# Copyright (C) 2023 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+"""
+This is a Utility script to unpack tiff files into a directory of images using lossless compression.
+This script is temporary while larger work is undertaken to improve how datasets are stored and accessed within Mantid.
+
+Usage:
+    python tiff_unpacker.py -s <path to tiff file> -d <path to output directory>
+    python tiff_unpacker.py
+        -s <path to tiff file>
+        -d <path to output directory>
+        -r <start> <stop>
+        -n <naming convention>
+"""
+import argparse
+from tqdm import tqdm
+from pathlib import Path
+from tifffile import TiffFile, imwrite
+from logging import getLogger, basicConfig, INFO
+
+basicConfig(format='%(asctime)s %(message)s', datefmt='%m/%d/%Y %I:%M:%S %p', level=INFO)
+logger = getLogger(__name__)
+
+
+def unpack_tiff(source: Path, destination: Path, img_range: tuple[int, int], naming_convention: str) -> None:
+    """
+    Unpack a tiff file into a directory of LZW compressed tif images
+
+    Parameters:
+        source: Path to tiff file
+        destination: Path to output directory
+
+    Returns:
+        None
+    """
+    with TiffFile(source) as tif:
+        for i, img in enumerate(tqdm(tif.pages[img_range[0]:img_range[1]], desc="Unpacking tiff file")):
+            imwrite(destination / f"{naming_convention}{i}.tiff",
+                    img.asarray(),
+                    compression="zlib",
+                    predictor=True,
+                    metadata={
+                        'axes': img.axes,
+                    })
+
+
+def validate_input_range(data_length: int, start: int, stop: int) -> tuple[int, int]:
+    """
+    Validate the input range
+
+    If start is greater than stop then raise ValueError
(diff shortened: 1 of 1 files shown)


scripts/tiff_unpacker.py
```diff
@@ -0,0 +1,126 @@
...
+    with TiffFile(source) as tif:
+        for i, img in enumerate(tqdm(tif.pages[img_range[0]:img_range[1]], desc="Unpacking tiff file")):
+            imwrite(destination / f"{naming_convention}{i}.tiff",
+                    img.asarray(),
+                    compression="zlib",
+                    predictor=True,
+                    metadata={
+                        'axes': 'TZCYX',
+                        'unit': 'um'
+                    })
```
samtygier-stfc, line 45: I think this is claiming that we are writing 5D data, but we should have 2D "YX" data here. We could use the axis from the page.
`'axes': img.axes`
I'm not sure where the `unit` is coming from

scripts/tiff_unpacker.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def validate_input_range(data_length: int, start: int, stop: int) -> tuple[int, int]:
+    """
+    Validate the input range
+
+    If start is greater than stop then raise ValueError
+    If stop is greater than the number of images in the tiff file then raise ValueError
```
samtygier-stfc, line 53: Comment here does not match behaviour

scripts/tiff_unpacker.py
```diff
@@ -0,0 +1,126 @@
...
+    Parameters:
+        source: Path to tiff file
+        destination: Path to output directory
+
+    Returns:
+        None
+
+    Raises:
+        ValueError: If source is not a file
+        ValueError: If destination is not a directory
```
samtygier-stfc, line 76: Comment here does not match behaviour

scripts/tiff_unpacker.py
```diff
@@ -0,0 +1,126 @@
+# Copyright (C) 2023 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+"""
+This is a Utility script to unpack tiff files into a directory of images using lossless compression.
+This script is temporary while larger work is undertaken to improve how datasets are stored and accessed within Mantid.
+
+Usage:
+    python tiff_unpacker.py <path to tiff file> <path to output directory>
```
samtygier-stfc, line 8: `python tiff_unpacker.py -s <path to tiff file> -d <path to output directory>`

scripts/tiff_unpacker.py
```diff
@@ -0,0 +1,126 @@
...
+def input_parser():
+    """
+    Input parser for the script
+
+    Expected input:
+        python tiff_unpacker.py <path to tiff file> <path to output directory>
+
+    Returns:
+        Path to tiff file
+        Path to output directory
```
samtygier-stfc, line 94: Also range and name

samtygier-stfc (changes requested):

scripts/tiff_unpacker.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    with TiffFile(source) as tif:
+        for i, img in enumerate(tqdm(tif.pages[img_range[0]:img_range[1]], desc="Unpacking tiff file")):
+            imwrite(destination / f"{naming_convention}{i}.tiff",
+                    img.asarray(),
+                    compression="zlib",
+                    predictor=True,
+                    metadata={
+                        'axes': img.axes,
+                        'unit': 'um'
```
samtygier-stfc, line 44: I'm still unsure about adding the unit to the metadata without having a source for it. Can we remove this line.

JackEAllen: Removed and added shape to metadata for ref

samtygier-stfc (approved):
Thanks

PR #2015, opened 2024-01-22
Change ROI Colour

### LINKED ISSUE #1992: Spectrum Viewer: Change ROI Colour Through Right Click Dialog Menu

ashmeigh:
### Issue

Closes #1992

### Description

This pull request addresses the issue reported in #1992. This pull request introduces an enhancement to the Spectrum Viewer, targeting the functionality around ROIs. Previously, users could not easily change the colour of an ROI within the viewer. With this update, I have implemented a right-click context menu for ROIs, providing users with an interface to select new colours. This enhancement improves user interaction with the Spectrum Viewer

### Testing 

### Acceptance Criteria 

Right-clicked on various ROIs within the Spectrum Viewer to ensure the context menu appears consistently.
Selected different colours from the colour dialog and verified that the selected ROI updated to the new colour immediately.
Tested with multiple ROIs to ensure that changing the colour of one ROI does not affect others.

### Documentation

Noted the changes in the appropriate file in docs/release_notes.

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
@@ -7,7 +7,7 @@ from typing import TYPE_CHECKING, Optional
 from PyQt5.QtCore import pyqtSignal, Qt, QSignalBlocker
 from PyQt5 import QtGui, QtWidgets
 from pyqtgraph import ROI, GraphicsLayoutWidget, LinearRegionItem, PlotItem, mkPen
-from PyQt5.QtGui import QPen
+
 from mantidimaging.core.utility.close_enough_point import CloseEnoughPoint
 from mantidimaging.core.utility.sensible_roi import SensibleROI
 from mantidimaging.gui.widgets.mi_mini_image_view.view import MIMiniImageView
@@ -53,8 +53,6 @@ class SpectrumROI(ROI):
                 new_color = (selected_color.red(), selected_color.green(), selected_color.blue(), 255)
                 self.colour = new_color  # Update the ROI color
                 self.sig_colour_change.emit(self._name, new_color)
-
-
 
     @property
     def name(self) -> str:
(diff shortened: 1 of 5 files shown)


ashmeigh:
roi_screen

samtygier-stfc (changes requested):

mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
```diff
@@ -38,6 +41,20 @@ def __init__(self, name: str, sensible_roi: SensibleROI, *args, **kwargs):
         self.addScaleHandle([0, 1], [1, 0])
         self._selected_row = None
 
+        menu = self.getMenu()
+        change_color_action = QAction("Change ROI Colour", self)
+        change_color_action.triggered.connect(self.onChangeColor)
+        menu.addAction(change_color_action)
+        menu.exec_(QCursor.pos())
```
samtygier-stfc, line 48: This should work, but looking into the PyQtGraph [ROI source](https://pyqtgraph.readthedocs.io/en/latest/_modules/pyqtgraph/graphicsItems/ROI.html), it seems that they have tied the logic for displaying the menu to whether the ROI is `removable` from the context menu. So it will have to be done in a more hacky way until PyQtGraph is fixed.

So in the `__init__()` replace those 5 lines with
```python
        self.menu = QtWidgets.QMenu()
        change_color_action = QAction("Change ROI Colour", self)
        change_color_action.triggered.connect(self.onChangeColor)
        self.menu.addAction(change_color_action)
```
Then we need to add a method to prevent the context menu being disabled
```python
    def contextMenuEnabled(self):
        return True
```

samtygier-stfc: (I have reported this limitation to pyqtgraph https://github.com/pyqtgraph/pyqtgraph/issues/2931 )

samtygier-stfc (approved):
Looks good. Thanks.

PR #2016, opened 2024-01-23
Use "Rotate Image" operation in the Live Viewer

### LINKED ISSUE #2003: Live Viewer: Rotate Images
### Desired Behaviour

Some cameras are mounted at different angles meaning arriving data is not always displayed at the correct angle.

Customise right click menu of image within Live Viewer to rotate all images within by 90, 180 or 270. 
For a given selected angle, all existing images and newly arriving images should be rotated to the new angle so a user does not need to keep re-selecting a new angle for every new image which arrives.

I recommend placing this in a right menu over the image to minimise UI changes to the Live viewer itself. 

Users like the Live Viewer window to be minimal to leave on a monitor maximised.

To tackle this issue, use existing rotate stack code from operations window:
- `mantidimaging/core/operations/rotate_stack/rotate_stack.py`

This issue will act as a good test for moving existing operations to work with live/changing data stacks which currently, they are not.

### Current Behaviour

Live data can not be rotated.

image

MikeSullivan7:
### Issue

Closes #2003 

### Description

The images in the Live Viewer can now be rotated by 90, 180, or 270 degrees via a right-click menu option. Once selected, this is apply to all images shown in the Live Viewer so the option only needs to be chosen once.
This functionality uses the existing framework in `mantidimaging.core.operations.rotate_stack.rotate_stack.py` successfully suggesting that other operations could be implemented in the Live Viewer in the future.

image

image

### Testing 

`make check` runs successfully but unit tests will be added in the future. Currently no testing exists for the Live Viewer window.

### Acceptance Criteria 

Load images into the Live Viewer and check that the data is rotated as expected.

### Documentation

Will be added in release notes

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from PyQt5.QtCore import QSignalBlocker, QCoreApplication
+from PyQt5.QtCore import QSignalBlocker
 from PyQt5.QtWidgets import QVBoxLayout
 from PyQt5.Qt import QAction, QActionGroup
 
@@ -16,8 +16,6 @@ import numpy as np
 
 if TYPE_CHECKING:
     from mantidimaging.gui.windows.main import MainWindowView  # noqa:F401  # pragma: no cover
-
-translate = QCoreApplication.translate
 
 
 class LiveViewerWindowView(BaseMainWindowView):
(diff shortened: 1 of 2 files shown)


mantidimaging/gui/windows/live_viewer/view.py
```diff
@@ -34,6 +37,23 @@ def __init__(self, main_window: 'MainWindowView', live_dir_path: Path) -> None:
         self.live_viewer = LiveViewWidget()
         self.imageLayout.addWidget(self.live_viewer)
         self.live_viewer.z_slider.valueChanged.connect(self.presenter.select_image)
+        self.image_rotation_angle = 0
+        self.right_click_menu = self.live_viewer.image.vb.menu
+        rotate_menu = self.right_click_menu.addMenu(translate("ViewBox", "Rotate Image"))
+        rotate_angles_group = QActionGroup(self)
+        rotate_0 = QAction(translate("ViewBox", "0" + u'\N{DEGREE SIGN}'), rotate_angles_group)
```
samtygier-stfc, line 44: Given that we are not using translation anywhere else, I don't think we should start here. (Whether we should make everything translatable is an interesting question).

Also, in python3 it should be fine using the degree symbol in a string.

`rotate_0 = QAction(translate("0°"), rotate_angles_group)`

MikeSullivan7: Fair enough, I used it as that's how its structured in the ViewBox module so kept it consistent with that. If we start getting a non-english user base then that would be a good question to ask!

samtygier-stfc:
Superseded by #2021

Issue #2018, opened 2024-01-24
Live Viewer Operations Pipeline


MikeSullivan7:
### Desired Behaviour

Create pipeline-like behaviour in the Live Viewer which Operations can be added. We want a generalised structure which can pull from the existing Operations structure and carry out the wanted operations on the image data as it is loaded into the Live Viewer.
Not all operations need to be added into the Live Viewer at this current time but there should be the ability to add needed operations when required.

### Current Behaviour

The "Rotate Image" operation can be used in the Live Viewer via a right click menu but the back-end is specific to the "Rotate Image" operation and not generalised.

PR #2021, opened 2024-01-26
Live viewer operations pipeline

MikeSullivan7:
### Issue

Closes issue #2018

### Description

The process for adding and activating operations to be used in the Live Viewer is now generalised into a pipeline.

The operations you want to use can be activated via a right click menu and the appropriate settings are used with the existing Operations framework.

image

image

Right now, the Rotate Stack operation can be used via the right click menu "Operations".
An instance of a Gaussian filter can also be activated in order to test the generalisation and ease of adding operations to the pipeline as needed.
The Gaussian filter can be deleted before merging as it may not have much use in real operations.

Once activated, the operations can be deactivated via "0" or "Off" options to prevent slow down if a large number of operations are used concurrently.

Screenshot of both Gaussian and Rotate Stack operations in use: 

image

### Testing 

make check

### Acceptance Criteria 

Check that the Rotate Stack and Gaussian filters can be used as expected in the Live Viewer. They should be able to be used independently and simultaneously with no notable slow down

### Documentation

Documentation of this feature will be added in the release notes.

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from PyQt5.QtCore import QSignalBlocker, QCoreApplication
+from PyQt5.QtCore import QSignalBlocker
 from PyQt5.QtWidgets import QVBoxLayout
 from PyQt5.Qt import QAction, QActionGroup
 
@@ -16,8 +16,6 @@ import numpy as np
 
 if TYPE_CHECKING:
     from mantidimaging.gui.windows.main import MainWindowView  # noqa:F401  # pragma: no cover
-
-translate = QCoreApplication.translate
 
 
 class LiveViewerWindowView(BaseMainWindowView):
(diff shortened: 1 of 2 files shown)


mantidimaging/gui/windows/live_viewer/view.py
```diff
@@ -74,3 +102,25 @@ def closeEvent(self, e) -> None:
         self.live_viewer.handle_deleted()
         super().closeEvent(e)
         self.presenter = None  # type: ignore # View instance to be destroyed -type can be inconsistent
+
+    def set_image_rotation_angle(self):
+        """Set the image rotation angle which will be read in by the presenter"""
+        if self.rotate_angles_group.checkedAction().text() == "0°":
+            pass
```
samtygier-stfc, line 109: `pass` here and below can be removed.

mantidimaging/gui/windows/live_viewer/view.py
```diff
@@ -35,6 +36,33 @@ def __init__(self, main_window: 'MainWindowView', live_dir_path: Path) -> None:
         self.imageLayout.addWidget(self.live_viewer)
         self.live_viewer.z_slider.valueChanged.connect(self.presenter.select_image)
 
+        self.activated_operations: Set[str] = set()
+        self.filter_params: Dict[str, Dict] = {}
```
samtygier-stfc, line 40: Can these be combined into just a dictionary. So when a filter is turned off we removed the entry in the dictionary.
```
del self.filter_params["Rotate Stack"]
```

samtygier-stfc:
Looks good. If you can remove the Gaussian bit, then I'll merge it.

samtygier-stfc (approved):
Looks good.

PR #2029, opened 2024-02-05
Fix TypeError: qWait(ms: int): unexpected type 'float' to 'int'

### LINKED ISSUE #1873: TypeError: qWait(ms: int): argument 1 has unexpected type 'float'
### Summary

Calls to qWait() should be int. We have some places where it might not be:

  File "/home/sam/git/mantidimaging/mantidimaging/gui/test/gui_system_base.py", line 98, in _wait_for_widget_visible
    QTest.qWait(delay * 1000

### Steps To Reproduce

The problem only occurs when one of the actual tests is failing, to to trigger you need to break a test. In the system tests comment out an trigger, so that one of the `click` methods has to wait.

```
        QTimer.singleShot(SHORT_DELAY, lambda: self._click_stack_selector())
        QTimer.singleShot(SHORT_DELAY * 2, lambda: self._click_messageBox("OK"))
        # self.main_window.actionSampleLoadLog.trigger()
```

Then run the test with:
`make test-system`

### Expected Behaviour

Should get a timeout on the click method

### Current Behaviour

qWait() call gives an error

### Context
Current main

### Failure Logs

```
mantidimaging/gui/test/gui_system_loading_test.py::TestGuiSystemLoading::test_load_log Traceback (most recent call last):
  File "/home/sam/git/mantidimaging/mantidimaging/gui/test/gui_system_loading_test.py", line 118, in <lambda>
    QTimer.singleShot(SHORT_DELAY, lambda: self._click_stack_selector())
  File "/home/sam/git/mantidimaging/mantidimaging/gui/test/gui_system_loading_test.py", line 48, in _click_stack_selector
    cls._wait_for_widget_visible(DatasetSelectorDialog)
  File "/home/sam/git/mantidimaging/mantidimaging/gui/test/gui_system_base.py", line 98, in _wait_for_widget_visible
    QTest.qWait(delay * 1000)
TypeError: qWait(ms: int): argument 1 has unexpected type 'float'
```

ashmeigh:
### Issue

Given the context and specifics of the error, the resolution involves ensuring that all calls to QTest.qWait() are made with integer arguments, as the function expects. This change directly addresses the TypeError encountered during tests.

Closes #1873

### Description

This update modifies instances in the test suite where QTest.qWait() is called with a float argument instead of an integer. This is achieved by wrapping the multiplication of delay * 1000 with an int() conversion, ensuring compatibility with the QTest.qWait() argument specifications.

### Testing 

Ran the modified test suite using make test-system to ensure that the TypeError no longer occurs.

### Documentation

All changes should be noted in the appropriate file in docs/release_notes*

### DIFF
--- a/docs/release_notes/next/fix-1873-unexpected-float
+++ b/docs/release_notes/next/fix-1873-unexpected-float
@@ -1,0 +1,1 @@
+#1873 : fix typeError: qWait(ms: int): argument 1 has unexpected type 'float'
--- a/mantidimaging/gui/test/gui_system_base.py
+++ b/mantidimaging/gui/test/gui_system_base.py
@@ -95,7 +95,7 @@ class GuiSystemBase(unittest.TestCase):
             for widget in cls.app.topLevelWidgets():
                 if isinstance(widget, widget_type) and widget.isVisible():
                     return True
-            QTest.qWait(delay * 1000)
+        QTest.qWait(int(delay * 1000))
         raise RuntimeError("_wait_for_stack_selector reach max retries")
 
     @mock.patch("mantidimaging.gui.windows.image_load_dialog.view.ImageLoadDialog.select_file")


samtygier-stfc (approved):
Thanks. Fixes the test case for me.

Issue #1776, opened 2023-04-28
Comparing Images:  ROI Doesn't Redraw Correctly When Moved


JackEAllen:
### Summary

Comparing Images Window, accessed by through `workflow >> Compare Images` and selecting okay in dialog menu.

ROI does not redraw smoothly when moving ROI around image stacks (See Images). Other operations which trigger the ROI to redraw resolve this such as scrolling left and right through the spectrum using the line plot below stacks when ROI is enabled.

### Steps To Reproduce

- Open "Compare Data" Menu `workflow >> Compare Images` select ok in dialog menu.
- Select "ROI" button
- Drag ROI around Image stack

### Expected Behaviour

ROI should move smoothly and be redrawn quickly not leaving and lines across the image stack that do not actually represent the ROI's location.

### Current Behaviour

ROI does not redraw smoothly when moving ROI around image stacks (See Images).

### Context

Update Manual Systems Test spreadsheet bugs page when complete

### Screenshot(s)

![image](https://user-images.githubusercontent.com/32413847/235125218-793a647a-d67d-4966-8f62-ec0420e3312f.png)

PR #2034, opened 2024-02-07
Avoid unneeded calculation of ROI in MIImageView

### LINKED ISSUE #2030: MIImageView spends lots of time in _update_roi_region_avg()
### Desired Behaviour

Looks like we are calling `_update_roi_region_avg()` at every time we move to a new slice in the main window. This is a slow operation because it calls mean on the whole ROI. This makes scrolling through the z-axis slow.

A quick fix is to avoid calling it if the ROI graph is not visible, but really it should only be called if the ROI changes.

### Details

To see the issue, Add an `ExecutionProfiler` in `mantidimaging/gui/widgets/mi_image_view/view.py` by adding the method

```python
    def timeLineChanged(self):
        from mantidimaging.core.utility import ExecutionProfiler
        with ExecutionProfiler(msg="timeLineChanged()"):
            super().timeLineChanged()

```
Load a big dataset. Click a point in the z-axis in the main window. A profile message will be output to the console. Note that the output contains `_update_roi_region_avg` and time spend in `.../site-packages/numpy/core/_methods.py:162(_mean)`

With a 2k x 2k x 1k dataset, I see a total time for `timeLineChanged()` of about 0.4 seconds.

samtygier-stfc:
### Issue

Closes #2030

### Description

Previously the ROI line plot was recalculated for every movement to a new slice, even though this does not change the plot and the plot may not be visible.

Now, this is only done on changes to the ROI and only if the ROI plot is visible.

There is a change to only show the ROI info in the message if the ROI is visible. This will need the screenshots updating.

There is also an optimisation to only call `mean()` once.

For me this takes the run time for `timeLineChanged()` with the 2k x 2k x 1k flower dataset from ~0.4 to ~0.01 seconds. 
```
Before:
INFO:         1    0.000    0.000    0.469    0.469 /.../ImageView.py:810(timeLineChanged)
INFO:         1    0.000    0.000    0.454    0.454 /.../ImageView.py:810(timeLineChanged)
INFO:         1    0.000    0.000    0.516    0.516 /.../ImageView.py:810(timeLineChanged)

After:
INFO:         1    0.000    0.000    0.010    0.010 /.../ImageView.py:810(timeLineChanged)
INFO:         1    0.000    0.000    0.006    0.006 /.../ImageView.py:810(timeLineChanged)
INFO:         1    0.000    0.000    0.008    0.008 /.../ImageView.py:810(timeLineChanged)
```

### Testing &  Acceptance Criteria 

Test the responsiveness of moving though a large dataset in the main window. Should be noticeable for a 1k x 1k or 2k x 2k image stack, or benchmarkable (see issue) for a smaller stack.

Check that ROI message and line plot show when clicking the ROI button, and go after unclicking it. Moving through the imagestack should still be fast even with the plot shown, as its only recalculated on ROI moves.

Also check that the ROI selector in the the crop operation works. And the image views in compare images and safe apply work.

### Documentation

Release notes

### DIFF
--- a/mantidimaging/gui/widgets/mi_image_view/view.py
+++ b/mantidimaging/gui/widgets/mi_image_view/view.py
@@ -197,23 +197,27 @@ class MIImageView(ImageView, BadDataOverlay, AutoColorMenu):
         self._refresh_message()
 
     def _update_roi_region_avg(self) -> Optional[SensibleROI]:
-        if not self.ui.roiBtn.isChecked():
-            return None
         if self.image.ndim != 3:
             return None
         roi_pos, roi_size = self.get_roi()
         # image indices are in order [Z, X, Y]
         left, right = roi_pos.x, roi_pos.x + roi_size.x
         top, bottom = roi_pos.y, roi_pos.y + roi_size.y
-        data = self.image[:, top:bottom, left:right]
 
-        data = data.mean(axis=(1, 2))
+        if self.roi.isVisible():
+            mean_val = self.image[self.timeLine.value(), top:bottom, left:right].mean()
+            self.roiString = f"({left}, {top}, {right}, {bottom}) | " \
+                             f"region avg={mean_val:.6f}"
+
+        if not self.ui.roiBtn.isChecked():
+            return None
+
+        data = self.image[:, top:bottom, left:right].mean(axis=(1, 2))
 
         if len(self.roiCurves) == 0:
             self.roiCurves.append(self.ui.roiPlot.plot())
         self.roiCurves[0].setData(y=data, x=self.tVals)
-        self.roiString = f"({left}, {top}, {right}, {bottom}) | " \
-                         f"region avg={data[int(self.timeLine.value())].mean():.6f}"
+
         return SensibleROI(left, top, right, bottom)
 
     def roiClicked(self):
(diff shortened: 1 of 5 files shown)


samtygier-stfc:
I may have reintroduced #1776 :-(

mantidimaging/gui/widgets/mi_image_view/view.py
```diff
@@ -204,17 +203,30 @@ def _update_roi_region_avg(self) -> Optional[SensibleROI]:
...
+
+        data = self.image[:, top:bottom, left:right].mean(axis=(1, 2))
+
+        if len(self.roiCurves) == 0:
+            self.roiCurves.append(self.ui.roiPlot.plot())
+        self.roiCurves[0].setData(y=data, x=self.tVals)
+
         return SensibleROI(left, top, right, bottom)
 
+    def roiClicked(self):
```
JackEAllen, line 224: I would be tempted to add a docstring here to explain why we need this. If i saw this in a year's time with no context, i might question why it is needed.

JackEAllen (changes requested):
Looks good! just a minor suggestion to add one docstring and happy to merge! 👍

JackEAllen:
Actually , just noticed a potential bug. I can't crop the image using operations>crop coordinates. after selecting a new roi shape, the roi returns back to 0,0,200,200

JackEAllen:
> Actually , just noticed a potential bug. I can't crop the image using operations>crop coordinates. after selecting a new roi shape, the roi returns back to 0,0,200,200


https://github.com/mantidproject/mantidimaging/assets/32413847/5cdeae38-f217-4bb7-b6d3-d678879639be

JackEAllen (approved):
Looks good! Happy to approve. We should probably write up an issue and associate it to this PR for reference to add a screenshot test for cropping coords within operations as I could see how this bug could have been introduced quite easily and missed in a review.

samtygier-stfc:
Yes, might be a good candidate for a system test

PR #2047, opened 2024-02-14
Implement Click Dialog for ROI Colour Changes in Table

ashmeigh:
### Issue

Closes #2046

### Description

This pull request introduces a feature allowing users to change the colour of Regions of Interest (ROIs) directly from the table representation in the Spectrum Viewer. By implementing a click dialogue, users can now select an ROI entry in the table, open a colour selection dialogue, and choose a new colour. This enhancement is aimed at providing a consistent and user-friendly interface across the Spectrum Viewer, improving overall usability

While the Spectrum Viewer currently supports changing ROI colours via a right-click context menu within the viewer, there has been no equivalent, intuitive method for altering ROI colours when interacting with the table representation of ROIs. This pull request addresses this gap by extending colour-changing functionality to the table, thereby streamlining the user experience and aligning it with the existing capabilities of the viewer component.

### Testing 

Single and multiple ROI colour changes through the table.
Synchronization of colour changes between the table, viewer, and data model.

### Documentation

Changes in docs/release_notes*

PR #2068, opened 2024-02-26
Table ROI color change: backport to 2.7

samtygier-stfc:
Backport #2047 to release branch. Release notes tweaked.

This completes the recent colour changing work, but adding a additional route to trigger it.

This has a small foot print, so unlikely to introduce any bugs. Unlikely to cause any problems with ROI state consistency.

### DIFF
--- a/docs/release_notes/2.7.rst
+++ b/docs/release_notes/2.7.rst
@@ -10,6 +10,7 @@ New Features
 - #2026: Time of flight graph in the Spectrum Viewer is now resizable
 - #2027: The Spectrum ROI details display in an adjustable table via spinboxes
 - #2048: Update spectrum documentation to reflect changes made to the Spectrum Viewer.
+- #2046 : Enhancement: Implement Click Dialog for Changing ROI Colors in Table
 
 Fixes
 -----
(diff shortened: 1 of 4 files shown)


MikeSullivan7 (changes requested):
Everything looks good with changing the colours in the ROI Tab, however is this colour changing supposed to work in the RITS mode as well?

When you change the colour of the ROI in the RITS mode, it changes the colour of the ROI on the image but does not update the colour of the Spectrum plot, i.e. changing the ROI to green through the right click option:

image

And then when you hover over the ROI, it reverts to its original colour: 

image

However, if you refresh but switching tabs back and forth, then the spectrum plot colour is updated.

If the colour changing option is unwanted for the RITS tab, then I would suggest removing the right click option menu while in the RITS tab.

samtygier-stfc:
Good spot. I don't think this change introduces that bug. I'll see if I can find a quick fix, but maybe not a blocker.

samtygier-stfc:
When the colour is updated by right clicking, it goes through `view.update_roi_color()` which updates the colour in the table. The table then emits a `dataChanged` signal, which takes us to `on_data_in_table_change()` in the view constructor. This calls `on_visibility_change()` which updates the plots.

If the ROI is not in table (e.g. the RITS ROI), then `update_roi_color()` does not update the table, so the change is missed.

Fixed by calling `on_visibility_change()` explicitly.

MikeSullivan7 (approved):
Looks good! Works well, cant find any issues :)

PR #2085, opened 2024-03-01
added --run-tests open in pytest

### LINKED ISSUE #1906: Implementation of --run-system-tests for pytests breaks collection options
### Summary

To enable running of system tests we have a flag `--run-system-tests`. However the way that this is implemented breaks using `--ignore` pytest option. This is probably because it messes with the test collection methods. See `conftest.py`

### Steps To Reproduce

`pytest  --ignore=mantidimaging/gui` should run tests that are not in the `gui` directory.

### Current Behaviour

All tests still run

`--collect-only` can be useful for debugging.

### Context
current main

### LINKED ISSUE #2082: Windows: Running Pytest runs system tests and screenshot tests
### Summary

On Windows running `make test` or `python -m pytest -n auto` runs system tests and Screenshot tests when they shouldn't.
Related to #1906

### Steps To Reproduce

Run `make test` on Windows

### Expected Behaviour

GUI System Tests and Screenshot tests should be ignored by default.

MikeSullivan7:
### Issue

Closes #2082.
Closes #1906.

### Description

On Windows, running `python -m pytest -n auto` was running the system tests and the screenshot eyes tests instead of just the "normal" tests.

A `--run-unit-tests` option is now available with pytest to explicitly only run the unittests by not collecting the `eyes_tests` and the `system-tests`. This functionality should be robust against future added tests.
You can also exclude the unit tests from pytest by not using the  `--run-unit-tests` option.

` --run-unit-tests` has been included into the makefile such that running `make test` should only run the unittest and not the system tests or screenshot eyes tests. These can be run with `make test-system` and `make test-screenshots` (or on windows: `make test-screenshots-win`) respectively. The `--ignore=<Path>` can now be used along with custom flags to explicitly exclude certain paths or tests in the collection phase of `pytest`.

### Acceptance Criteria 

run the following and make sure that the appropriate tests ran as expected:

1. `make test`
2. `make test-verbose`
3. `make test-system`
4. `make test-screenshots` or `make test-screenshots-win`
5. `make check`
6. use the `--ignore=<path>` flag along with a custom flag to check it works as expected, e.g. `python -m pytest --run-unit-tests --ignore=mantidimaging/core`

NOTE: running `--run-eyes-tests` will only work if the API local key is correctly set up. This can be done via the command line on linux but is done via the environment variables in Windows. The environment variables are taken care of in the makefile so running `make test-screenshots-win` should run the eyes tests locally on windows.

### Documentation

Will add to Developer release notes

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -28,8 +28,6 @@ skipped_tests = []
 def pytest_collection_modifyitems(config, items):
     if config.getoption("--run-system-tests"):
         allowed_markers.append(pytest.mark.system.mark)
-    if config.getoption("--run-unit-tests"):
-        allowed_markers.append(pytest.mark.unit.mark)
     if config.getoption("--run-eyes-tests"):
         allowed_markers.append(pytest.mark.eyes.mark)
     if config.getoption("--run-unit-tests") or len(allowed_markers) == 0:
--- a/conftest.py
+++ b/conftest.py
@@ -32,13 +32,15 @@ def pytest_collection_modifyitems(config, items):
         allowed_markers.append(pytest.mark.unit.mark)
     if config.getoption("--run-eyes-tests"):
         allowed_markers.append(pytest.mark.eyes.mark)
+    if config.getoption("--run-unit-tests") or len(allowed_markers) == 0:
+        allowed_markers.append(pytest.mark.unit.mark)
     for item in items:
-        if all(test not in item.nodeid for test in ["gui_system", "eyes_test"]):
-            item.add_marker(pytest.mark.unit)
         if "gui_system" in item.nodeid:
             item.add_marker(pytest.mark.system)
-        if "eyes_test" in item.nodeid:
+        elif "eyes_test" in item.nodeid:
             item.add_marker(pytest.mark.eyes)
+        else:
+            item.add_marker(pytest.mark.unit)
         if any(mark in allowed_markers for mark in item.own_markers):
             pass
         else:
--- a/conftest.py
+++ b/conftest.py
@@ -13,17 +13,25 @@ def _test_gui_system_filename_match(basename: str) -> bool:
     return "gui_system" in basename and "_test.py" in basename
 
 
+def _test_normal_filename_match(path) -> bool:
+    return "gui_system" not in path.basename and "eyes_tests" not in str(path) and "_test.py" in path.basename
+
+
 def pytest_addoption(parser):
     parser.addoption("--run-system-tests", action="store_true", default=False, help="Run GUI system tests")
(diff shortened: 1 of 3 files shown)


conftest.py
```diff
@@ -9,32 +9,41 @@
...
-        skip_system = pytest.mark.skip(reason="use --run-system-tests option to run")
-        for item in items:
-            if "system" in item.keywords:
-                item.add_marker(skip_system)
+    if config.getoption("--run-system-tests"):
+        allowed_markers.append(pytest.mark.system.mark)
+    if config.getoption("--run-unit-tests"):
+        allowed_markers.append(pytest.mark.unit.mark)
+    if config.getoption("--run-eyes-tests"):
+        allowed_markers.append(pytest.mark.eyes.mark)
```
samtygier-stfc, line 34: Could this be tweaked so that if neither of system or eyes is selected, then unit tests gets automatically switched on?

e.g.
```python
    if config.getoption("--run-system-tests"):
        allowed_markers.append(pytest.mark.system.mark)
    if config.getoption("--run-eyes-tests"):
        allowed_markers.append(pytest.mark.eyes.mark)
    if config.getoption("--run-unit-tests") or len(allowed_markers) == 0:
        allowed_markers.append(pytest.mark.unit.mark)
```
It would mean that existing commands would still work.

MikeSullivan7: Ah true that's a good way to do it, I'll change that

conftest.py
```diff
@@ -9,32 +9,41 @@
...
+    for item in items:
+        if all(test not in item.nodeid for test in ["gui_system", "eyes_test"]):
+            item.add_marker(pytest.mark.unit)
+        if "gui_system" in item.nodeid:
+            item.add_marker(pytest.mark.system)
+        if "eyes_test" in item.nodeid:
+            item.add_marker(pytest.mark.eyes)
```
samtygier-stfc, line 41: Maybe clearer?
```python
        if "gui_system" in item.nodeid:
            item.add_marker(pytest.mark.system)
        elif "eyes_test" in item.nodeid:
            item.add_marker(pytest.mark.eyes)
        else:
            item.add_marker(pytest.mark.unit)
```

samtygier-stfc (changes requested):
This works well, thanks.

Possible small tweak. If you could add the path to the system tests to the end of the line in make file rule, so that it it only collects those and does not print out so many skip messages.

```
	${XVFBRUN} python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true --run-system-tests mantidimaging/gui/test/

```

samtygier-stfc (approved):

PR #2086, opened 2024-03-04
Compute NaN Function 

ashmeigh:
Description

Changes made

### DIFF
--- a/mantidimaging/core/operations/nan_removal/nan_removal.py
+++ b/mantidimaging/core/operations/nan_removal/nan_removal.py
@@ -49,22 +49,27 @@ class NaNRemovalFilter(BaseFilter):
         :return: The ImageStack object with the NaNs replaced.
         """
 
-        params = {'replace_value': replace_value, 'mode_value': mode_value}
-        ps.run_compute_func(NaNRemovalFilter.compute_function, data.data.shape[0], data.shared_array, params, progress)
+        if mode_value == "Constant":
+            params = {'replace_value': replace_value}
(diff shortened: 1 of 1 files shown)


samtygier-stfc (commented):
It would be better to move the
```
if mode_value == "Constant":
    ...
else ... :
    ...
```
in to `filter_func()`, and then call separate compute functions.

That would mean we only need to check once, and can raise the value error once from outside the compute function.

samtygier-stfc (changes requested):
See comment

mantidimaging/core/operations/nan_removal/nan_removal.py
```diff
@@ -87,37 +96,20 @@ def register_gui(form: 'QFormLayout', on_change: Callable, view: 'BaseMainWindow
...
     @staticmethod
     def execute_wrapper(mode_field=None, replace_value_field=None):
         mode_value = mode_field.currentText()
         replace_value = replace_value_field.value()
         return partial(NaNRemovalFilter.filter_func, replace_value=replace_value, mode_value=mode_value)
-
-
-def _nan_to_median(data: np.ndarray, size: int, edgemode: str):
-    nans = np.isnan(data)
-    if np.any(nans):
```
samtygier-stfc, line 99: To be added to code comments
```
We need to convert any NaNs in the array into -inf, to prevent the median filter giving bad results. Individual NaN pixels will be removed, but blocks of NaN pixels may remain as blocks of -inf. So afterwards we need to convert any remaining -inf back to NaN.
```

samtygier-stfc (approved):

PR #2088, opened 2024-03-04
Median Compute Function

ashmeigh:
Description

Changes made

### DIFF
--- a/mantidimaging/core/operations/median_filter/median_filter.py
+++ b/mantidimaging/core/operations/median_filter/median_filter.py
@@ -104,14 +104,6 @@ class MedianFilter(BaseFilter):
         array[i] = _median_filter(array[i], size=size, mode=mode)
 
     @staticmethod
-    def cuda_compute_function(i: int, array: np.ndarray, params: Dict[str, Any]):
-        mode = params['mode']
-        progress = params['progress']
-
-        cuda = gpu.CudaExecuter(array.dtype)
-        cuda.median_filter(i, array, mode=mode, progress=progress)
-
-    @staticmethod
     def register_gui(form: 'QFormLayout', on_change: Callable, view) -> Dict[str, Any]:
 
         # Create a spin box for kernel size without add_property_to_form in order to allow a custom validate method
--- a/mantidimaging/core/operations/median_filter/median_filter.py
+++ b/mantidimaging/core/operations/median_filter/median_filter.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from functools import partial
+from logging import getLogger
 from typing import Callable, Dict, Any, TYPE_CHECKING, Tuple
 
 import numpy as np
@@ -15,6 +16,7 @@ from mantidimaging import helper as h
 from mantidimaging.core.gpu import utility as gpu
 from mantidimaging.core.operations.base_filter import BaseFilter
 from mantidimaging.core.parallel import shared as ps
+from mantidimaging.core.utility.progress_reporting import Progress
 from mantidimaging.gui.utility import add_property_to_form
 from mantidimaging.gui.utility.qt_helpers import Type, on_change_and_disable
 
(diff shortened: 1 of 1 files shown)


mantidimaging/core/operations/median_filter/median_filter.py
```diff
@@ -83,19 +82,32 @@ def filter_func(data: ImageStack, size=None, mode="reflect", progress=None, forc
...
+        mode = params['mode']
+        force_cpu = params['force_cpu']
+        size = params.get('size')
+        progress = params.get('progress')
+
+        if not force_cpu:
+            cuda = gpu.CudaExecuter(array.dtype)
+            cuda.median_filter(i, array, mode=mode, progress=progress)  # Call without size if it is None
+        else:
+            array[i] = _median_filter(array[i], size=size if size is not None else 3, mode=mode)
```
samtygier-stfc, line 109: It would be good to keep all of this logic outside the compute function. Its only the CPU path that needs to use the compute function, the cuda one can stay unchanged. (as it is now, it will start lots of cuda jobs which is not what we want).

mantidimaging/core/operations/median_filter/median_filter.py
```diff
@@ -133,36 +145,9 @@ def modes():
...
-def _median_filter(data: np.ndarray, size: int, mode: str):
-    # Replaces NaNs with negative infinity before median filter
-    # so they do not effect neighbouring pixels
+def _median_filter(data: np.ndarray, size: int, mode: str) -> np.ndarray:
     nans = np.isnan(data)
     data = np.where(nans, -np.inf, data)
     data = scipy_ndimage.median_filter(data, size=size, mode=mode)
-    # Put the original NaNs back
     data = np.where(nans, np.nan, data)
     return data
```
samtygier-stfc, line 153: It would be good to keep the comments here as they explain why it does odd things with the nans.

samtygier-stfc (changes requested):
Here it would be

mantidimaging/core/operations/median_filter/median_filter.py
```diff
@@ -83,19 +82,34 @@ def filter_func(data: ImageStack, size=None, mode="reflect", progress=None, forc
...
 
-        h.check_data_stack(data)
         return data
 
+    @staticmethod
+    def compute_function(i: int, array: np.ndarray, params: Dict[str, Any]):
+        mode = params['mode']
+        size = params['size']
+
+        array[i] = _median_filter(array[i], size=size if size is not None else 3, mode=mode)
```
samtygier-stfc, line 103: `size = size if size is not None else 3` -> `size = size` (Because None is caught outside compute fuction)

mantidimaging/core/operations/median_filter/median_filter.py
```diff
@@ -83,19 +82,34 @@ def filter_func(data: ImageStack, size=None, mode="reflect", progress=None, forc
...
         h.check_data_stack(data)
-
-        if not size or not size > 1:
+        if size is None or size <= 1:
             raise ValueError(f'Size parameter must be greater than 1, but value provided was {size}')
 
+        params = {'mode': mode, 'size': size, 'force_cpu': force_cpu, 'progress': progress}
         if not force_cpu:
-            _execute_gpu(data.data, size, mode, progress)
+            ps.run_compute_func(MedianFilter.cuda_compute_function, data.data.shape[0], data.shared_array, params)
```
samtygier-stfc, line 92: Stay as `_execute_gpu(data.data, size, mode, progress)` (because parallelisation is in cuda)

samtygier-stfc (changes requested):
see comments

mantidimaging/core/operations/median_filter/median_filter.py
```diff
@@ -83,19 +84,33 @@ def filter_func(data: ImageStack, size=None, mode="reflect", progress=None, forc
...
+
+        array[i] = _median_filter(array[i], size=size, mode=mode)
+
+    @staticmethod
+    def cuda_compute_function(i: int, array: np.ndarray, params: Dict[str, Any]):
+        mode = params['mode']
+        progress = params['progress']
+
+        cuda = gpu.CudaExecuter(array.dtype)
+        cuda.median_filter(i, array, mode=mode, progress=progress)
```
samtygier-stfc, line 112: This is not need now

samtygier-stfc (changes requested):
Looks good. Just a left over method that can be removed.

samtygier-stfc (approved):

PR #2091, opened 2024-03-04
Clip value compute

ashmeigh:
Description

Changes made

### DIFF
--- a/mantidimaging/core/operations/clip_values/clip_values.py
+++ b/mantidimaging/core/operations/clip_values/clip_values.py
@@ -71,19 +71,15 @@ class ClipValuesFilter(BaseFilter):
 
     @staticmethod
     def compute_function(i: int, arrays: List[np.ndarray], params: Dict[str, Any]):
-        array = arrays[0][i]
+        sample = arrays[i]
+        clip_min = params['clip_min'] if params['clip_min'] is not None else sample.min()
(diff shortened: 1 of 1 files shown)


mantidimaging/core/operations/clip_values/clip_values.py
```diff
@@ -55,27 +57,30 @@ def filter_func(data,
...
-            # the clipping in place and ends up copying the data
-            sample[sample < clip_min] = clip_min_new_value
-            sample[sample > clip_max] = clip_max_new_value
+        ps.run_compute_func(ClipValuesFilter.compute_function, data.data.shape[0], [data.shared_array], params,
+                            progress)
 
         return data
 
+    @staticmethod
+    def compute_function(i: int, arrays: List[np.ndarray], params: Dict[str, Any]):
```
samtygier-stfc, line 73: The type of `arrays` here should be `np.ndarray`. This is because if a single array is passed to `run_compute` it get unpacked from the list. So maybe do it as
```
def compute_function(i: int, array: np.ndarray, params: Dict[str, Any]):
    slice = array[i]
```

samtygier-stfc (changes requested):
Just a small tweak on types (names). Otherwise looks good. This also was not parallelised before, so there it is a performance improvement :-).

samtygier-stfc (approved):

PR #2103, opened 2024-03-05
Disable no site packages

### LINKED ISSUE #2008: Mypy: stop using --no-site-packages
### Desired Behaviour

We currently set mypy to ignore python modules from the environment (site-packages) with the flag `--no-site-packages`. The last remaining package that seems to be causing problems for mypy is `applitools`/`eyes-images` (this has been reported to applitools).

If this could be resolved, and the flag removed, it would open up a lot more code to type checking.

The docs suggest that individual imports can be ignored with `# type: ignore`, but this does not seem to work.

Note: this will throw up lots of new errors, where external types previously caused some annotations to be ignored. E.g.
`mantidimaging/core/utility/data_containers.py:83: error: Incompatible types in assignment (expression has type "ndarray[Any, Any]", base class "SingleValue" defined the type as "float")  [assignment]`

Fixing has been split into 3 PRs

- [x] #2100 
- [x] #2104 
- [ ] #2103

samtygier-stfc:
### Issue

Closes #2008

~~Needs for #2100 and #2104 to be merged first~~

### Description
Final bits for removing the `--no-site-packages` option to mypy.

Add some type stub packages.

Includes a hack for applitools, which has some invalid types.

This reduces the imprecise typing from 42.41% to 41.89% (by my maths, 175 newly precise lines)

Test should pass.

If testing locally you may need to create a new environment

### Documentation

Release notes

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -92,9 +92,8 @@ jobs:
 
       - name: mypy
         shell: bash -l {0}
-        # COMPAT: applitools has some bad signatures, so use --no-site-packages
         run: |
-          mypy --ignore-missing-imports --no-site-packages mantidimaging
+          mypy --ignore-missing-imports mantidimaging
 
       - name: pytest
         timeout-minutes: 10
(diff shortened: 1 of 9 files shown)


samtygier-stfc:
Rebuilding docker to fix those failures

JackEAllen (approved):
Looks good and tests pass locally with an updated dev env

PR #2110, opened 2024-03-06
Operation test tuple params

### LINKED ISSUE #2109: Allow tuple params in operations test case
### Desired Behaviour

Add a way to allow storing a tuple in the `test_cases.json` file.

This is needed for the rebin operations, where the output size can be passed as a `tuple[int, int]`

samtygier-stfc:
### Issue
Closes #2109

### Description

Add a method to allow tuples (and maybe other types) to be included in the test cases.

Encode as a list, where the first element describes the type

Run 
`python scripts/operations_tests/operations_tests.py --verbose --mode compare -k rebin`

### Documentation
Release notes


scripts/operations_tests/operations_tests.py
```diff
@@ -67,6 +67,16 @@ def __bool__(self):
         return self.status == "pass"
 
 
+def process_params(param):
+    """
+    Handle parameter values that cannot be encoded natively in json
+    """
+    if isinstance(param, list):
+        if param[0] == "tuple":
+            return tuple(param[1:])
```
ashmeigh, line 76: Add a ValueError message something like not supported format, first element needs to be "tuple"

samtygier-stfc: This is not needed here, because we want to fall through and return the original if the first item is not `tuple`.

ashmeigh (approved):

Issue #2117, opened 2024-03-08
Enable UP rules in Ruff


samtygier-stfc:
### Desired Behaviour

Enable the [pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up) rules.

PR #2124, opened 2024-03-11
Fix pep 585 and 604 (ruff UP006, UP007) issues

samtygier-stfc:
### Issue

Part of #2117

### Description

For typing standard containers, use the new generic syntax, not the typing module versions. Use `|` notation instead of `Optional` and `Union`
Old methods are deprecated.

Commit generated using Ruff's auto fix

`make ruff`
and
`make mypy`

Should pass

### Documentation

Not needed yet

### DIFF
--- a/CheckLicensesInFiles.py
+++ b/CheckLicensesInFiles.py
@@ -11,14 +11,13 @@ from __future__ import annotations
 
 import json
 import sys
-from typing import List, Dict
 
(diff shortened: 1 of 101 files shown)


JackEAllen (approved):
Changes look good, nice to see we are moving closer to modernising the codebase typing a little.

Issue #2051, opened 2024-02-16
Move operations to compute_function style


samtygier-stfc:
Most of our operations use the partial function style, e.g. in `mantidimaging/core/operations/gaussian/gaussian.py`

```python
f = ps.create_partial(scipy_ndimage.gaussian_filter, ps.return_to_self, sigma=size, mode=mode, order=order)
...
ps.execute(f, [images.shared_array], images.data.shape[0], progress, msg="Gaussian filter")
```
We would like to move them to the Compute Function style e.g. in `mantidimaging/core/operations/arithmetic/arithmetic.py`
```python
ps.run_compute_func(cls.compute_function, images.data.shape[0], images.shared_array, params, progress)
...
    def compute_function(i: int, array: np.ndarray, params: Dict[str, float]):
        array[i] = array[i] * (params["mult"] / params["div"]) + (params["add"] - params["sub"])
```

The `operations_tests.py` script is good for checking that the output from the change is not effected. Most operations will need some examples added to the `test_cases.json`.

See d703d228d8280308cd868d8e86a7d06fe49f4330 for example of changing from partial to compute

PR #2126, opened 2024-03-12
compute operation Normalising  

ashmeigh:
### Issue

This issue was part of #2051 

### Description

###  Changes 

### DIFF
--- a/mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
+++ b/mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
@@ -3,16 +3,12 @@
 from __future__ import annotations
 
 from functools import partial
-from logging import getLogger
 from typing import TYPE_CHECKING
 
 import numpy as np
-
 from mantidimaging import helper as h
 from mantidimaging.core.operations.base_filter import BaseFilter, FilterGroup
 from mantidimaging.core.parallel import shared as ps
-from mantidimaging.core.parallel import utility as pu
-from mantidimaging.core.utility.progress_reporting import Progress
 from mantidimaging.core.utility.sensible_roi import SensibleROI
 from mantidimaging.gui.utility import add_property_to_form
 from mantidimaging.gui.utility.qt_helpers import Type
@@ -33,11 +29,8 @@ class RoiNormalisationFilter(BaseFilter):
     """Normalises the image data by using the average value in a no-sample (air) region of interest (ROI) of the
     image. This scaling operation allows to account for beam fluctuations and different exposure times of
     projections.
-
     Intended to be used on: Projections
-
     When: Always, to ensure that any fluctuations in beam intensity are normalised.
-
     Note: Beam fluctuations are visible by horizontal lines in the sinograms, as  some projections are
     brighter/darker than the neighbours. This can be fixed with this operation.
     """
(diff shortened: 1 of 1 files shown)


samtygier-stfc:
Here are the notes from our catch up.

Neutron flux is variable, so each image has different exposure. The aim of the algorithm is to make the mean air region in each slice equal to each other, by multiplying each frame by a normalisation factor. So for both cases the final step is to have a normalisation array the length of the number of slices, and multiple each slice by the corresponding value.

For the Stack Average case, we take the average of the air region for each slice to get the normalisation array. Then this itself needs normalising to its own mean, before applying to the full image stack.
A non-parallel version would look like:

```python
air_means = images.data[:, air_top:air_bottom, air_left:air_right].mean(axis=1,2)
#[50,48,52,49, ...]
normed_air_means = air_means / air_means.mean()
#[1,0.96,1.04, ...]

for i in range(images.data.shape[0]):
    images.data[i] /= normed_air_means[i]

# so if you looked at the means again, they would all be the same value
new_means = images.data[:, air_top:air_bottom, air_left:air_right].mean(axis=1,2)
# [50,50,50,50, ...]   
```

For the Flat Field mode, we normalise to the flat field. e.g.
```python
air_means = images.data[:, air_top:air_bottom, air_left:air_right].mean(axis=1,2)
#[50,48,52,49, ...]
flat_means = flat_field.data[:, air_top:air_bottom, air_left:air_right].mean(axis=(1, 2))
#[40,38,42,39, ...]
normed_air_means = air_means / flat_means.mean() # 40

for i in range(images.data.shape[0]):
    images.data[i] /= air_means[i]

# again if you looked at the means here, they'd all be equal to the mean of the region in the flat images
air_means = images.data[:, air_top:air_bottom, air_left:air_right].mean(axis=1,2)
# [40,40,40,40, ...]   
```

The compute function for calculating means will need to take a list of arrays, the images and a 1D array to put the means into. It will look something like:
`arrays[1][i] = arrays[0][i, air_top:air_bottom, air_left:air_right].mean()`

mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
```diff
@@ -51,44 +44,80 @@ def filter_func(images: ImageStack,
...
         return images
 
+    @staticmethod
+    def calculate_global(images, region_of_interest, normalisation_mode, flat_field):
+        global_params = {}
+        air_means = np.array([
+            RoiNormalisationFilter._calc_mean(images.data[i], region_of_interest.left, region_of_interest.top,
+                                              region_of_interest.right, region_of_interest.bottom)
+            for i in range(images.data.shape[0])
+        ])
```
samtygier-stfc, line 94: We can do this a single numpy call:
`air_means = images.data[:, region_of_interest.top:region_of_interest.bottom, region_of_interest.left:region_of_interest.right].mean(axis=(1, 2))`

Then can remove `_calc.mean()`

mantidimaging/core/operations/roi_normalisation/roi_normalisation.py
```diff
@@ -51,44 +44,80 @@ def filter_func(images: ImageStack,
...
+            flat_means = flat_field.data[:, region_of_interest.top:region_of_interest.bottom,
+                                         region_of_interest.left:region_of_interest.right].mean(axis=(1, 2))
+            normed_air_means = air_means / flat_means.mean()
+            global_params['normalisation_factors'] = normed_air_means
+        else:
+            raise ValueError(f"Unknown normalisation_mode: {normalisation_mode}")
+
+        print("Global Params:", global_params)
+
+        return global_params
```
samtygier-stfc, line 108: Could remove the dict, and just pass back `normed_air_means`

samtygier-stfc (changes requested):
Passing the tests now.

A few small tweaks. Also once everything is working the debug `print` calls can be removed.

samtygier-stfc (approved):
`operations_tests.py` passes and code looks good. Thanks

PR #2130, opened 2024-03-12
Compute rescale operation 

ashmeigh:
### Issue

This issue was part of #2051 

### Description

###  Changes 

### DIFF
--- a/mantidimaging/core/operations/rescale/rescale.py
+++ b/mantidimaging/core/operations/rescale/rescale.py
@@ -43,7 +43,8 @@ class RescaleFilter(BaseFilter):
         """
 
         params = {'min_input': min_input, 'max_input': max_input, 'max_output': max_output}
-        ps.run_compute_func(RescaleFilter.compute_function, len(images.data), images.data, params)
+        ps.run_compute_func(RescaleFilter.compute_function, len(images.data), [images.shared_array],
+                            params)  # Pass 'params' here
         return images
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


mantidimaging/core/operations/rescale/rescale.py
```diff
@@ -42,9 +42,15 @@ def filter_func(images: ImageStack,
...
         return images
 
+    @staticmethod
+    def compute_function(index: int, array: np.ndarray, params: dict):
+        min_input, max_input, max_output = params['min_input'], params['max_input'], params['max_output']
+        array[index] = np.interp(array[index], [min_input, max_input], [0, max_output])
+
     @staticmethod
     def filter_array(image: np.ndarray, min_input: float, max_input: float, max_output: float) -> np.ndarray:
         image[:] = np.interp(image, [min_input, max_input], [0, max_output])
```
samtygier-stfc, line 56: It looks like we can't remove `filter_array()` because it gets used in a few other places in the code. So it would be good to call `filter_array()` from `compute_function()` to avoid duplication.

Something like
```python
    def compute_function(index: int, array: np.ndarray, params: dict):
        min_input, max_input, max_output = params['min_input'], params['max_input'], params['max_output']
        array[index] = filter_array(array[index], min_input, ...

```

samtygier-stfc (changes requested):

samtygier-stfc (approved):
Looks good.

PR #2133, opened 2024-03-13
Zoomable spectrum plot in the Spectrum Viewer

### LINKED ISSUE #2132: Spectrum Viewer: add draggable zoom box to spectrum plot
### Desired Behaviour

IMAT Scientists would like the option to be able to use a click and drag zoom area to be able to see parts of the spectrum plot in better detail

### Current Behaviour

Currently, the mouse wheel can be used to zoom in and out of the plot but this is hard to control and fairly clunky to use.

MikeSullivan7:
### Issue

Closes #2132 

### Description

The `SpectrumPlotWidget` in `spectrum_widget.py` now uses a custom viewbox which overrides the mouse events of the standard ViewBox class. The custom viewbox uses the `RectMode` mouse mode which uses a built-in zoom drag box. Now to pan the data, the middle mouse button can be used (by temporary using the `PanMode` mouse mode). The right mouse button can be clicked to reset the zoom to its default. The right click menu for this plot has been disabled as it is not useful for this plot.

### Testing 

make check

### Acceptance Criteria 

- Load data into MI and use the Spectrum Viewer.
- Click and drag a zoom box over the data to check the data is zoomed in as expected.
- NOTE: the ToF range box has to be moved to zoom as clicking and dragging this box will move the box instead of creating a zoom box, e.g. below:
image

- Use the middle and right mouse buttons to check they have the behaviour as mentioned above.

### Documentation

will add release note

### DIFF
--- a/mantidimaging/gui/widgets/mi_mini_image_view/view.py
+++ b/mantidimaging/gui/widgets/mi_mini_image_view/view.py
@@ -40,18 +40,12 @@ class MIMiniImageView(GraphicsLayout, BadDataOverlay, AutoColorMenu):
                  name: str = "MIMiniImageView",
                  parent: QWidget | None = None,
                  recon_mode: bool = False,
-                 view_box: ViewBox | None = None):
+                 view_box_type: type[ViewBox] = ViewBox):
         super().__init__()
 
         self.name = name.title()
         self.im = ImageItem()
-        if view_box is None:
-            self.vb = ViewBox(invertY=True, lockAspect=True, name=name)
-        else:
-            self.vb = view_box
-            self.vb.invertY(True)
-            self.vb.setAspectLocked(True)
-            self.vb.register(name)
+        self.vb = view_box_type(invertY=True, lockAspect=True, name=name)
         self.vb.addItem(self.im)
         self.hist = HistogramLUTItem(self.im)
         graveyard.append(self.vb)
--- a/mantidimaging/gui/widgets/mi_mini_image_view/view.py
+++ b/mantidimaging/gui/widgets/mi_mini_image_view/view.py
@@ -38,9 +38,9 @@ class MIMiniImageView(GraphicsLayout, BadDataOverlay, AutoColorMenu):
 
     def __init__(self,
                  name: str = "MIMiniImageView",
-                 parent: 'Optional[QWidget]' = None,
+                 parent: QWidget | None = None,
                  recon_mode: bool = False,
-                 view_box: 'Optional[ViewBox]' = None):
+                 view_box: ViewBox | None = None):
         super().__init__()
 
         self.name = name.title()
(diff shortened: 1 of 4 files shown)


samtygier-stfc:
I'd like to find a way to do this that can be consistent across all our image views and plots. So maybe holding down Ctrl and dragging a box. Or a magnifying glass icon, that you can click on and then zoom, like in matplotlib.

E.g.
```
python -c "from matplotlib import pyplot; pyplot.plot([1,2,3,2,1]); pyplot.show()"
```

MikeSullivan7:
I could see if I can make the zoom box activate upon holding down the ctrl key, but this behaviour should be able to to made consistent with all plots etc by setting the viewbox as `CustomViewBox` or perhaps could be renamed to `ZoomViewBox`.

I thought I would apply this behaviour to the Spectrum Viewer plot only at first to test and see if the scientists like the functionality of it, etc

MikeSullivan7:
Note:
We need to make sure that the ZoomBox plays well with other things in the image views, i.e. the ROIs and the Range Controls in the spectrum viewer. This could perhaps be done by making it such that when `Ctrl` is held done, the mouse interaction is disabled for other objects in the image view. This might have to be done on a case by case basis but it would be good to have this generalised

mantidimaging/gui/widgets/mi_mini_image_view/view.py
```diff
@@ -36,12 +36,22 @@ class MIMiniImageView(GraphicsLayout, BadDataOverlay, AutoColorMenu):
     bright_levels: None | list[int] = None
     levels: list[float]
 
-    def __init__(self, name: str = "MIMiniImageView", parent: QWidget | None = None, recon_mode: bool = False):
+    def __init__(self,
+                 name: str = "MIMiniImageView",
+                 parent: QWidget | None = None,
+                 recon_mode: bool = False,
+                 view_box: ViewBox | None = None):
```
samtygier-stfc, line 43: Could we make the caller pass in the type rather than an instance. e.g 
`view_box_type: type[ViewBox] = ViewBox`

When the `MIMiniImageView` is created, we pass in the type like
`self.image = MIMiniImageView(name="Projection", view_box_type=CustomViewBox)`

mantidimaging/gui/widgets/mi_mini_image_view/view.py
```diff
@@ -36,12 +36,22 @@ class MIMiniImageView(GraphicsLayout, BadDataOverlay, AutoColorMenu):
...
         self.name = name.title()
         self.im = ImageItem()
-        self.vb = ViewBox(invertY=True, lockAspect=True, name=name)
+        if view_box is None:
+            self.vb = ViewBox(invertY=True, lockAspect=True, name=name)
+        else:
+            self.vb = view_box
+            self.vb.invertY(True)
+            self.vb.setAspectLocked(True)
+            self.vb.register(name)
```
samtygier-stfc, line 54: Then this can change to
`self.vb = view_box_type(invertY=True, lockAspect=True, name=name)`

samtygier-stfc (changes requested):
Mostly working well. A couple of code comments.

When I first open the spectum window, I am unable to move the ends of the range in the spectrum view. After pressing and releasing ctrl I can.

Also, does it work to make this work with the ROIs in the keyPressEvent and keyReleaseEvents.

samtygier-stfc (approved):

PR #2141, opened 2024-03-14
Doc extensions: use StringList instead of ViewList

### LINKED ISSUE #2139: mypy error with ViewList in sphinx extensions
### Summary

A recent update to `types-docutils` (from 0.20.0.20240311 to 0.20.0.20240314) has added typing to `ViewList`. ( https://github.com/python/typeshed/commit/a1bfd65e9fa92e1bb61a475c7622ba0376d936d5 https://github.com/python/typeshed/pull/11469 ). We also recently turned on mypy typing for 3rd party types, so are getting the following errors

```
docs/ext/release_notes.py:30: error: Need type annotation for "rst"  [var-annotated]
docs/ext/release_notes.py:56: error: Argument 2 to "nested_parse_with_titles" has incompatible type "ViewList[Any]"; expected "StringList"  [arg-type]
docs/ext/operations_user_doc.py:84: error: Need type annotation for "rst"  [var-annotated]
docs/ext/operations_user_doc.py:91: error: Argument 2 to "nested_parse_with_titles" has incompatible type "ViewList[Any]"; expected "StringList"  [arg-type]
```
This is because `ViewList` is a generic container.
https://github.com/python/typeshed/blob/a1bfd65e9fa92e1bb61a475c7622ba0376d936d5/stubs/docutils/docutils/statemachine.pyi#L114

mypy can only infer the type if the generic type is used in the constructor. This is unlike built in containers, which can infer from the constructor or a method call
```python
l1 = [1]
reveal_type(l1) # -> Revealed type is "builtins.list[builtins.int]

l2 = []
l2.append(1)
reveal_type(l2) # -> Revealed type is "builtins.list[builtins.int]"
```
For custom classes this does not work https://github.com/python/mypy/issues/13134

```python
T = TypeVar("T")
class MyContainer(Generic[T]):
	def set(self, x: T):
		self.item = x

mc = MyContainer()
mc.set(1)
reveal_type(mc) # -> Revealed type is "custom_class.MyContainer[Any]
```

samtygier-stfc:
### Issue
Closes #2139 

### Description

Solved by switching from `ViewList` to `StringList`, as we were using it to hold strings.

We could also have done something like `ViewList[str]` 

Build docs with
`make build-docs`

Check that the release notes for next, and the operations list pages build ok

### Documentation

Release notes


MikeSullivan7 (approved):
Everything looks good, docs build like normal

PR #2142, opened 2024-03-15
Fixes for remaining ruff UP rules

samtygier-stfc:
### Issue
More for #2117

### Description

This fixes the last batch of rules.

Reasons for the rules can be found at https://docs.astral.sh/ruff/rules/#pyupgrade-up

Run
`ruff check --select UP .`

You will see just a few errors remaining. A couple of type annotations (UP035, UP038, UP006). This a from PRs in the past few days. I'll clean up those and any more in a final PR when this get switched on. Some UP014, I don't think we need to change this, so I'll add an ignore in the final commit.

### Documentation

Not needed


JackEAllen (approved):
looks good

Issue #2099, opened 2024-03-04
Font Size Settings


JackEAllen:
### Desired Behaviour

User request to have a setting to increase font sizes across Mantid Imaging - Request first explained using the Spectrum Viewer as an example where text for x, y, and value underneath image view is quite small and hard to read and also axis text for spectrum line plot graph was described as small and hard to read. 

User request asked that maybe the font sizes for just the sections mentioned could be increased or if a setting within Mantid Imaging could be developed which allowed users to alter the font sizes for different sections within Mantid Imaging and save them for future use should Mantid Imaging be closed and re-opened again later. 

This should be consistent with the operating system settings.

### Current Behaviour

Mantid Imaging does not currently offer users the ability to alter font sizes within mantid Imaging.

PR #2144, opened 2024-03-15
Added Settings menu to Main Window with Theme options

### LINKED ISSUE #2143: Theme Selection through a Settings Menu
### Desired Behaviour

It would be useful to have a settings menu accessible from the Main Window -> File -> Settings where we could set various preferences that the User could change to their liking.
An example of which would be a theme selection option which changes the visuals of Mantid Imaging.

### Current Behaviour

Settings menu and theme selection does not currently exist

MikeSullivan7:
### Issue

closes #2143 and #2099.

### Description

A Settings menu has been added via File -> Settings in the Main Window. Clicking on the Settings option opens up a new window where more settings can be added in the future. There is an option to change the current theme of the program to the standard "Fusion" theme or one of the many `qt-material` themes. By changing the theme, all windows of MI will be updated to the new theme during Runtime so the user can see them in real time. The new settings are saved via `QSettings` and are checked on startup so once changed to the users liking, it does not need to be changed again! 

If "Fusion" is the chosen theme, then a dark mode can be activated by clicking on the "dark mode" checkbox. This checkbox is disabled if any of the `qt-material` themes are chosen as light/dark is specified in the names of the `qt-material` themes.

Future work can be done via `qt-material` to make more custom themes chosen be the user.

### Testing 

make check

### Theme Testing

- Open MI and load data in.
- In the Main Window, go to File -> Settings
- Change the theme to and check that all open windows update to the new theme during Runtime.
- With the theme changed, open new windows (recon, spectrum, live viewer, etc) and check that opens in the new theme.
- Repeat with a few different themes to check that it functions as expected.
- With the selected theme not being the default "Fusion", exit out of MI and reopen again, the selected theme should remain.
- All other aspects of MI should function the same as previously.

To simulate a "first time start-up", you need to run `QSettings.clear()` with the appropriate arguments before starting MI. For example you could add `settings.clear()` on line 51 of `mantidimaging/gui/windows/main/presenter.py`, then start MI. This will cause MI to crash on startup but this is good enough to just clear all QSettings for MI. You can then remove the `settings.clear()` line and start MI as usual.

- With a first start-up, check that MI starts up with the same theme as you OS.
- Clear QSettings, change your OS theme and then open a "first startup" MI and check that it now opens with the new theme of your OS.
- You should be able to now change the theme from within MI, and restart MI where the previously chosen MI theme should remain.

### Font Testing

- The fonts used in MI can now be dynamically changed from the settings menu.
- Open the settings menu and another window (i.e. operations, SV, etc), change the font size via the settings and check that the fonts are updated across all windows.
- with the font changed, open a new window and check that the fonts are correct in that window.
- with the font size changed, close MI and reopen. Check that the font size previously chosen is still active.

### OS Theme check

- on first startup, MI will use OS defaults for font size and dark mode
- This can be disabled by unchecking the "use OS defaults" checkbox in the Settings Window.
- After a custom theme or font size has been used, the OS defaults can be toggled on or off via the checkbox

### Documentation

Will add release note!

### DIFF
--- a/mantidimaging/gui/ui/settings_window.ui
+++ b/mantidimaging/gui/ui/settings_window.ui
@@ -72,7 +72,7 @@
             </property>
            </widget>
           </item>
-          <item row="1" column="1">
+          <item row="2" column="1">
            <spacer name="verticalSpacer">
             <property name="orientation">
              <enum>Qt::Vertical</enum>
@@ -84,6 +84,23 @@
              </size>
             </property>
            </spacer>
+          </item>
+          <item row="1" column="1">
+           <widget class="QComboBox" name="menuFontSizeChoice">
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
+           </widget>
+          </item>
+          <item row="1" column="0">
+           <widget class="QLabel" name="menuFontSizeLabel">
+            <property name="text">
+             <string>Menu font size:</string>
+            </property>
+           </widget>
           </item>
          </layout>
         </widget>
--- a/mantidimaging/gui/ui/settings_window.ui
+++ b/mantidimaging/gui/ui/settings_window.ui
@@ -6,14 +6,26 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>418</width>
-    <height>348</height>
+    <width>454</width>
+    <height>444</height>
    </rect>
+  </property>
+  <property name="maximumSize">
+   <size>
+    <width>456</width>
+    <height>449</height>
+   </size>
   </property>
   <property name="windowTitle">
    <string>MainWindow</string>
   </property>
   <widget class="QWidget" name="centralwidget">
+   <property name="sizePolicy">
+    <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
+     <horstretch>0</horstretch>
+     <verstretch>0</verstretch>
+    </sizepolicy>
+   </property>
    <widget class="QWidget" name="verticalWidget" native="true">
     <property name="geometry">
      <rect>
@@ -24,7 +36,7 @@
      </rect>
     </property>
     <property name="sizePolicy">
-     <sizepolicy hsizetype="Preferred" vsizetype="MinimumExpanding">
+     <sizepolicy hsizetype="Expanding" vsizetype="MinimumExpanding">
       <horstretch>1</horstretch>
       <verstretch>1</verstretch>
      </sizepolicy>
@@ -50,8 +62,8 @@
           <rect>
            <x>10</x>
            <y>10</y>
-           <width>341</width>
-           <height>241</height>
+           <width>411</width>
+           <height>361</height>
           </rect>
          </property>
          <layout class="QFormLayout" name="formLayout">
@@ -72,6 +84,13 @@
             </property>
            </widget>
           </item>
+          <item row="1" column="0">
+           <widget class="QLabel" name="menuFontSizeLabel">
+            <property name="text">
+             <string>Menu font size:</string>
+            </property>
+           </widget>
+          </item>
           <item row="1" column="1">
            <widget class="QComboBox" name="menuFontSizeChoice">
             <property name="sizePolicy">
@@ -81,26 +100,6 @@
              </sizepolicy>
             </property>
            </widget>
-          </item>
-          <item row="1" column="0">
-           <widget class="QLabel" name="menuFontSizeLabel">
-            <property name="text">
-             <string>Menu font size:</string>
-            </property>
-           </widget>
-          </item>
-          <item row="3" column="1">
-           <spacer name="verticalSpacer">
-            <property name="orientation">
-             <enum>Qt::Vertical</enum>
-            </property>
-            <property name="sizeHint" stdset="0">
-             <size>
-              <width>20</width>
-              <height>40</height>
-             </size>
-            </property>
-           </spacer>
           </item>
           <item row="2" column="0">
            <widget class="QLabel" name="useOsThemeLabel">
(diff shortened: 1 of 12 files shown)


samtygier-stfc:
Is it possible to have it pick up the system font on linux? For me I have the system font set to 12 on a first run with this I get the font reduced to 10.
Screenshot at 2024-03-20 13-32-16 main
Screenshot at 2024-03-20 13-42-40 921c790

samtygier-stfc:
On IDAaaS you can use xfce4-appearance-settings to change font sizes

MikeSullivan7:
Printing out the typings used in `do_update_UI()`
``` 
    def do_update_UI(self) -> None:
        theme = settings.value('theme_selection')
        extra_style = settings.value('extra_style')
        os_theme = settings.value('os_theme')
        use_dark_mode = settings.value('use_dark_mode')
        override_os_theme = settings.value('override_os_theme')
        print(f"theme type: {type(theme)}")
        print(f"extra_style: {type(extra_style)}")
        print(f"os_theme: {type(os_theme)}")
        print(f"use_dark_mode: {type(use_dark_mode)}")
        print(f"override_os_theme: {type(override_os_theme)}")
```
gives:
```
theme type: <class 'str'>
extra_style: <class 'dict'>
os_theme: <class 'str'>
use_dark_mode: <class 'int'>
override_os_theme: <class 'str'>
```
so using `if use_dark_mode:` is fine but would be better to set as a bool or specify `==0` or `==1` etc

Do you know the steps to reproduce getting stuck in dark mode on Linux as I cant seem to reproduce it in Windows.
I'll try using IDaaS to reproduce it

samtygier-stfc:
Steps to reproduce on linux.
Delete `Mantid\ Imaging.conf Mantid\ Imaging.conf`
Open MI, open settings.
Tick and then untick Dark mode.
Quit MI.
Reopen MI, it will be back in dark mode.

This might be because the qsettings in linux are in ini/toml format, so don't have a type and are treated as strings by default.

MikeSullivan7:
Checking the settings type in linux I get:
```
theme type: <class 'str'>
extra_style: <class 'dict'>
os_theme: <class 'str'>
use_dark_mode: <class 'str'>
override_os_theme: <class 'str'>
```

so looks like it would be a better idea to store the settings as strictly strings to be consistent in Linux and Windows, good catch!

MikeSullivan7:
The default font family and size are found via QFont and then stored in QSettings. For Windows, this is "MS Shell Dlg 2" and font size 12. Once the font size is changed in the settings, the default size is overridden. May be an idea to employ a "Return to defaults" button in the future.

Current QSettings are: 

image

They are stored in `Computer\HKEY_CURRENT_USER\SOFTWARE\mantidproject\Mantid Imaging` in Windows and in a `.conf` file on Linux.

mantidimaging/main.py
```diff
@@ -57,7 +66,13 @@ def parse_args() -> argparse.Namespace:
 def setup_application() -> QApplication:
     QGuiApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
     q_application = QApplication(sys.argv)
-    q_application.setStyle('Fusion')
+    if theme_selection:
+        q_application.setStyle(settings.value('theme_selection'))
+
+    default_font = QFont('-1')
```
samtygier-stfc, line 72: For me on linux I needed to change this to just `default_font = QFont()` to get the OS value. With the `-1` it will always give 12, Noto Sans.

```
QFontInfo(QFont('-1')).pointSize()=12
QFontInfo(QFont('-1')).family()='Noto Sans'
QFontInfo(QFont()).pointSize()=14
QFontInfo(QFont()).family()='Times New Roman'
```

samtygier-stfc (changes requested):
Could some of the logic be made simpler if there was a "Use system theme" check box.

mantidimaging/main.py
samtygier-stfc: I'm happy for the check to only be on linux. One last thing to try is getting it from a widget.
```
QWidget().font().pointSize()
QWidget().font().family()
```

samtygier-stfc:
So I guess use default_font_info to get the system font on linux only

MikeSullivan7:
I think its fine to use on both Windows and Linux, Windows seems to handle the fonts well as it interacts with the scaling well. It looks like it just uses the font size as one relative to the scaling rather than an "absolute" font size

MikeSullivan7:
# Testing on Linux

image

image

`Mantid Imaging.conf`:

```
[General]
default_font_family=DejaVu Sans
default_font_size=11
extra_style=@Variant(\0\0\0\b\0\0\0\x2\0\0\0\x12\0\x66\0o\0n\0t\0_\0s\0i\0z\0\x65\0\0\0\n\0\0\0\b\0\x31\0\x31\0p\0x\0\0\0\x1a\0\x64\0\x65\0n\0s\0i\0t\0y\0_\0s\0\x63\0\x61\0l\0\x65\0\0\0\n\0\0\0\x4\0-\0\x35)
os_theme=Light
theme_selection=Fusion

[welcome_screen]
last_run_version=0.0.0.dev1
show_at_start=true
```

The OS font is now being displayed correctly on Linux

mantidimaging/gui/windows/main/presenter.py
```diff
@@ -29,6 +32,8 @@
 
 RECON_TEXT = "Recon"
 
+settings = QSettings('mantidproject', 'Mantid Imaging')
+
```
samtygier-stfc, line 36: I don't think this is used

samtygier-stfc (changes requested):
This is working much better now. Just a couple of small code issues, and the performance tab.

I see some sizing issues if I use the material theme. And there seems to be case where some text in the image views does not adjust. But happy to merge once above issues done, and deal with these later.

mantidimaging/gui/windows/main/presenter.py
MikeSullivan7: `settings` is used in `do_update_UI()`

```
    def do_update_UI(self) -> None:
        if settings.value('use_os_defaults', defaultValue='True') == 'True':
            extra_style = settings.value('extra_style_default')
            theme = 'Fusion'
            override_os_theme = 'False'
        else:
            extra_style = settings.value('extra_style')
            use_dark_mode = settings.value('use_dark_mode')
            theme = settings.value('theme_selection')
            override_os_theme = settings.value('override_os_theme')
```

samtygier-stfc: Sorry, I put the comment in the wrong place, should have been in `view.py` line 58. I think that one is not used

mantidimaging/gui/ui/settings_window.ui
```diff
@@ -0,0 +1,197 @@
...
+              <width>20</width>
+              <height>40</height>
+             </size>
+            </property>
+           </spacer>
+          </item>
+         </layout>
+        </widget>
+       </widget>
+       <widget class="QWidget" name="performanceTab">
```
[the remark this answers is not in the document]
MikeSullivan7: I think this happened because I had the .ui file open in QT Designer and it didnt switch properly when changing between the two branches, i'll fix it now

MikeSullivan7:
Yes there are some positioning and sizing issues when using the qt_material themes but to fix this would take some modification of the themes style sheets which may take some time. I think it would be best to note that these themes are "experimental" and come back to it at a later PR

mantidimaging/gui/windows/main/presenter.py
MikeSullivan7: Yes you're right, thats a remnant of me refactoring some things from the view to the presenter

samtygier-stfc (approved):
Thanks

PR #2151, opened 2024-03-26
Switch Time of Flight units in Spectrum Viewer

### LINKED ISSUE #2097: Spectrum Viewer: Toggle Axis From ToF to Wavelength
### Desired Behaviour

Users have stated that Wavelength is a more useful metric to use over ToF range (for the most part, ToF range is still useful)

We should make wavelength the default unit of measurement for the spectrum line plot graph, but still allow for ToF range to be used when required.

* Use wavelength as the default unit of measurement in replacement of ToF range for the spectrum line plot graph
* Create a right click setting accessible from the spectrum line plot graph to select between ToF range and Wavelength as the unit of measurement.

### Current Behaviour

Only ToF range is used as a unit of measurement for the spectrum line plot graph within the Spectrum Viewer.

MikeSullivan7:
### Issue

Closes #2097.

### Description

In the Spectrum Viewer, the spectrum plot can be used in different Time of Flight related units, i.e. Neutron Energy, Neutron Wavelength via a right click menu. This menu is greyed out and the "Image Index" unit is selected with no viable sprectra.txt loaded.

image

### Testing 

make check

### Acceptance Criteria 

- Load data in MI with a spectra file (e.g. Brass dataset)
- Right click the spectrum plot and switch to the various units and check that the data displays correctly.
- Move the range window around to check that the "Image index range" label under the plot responds appropriately. Also check that the averaged image in the image view behaves as expected.
- Move the range window from its default position and change the selected unit, check that the range window resizes to the new unit range.
- With the appropriate data loaded, check the `Normalise to open beam` checkbox and see if the plots display correctly.
- Load any data without a spectra file and switch to that data via the `Sample` dropdown menu in the SV.
- Check that the spectrum plot displays with the image index selected as the unit in the right click menu,. The right click menu should also be greyed out to prevent bugs.

### Documentation

Added release notes


mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -55,14 +63,25 @@ class SpectrumViewerWindowModel:
...
     def __init__(self, presenter: SpectrumViewerWindowPresenter):
         self.presenter = presenter
         self._roi_id_counter = 0
         self._roi_ranges = {}
         self.special_roi_list = [ROI_ALL]
 
+        self.tof_data = self.get_stack_time_of_flight()
+        if self.tof_data is None:
+            self.tof_mode = ToFUnitMode.IMAGE_NUMBER
+            #self.presenter.view.units_menu.setEnabled(False)
```
samtygier-stfc, line 81: Does this commented line still need to be here?

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -312,3 +344,25 @@ def do_remove_roi(self, roi_name: str | None = None) -> None:
...
+        tof_mode = self.model.tof_mode
+        tof_axis_label = ""
+        if tof_mode == ToFUnitMode.IMAGE_NUMBER:
+            tof_axis_label = "Image index"
+        if tof_mode == ToFUnitMode.TOF_US:
+            tof_axis_label = "Time of Flight (\u03BC s)"
+        if tof_mode == ToFUnitMode.WAVELENGTH:
+            tof_axis_label = "Neutron Wavelength (\u212B)"
+        if tof_mode == ToFUnitMode.ENERGY:
+            tof_axis_label = "Neutron Energy (MeV)"
```
samtygier-stfc, line 361: Can these strings be put in a dictionary, so this looks more like
`tof_axis_label = UNIT_NAMES[self.model.tof_mode]`

There is already the `allowed_modes` dict. Maybe that could have the full strings with units. Then when making the label you could use `.partition(' (')[0]` to remove the unit.

mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
```diff
@@ -272,31 +271,40 @@ class SpectrumPlotWidget(GraphicsLayoutWidget):
...
+    def add_range(self, range_min: int | float, range_max: int | float) -> None:
         with QSignalBlocker(self.range_control):
             self.range_control.setBounds((range_min, range_max))
             self.range_control.setRegion((range_min, range_max))
         self.spectrum.addItem(self.range_control)
-        self._set_tof_range_label(range_min, range_max)
+        self.set_tof_range_label(range_min, range_max)
+
+    def set_tof_range_label(self, range_min: float, range_max: float) -> None:
+        self._tof_range_label.setText(f'ToF range: {round(range_min, 3)} - {round(range_max, 3)}')
```
samtygier-stfc, line 301: Can do the rounding as a format
`f'... {range_min:.3f} ...'`

mantidimaging/gui/windows/spectrum_viewer/view.py
```diff
@@ -81,6 +81,27 @@ def __init__(self, main_window: MainWindowView):
...
+        self.spectrum_right_click_menu = self.spectrum_widget.spectrum_plot_widget.spectrum.vb.menu
+        self.units_menu = self.spectrum_right_click_menu.addMenu("Units")
+        self.tof_mode_select_group = QActionGroup(self)
+
+        self.allowed_modes = {
+            "Image Index": ToFUnitMode.IMAGE_NUMBER,
+            "Wavelength": ToFUnitMode.WAVELENGTH,
+            "Energy": ToFUnitMode.ENERGY,
+            "us": ToFUnitMode.TOF_US
+        }
```
samtygier-stfc, line 93: The ToF label should be `Time of Flight` to be consistent

mantidimaging/core/utility/unit_conversion.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (C) 2024 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+from __future__ import annotations
+
+import numpy as np
+
+
+class UnitConversion:
```
samtygier-stfc, line 8: Would be good to have some tests for this class.

samtygier-stfc (changes requested):
Looking good. A few small code things.

mantidimaging/gui/windows/spectrum_viewer/presenter.py
MikeSullivan7: I've changed it such that the `allowed modes` dict now looks like 
```
self.allowed_modes = {
            "Image Index": [ToFUnitMode.IMAGE_NUMBER, "Image index"],
            "Wavelength": [ToFUnitMode.WAVELENGTH, "Neutron Wavelength (\u212B)"],
            "Energy": [ToFUnitMode.ENERGY, "Neutron Energy (MeV)"],
            "us": [ToFUnitMode.TOF_US, "Time of Flight (\u03BC s)"]
        }
```

so the tof label is now handled as: 

```
selected_mode = self.view.tof_mode_select_group.checkedAction().text()
        self.model.tof_mode = self.view.allowed_modes[selected_mode][0]
        self.model.set_relevant_tof_units()
        self.view.spectrum_widget.spectrum_plot_widget.set_tof_axis_label(self.view.allowed_modes[selected_mode][1])
```

which is a lot more streamlined!

mantidimaging/core/utility/unit_conversion.py
MikeSullivan7: I have changed the UnitConversion class a bit in PR#2159 so I think that it would be better to write the tests for that PR instead

samtygier-stfc (approved):
Looks good. Thanks

PR #2153, opened 2024-03-27
fix eyes_test GitHub actions

MikeSullivan7:
### Description

The eyes tests have not been running on GitHub Actions as the `--run-eyes-tests` flag was not added to the Windows and Conda work flows. This PR adds those flags so the applitools tests will now run!


samtygier-stfc (approved):
Thanks for spotting this

PR #2155, opened 2024-04-03
Optimize ROI Adjustments in Spectrum Viewer by delaying redraw

### LINKED ISSUE #2076: ROI speed up
Moving the ROIs in spectrum viewer can be slow with big datasets, because a mean is being calculate over a large amount of data. Investigate if something clever can be done to speed it up.

This is likely to be memory bound so probably can't be improved much with threading or vectorisation.

Some related work has already been done #1708 and in #1828

Ideas (solution may be a mix of these):
If we have calculated the mean for and ROI (0,0,100,100) and it gets resized to (0,0,100,90) then maybe we can use the previous result and only calculate the difference (0,90,100,100). Needs some design thought to avoid complexity, avoid accumulating error, and clear cached values when the image stack is modified.

Only recalculate when use ROI move is complete, rather than updating in realtime. Or do realtime updates in a way that does not effect interactivity.

Update spectrum plot in segments, so user can see that something is happening.

ashmeigh:
### Issue 

Progress on #2076 

### Description

This pull request introduces significant performance optimizations to the spectrum viewer, specifically targeting the handling of Regions of Interest (ROIs) within large datasets. The core of this optimization involves a strategic shift from using the sigRegionChanged signal to the sigRegionChangeFinished signal in PyQtGraph's ROI management. This change is aimed at reducing unnecessary computations during the interactive adjustment of ROIs, thus enhancing the responsiveness and efficiency of the spectrum viewer.

### Testing 

The optimizations introduced have been thoroughly tested on a variety of datasets, with a particular focus on large datasets that previously exhibited performance issues during ROI adjustments in the spectrum viewer. The testing process involved the following key steps to ensure comprehensive coverage and validation of the changes:

### Acceptance Criteria 

use datasets, especially those that are large or have previously highlighted performance issues. Adjusting ROIs on these datasets should now result in a noticeably smoother experience, with reduced lag or delay.

### Documentation

All changes should be noted in the appropriate file in docs/release_notes*

### DIFF
--- a/docs/release_notes/next/fix-2076-ROI-speed
+++ b/docs/release_notes/next/fix-2076-ROI-speed
@@ -1,0 +1,1 @@
+#2076: ROI speed up
--- a/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
@@ -205,7 +205,7 @@ class SpectrumWidget(QWidget):
 
         self.roi_dict[name] = roi_object.roi
         self.max_roi_size = roi_object.size()
-        self.roi_dict[name].sigRegionChanged.connect(self.roi_changed.emit)
+        self.roi_dict[name].sigRegionChangeFinished.connect(self.roi_changed.emit)
         self.roi_dict[name].sigClicked.connect(self.roi_clicked.emit)
         self.image.vb.addItem(self.roi_dict[name])
         self.roi_dict[name].hoverPen = mkPen(self.roi_dict[name].colour, width=3)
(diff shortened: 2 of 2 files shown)


samtygier-stfc:
Looks good, but there seem to be some cases where its still recalculating as the ROI is adjusted. I see this on the first ROI, or if the visibility of an ROI is toggle off then on again.

Looks like the signal also set in `set_roi_visibility_flags()`. Is that even needed?

I think it would also make sense to do it for the `range_control` in `SpectrumPlotWidget`, as that can be slow for the same reason.

samtygier-stfc (changes requested):
See above comment

samtygier-stfc (approved):
Thanks. Solves the slow movement of ROIs for me.

PR #2159, opened 2024-04-05
Enter Time of Flight user properties in Spectrum Viewer

MikeSullivan7:
### Issue

Closes #2151.

### Description

A section has been added to the left hand side of the spectrum viewer which allows users to enter the Flight Path (meters) and the Time Delay (microseconds) used as these can change from experiment to experiment. The time delay can be used to artificially alter the neutron wavelength range. It was requested by IMAT that these option be "on show" rather than hidden away in a right click menu for example.

When the values in the spinboxes are changed, the associated attribute is altered in the `unit_conversion` object in the spectrum viewer model. This ensures that the converted data is recalculated properly when the ToF properties are changed and any alterations are contained in the `UnitConversion` class.

### Testing 

make check

### Acceptance Criteria 

- Open MI and load data with no spectrum.txt data
- In the SV, the Right Click `Units` option should be disabled and the ToF properties options on the left should be greyed out
- Load in data with a spectrum.txt file, e.g. the Brass dataset.
- The ToF properties box will be enabled and you can alter the Flight Path (FP) and Time Delay (TD).
- The defaults should be FP = 56m and TD = 0.
- For the Image Index mode, the FP and TD should not alter the data.
- For the Wavelength mode, the increasing the FP implies faster neutrons (with a fixed time of flight) so the wavelength should shift to a smaller number.
- For the Energy mode, the increasing the FP implies faster neutrons (with a fixed time of flight) so the wavelength should shift to a higher number.
- For the ToF us mode, changing the FP should not change the data but the TD will shift the data to the left or right by the value of the TD.
- For the Corrected Brass Dataset (`\\isis\shares\IMAT\ExampleData\Brass\Corrected_Sample_PH20`), there should be a Bragg Edge around 4.3 angstroms.

### Documentation

Will add Release Note

### DIFF
--- a/mantidimaging/core/utility/unit_conversion.py
+++ b/mantidimaging/core/utility/unit_conversion.py
@@ -13,11 +13,12 @@ class UnitConversion:
     mega_electro_volt: float = 1.60217662e-19 / 1e6
     target_to_camera_dist: float = 56  # [m]
     data_offset: float = 0  # [us]
+    tof_data_to_convert: np.ndarray
+    velocity: np.ndarray
 
     def __init__(self, data_to_convert: np.ndarray | None = None) -> None:
-        self.tof_data_to_convert = data_to_convert
         if data_to_convert is not None:
-            self.velocity = self.target_to_camera_dist / (self.tof_data_to_convert + self.data_offset)
+            self.set_data_to_convert(data_to_convert)
 
     def tof_seconds_to_wavelength(self) -> np.ndarray:
         self.check_data()
--- a/mantidimaging/core/utility/unit_conversion.py
+++ b/mantidimaging/core/utility/unit_conversion.py
@@ -32,17 +32,21 @@ class UnitConversion:
         energy_evs = energy / self.mega_electro_volt
         return energy_evs
 
+    def tof_seconds_to_us(self) -> np.ndarray:
+        self.check_data()
+        return (self.tof_data_to_convert + self.data_offset) * 1e6
+
     def set_target_to_camera_dist(self, target_to_camera_dist: float) -> None:
         self.target_to_camera_dist = target_to_camera_dist
 
     def set_data_to_convert(self, data_to_convert: np.ndarray) -> None:
         self.tof_data_to_convert = data_to_convert
-        self.velocity = self.target_to_camera_dist / (self.tof_data_to_convert + self.data_offset)
 
     def check_data(self) -> None:
         if self.tof_data_to_convert is None:
             raise TypeError("Data is not present")
+        else:
+            self.velocity = self.target_to_camera_dist / (self.tof_data_to_convert + self.data_offset)
 
     def set_data_offset(self, data_offset: float) -> None:
         self.data_offset = data_offset * 1e-6
-        self.velocity = self.target_to_camera_dist / (self.tof_data_to_convert + self.data_offset)
--- a/mantidimaging/core/utility/unit_conversion.py
+++ b/mantidimaging/core/utility/unit_conversion.py
@@ -20,7 +20,7 @@ class UnitConversion:
         if data_to_convert is not None:
             self.set_data_to_convert(data_to_convert)
 
-    def tof_seconds_to_wavelength(self) -> np.ndarray:
+    def tof_seconds_to_wavelength_in_angstroms(self) -> np.ndarray:
         self.check_data()
         wavelength = self.planck_h / (self.neutron_mass * self.velocity)
         wavelength_angstroms = wavelength / self.angstrom
@@ -36,9 +36,6 @@ class UnitConversion:
         self.check_data()
         return (self.tof_data_to_convert + self.data_offset) * 1e6
 
-    def set_target_to_camera_dist(self, target_to_camera_dist: float) -> None:
-        self.target_to_camera_dist = target_to_camera_dist
-
     def set_data_to_convert(self, data_to_convert: np.ndarray) -> None:
         self.tof_data_to_convert = data_to_convert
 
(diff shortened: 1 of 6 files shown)


mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -349,11 +355,34 @@ def handle_tof_unit_change(self) -> None:
...
+        tof_axis_label = ""
+        if tof_mode == ToFUnitMode.IMAGE_NUMBER:
+            tof_axis_label = "Image index"
+        if tof_mode == ToFUnitMode.TOF_US:
+            tof_axis_label = "Time of Flight (\u03BC s)"
+        if tof_mode == ToFUnitMode.WAVELENGTH:
+            tof_axis_label = "Neutron Wavelength (\u212B)"
+        if tof_mode == ToFUnitMode.ENERGY:
+            tof_axis_label = "Neutron Energy (MeV)"
+        self.view.spectrum_widget.spectrum_plot_widget.set_tof_axis_label(tof_axis_label)
```
samtygier-stfc, line 368: Is it not possible to keep using `allowed_modes` here?

mantidimaging/core/utility/unit_conversion.py
```diff
@@ -11,18 +11,42 @@ class UnitConversion:
     planck_h: float = 6.62606896e-34  # [JHz-1]
     angstrom: float = 1e-10  # [m]
     mega_electro_volt: float = 1.60217662e-19 / 1e6
+    target_to_camera_dist: float = 56  # [m]
+    data_offset: float = 0  # [us]
```
samtygier-stfc, line 15: Is this stored in us or s?

mantidimaging/core/utility/unit_conversion.py
```diff
@@ -11,18 +11,42 @@ class UnitConversion:
...
         energy = self.neutron_mass * self.velocity / 2
         energy_evs = energy / self.mega_electro_volt
         return energy_evs
+
+    def tof_seconds_to_us(self) -> np.ndarray:
+        self.check_data()
+        return (self.tof_data_to_convert + self.data_offset) * 1e6
+
+    def set_target_to_camera_dist(self, target_to_camera_dist: float) -> None:
+        self.target_to_camera_dist = target_to_camera_dist
```
samtygier-stfc, line 40: Given that the method does not do anything fancy, I just let users of class set the value directly. So the presenter.py below would do
```
self.model.units.target_to_camera_dist = self.view.flightPathSpinBox.value()
```

mantidimaging/core/utility/unit_conversion.py
```diff
@@ -11,18 +11,42 @@ class UnitConversion:
...
+        self.tof_data_to_convert = data_to_convert
+
+    def check_data(self) -> None:
+        if self.tof_data_to_convert is None:
+            raise TypeError("Data is not present")
+        else:
+            self.velocity = self.target_to_camera_dist / (self.tof_data_to_convert + self.data_offset)
+
+    def set_data_offset(self, data_offset: float) -> None:
+        self.data_offset = data_offset * 1e-6
```
samtygier-stfc, line 52: Again, I'd let the user modify the attribute directly, though you'd then need to do the unit conversion in the presenter. Might be good to have the bits in core all in SI units and then do the s-us and m-angstroms in the GUI.

Or have the methods for getting the converted values have the unit in the name. E.g. `tof_seconds_to_wavelength()` -> `tof_wavelength_in_angstroms()`

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -81,6 +81,8 @@ def __init__(self, presenter: SpectrumViewerWindowPresenter):
         else:
             self.tof_mode = ToFUnitMode.WAVELENGTH
 
+        self.units = UnitConversion()
```
samtygier-stfc, line 84: You could move this into `set_relevant_tof_units()` so it does not need to be an instance variable.

mantidimaging/core/utility/unit_conversion.py
samtygier-stfc: `tof_wavelength_in_Å()` is a valid function name in python, but is probably a bad idea 😈

mantidimaging/core/utility/unit_conversion.py
MikeSullivan7: I left it generalised like that just in case something like that would be required in the future but for now we could just set the attribute directly

samtygier-stfc: We can always turn it into a property in the future if needed. (This is a python specific difference from the refactoring book).

mantidimaging/gui/windows/spectrum_viewer/model.py
MikeSullivan7: The only issue with not having it as an instance variable is how to dynamically set the flight path and time delay changes, i.e. 

```
    def handle_flight_path_change(self) -> None:
        self.model.units.target_to_camera_dist = self.view.flightPathSpinBox.value()
        self.model.set_relevant_tof_units()
        self.refresh_spectrum_plot()

    def handle_time_delay_change(self) -> None:
        self.model.tof_data = self.model.get_stack_time_of_flight()
        self.model.units.set_data_offset(self.view.timeDelaySpinBox.value())
        self.model.units.data_offset = self.view.timeDelaySpinBox.value() * 1e-6
        self.model.set_relevant_tof_units()
        self.refresh_spectrum_plot()
```

we wouldn't be able to set them via `self.model.units`.

We could I suppose set them in the constructor, e.g. `__init__(self, data_to_convert, flight_path-56, time_delay=0)` in `UnitConversion` and then write `UnitConversion(self.tof_data, flight_path, time_delay).tof_seconds_to_us()` for example but for me that takes away the feeling of having the `UnitConversion` class more generalised in the future.

samtygier-stfc: My bad, I  missed those and though it was only being used from in `set_relevant_tof_units()`

samtygier-stfc (approved):

PR #2161, opened 2024-04-05
Don't run tests when push only has documentation changes

### LINKED ISSUE #2116: GitHub Action: Don't Run Docs Tests if Docs not Updated
### Desired Behaviour

We should modify our GitHub actions to not perform docs testing if docs are not altered to save on unnecessary compute time which probably adds up over time.
This should include release note changes not triggering docs tests.

By not running if docs are not modified, we are saving compute resources

### Current Behaviour

GitHub actions always run docs tests even if docs are not modified.

samtygier-stfc:
We don't need to run unit, system or screenshot tests

### Issue

Closes #2116 (as best as we can)

### Description
Add a `paths` with negated path, to avoid running main tests suite when there are only documentation changes.

### Testing 

I have tested lots of combinations. See screenshot below, showing that when also excluding the .workflows directory the tests did not run on this PR.

### Documentation

Release notes

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -6,6 +6,8 @@ on:
     - 'main'
     - 'release-*'
   pull_request:
+    paths:
+      - '!docs/**'
   release:
   merge_group:
 
--- a/.github/workflows/cos7_testing.yml
+++ b/.github/workflows/cos7_testing.yml
@@ -6,6 +6,8 @@ on:
     - 'main'
     - 'release-*'
   pull_request:
+    paths:
+      - '!docs/**'
   merge_group:
 
 jobs:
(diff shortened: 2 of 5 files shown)


samtygier-stfc:
Testing with
```
   pull_request:
+    paths:
+      - '!docs/**'
+      - '!.workflows/**'
```
stops those workflows running on this PR. (workflow line is just for testing, in order to ignore that change).

image

From testing here is seems that `paths-ignore` is broken. `paths` works, but considers every change on the PR, not just recent untested ones. So editing a release note on a PR with a code change will still trigger all the tests. But a PR with just doc changes won't run all tests.

samtygier-stfc:
Now I have switched back to just excluding `docs/**` and the tests are not running. So maybe some element of github actions being weird when change these rules on a PR.

I think we need to merge and see what happens.

samtygier-stfc:
:-( I rebased #2162 and it did not run all the tests (it should be cause there are code changes outside of /docs).
image

I'll try again.

PR #2173, opened 2024-04-17
Set style in before compiling ui file for consistent sizing

### LINKED ISSUE #2167: Screenshot tests: spectrum viewer size is non deterministic
### Summary

The size of the spectrum viewer seems to change non deterministically. This is causing failures of the screenshot tests, e.g. `test_spectrum_viewer_add_new_roi`. It can also be seen when running the application, e.g. with

`python -m mantidimaging --path '/data/imaging/IMAT00010675_128/Tomo' -sv`

image
or
image

In `SpectrumViewerWindowView.__init__()` the size always starts as:
`self.size()=PyQt5.QtCore.QSize(905, 752)`
some times it will stay at this size, or some times it will be resized in show()
`self.size()=PyQt5.QtCore.QSize(943, 978)`

### Steps To Reproduce

`python -m mantidimaging --path '/data/imaging/IMAT00010675_128/Tomo' -sv`
or run the screenshot tests with 
`-k test_spectrum_viewer_add_new_roi`

### Expected Behaviour

Consistent size

### Current Behaviour

Varies, but not clear why or when

### Context

And as far as I can tell it is always tall at 5b827ded3bbfd67c28055ea63eaf1d5c87c0edd7 (#2151), but non-deterministic at ea787c4f7a28482b1ac5d72fc7682232204379ff

MikeSullivan7:
### Issue

Close #2167.

### Description

In `gui/mvp_base/view.py` the Qapplication instance style is now set before the .ui file is compiled, which avoids the Spectrum Viewer from unreliably showing in differing sizes, which caused issues with reliable eyes tests.

### Acceptance Criteria 

As in the issue #2167, check open MI with data and open and close the spectrum viewer a large number of times to check that it is consistently being produced at the same size which should be around QSize(914, 978) but may differ depending on your resolution. You can use `python -m mantidimaging --path '/data/imaging/IMAT00010675_128/Tomo' -sv`

Alternatively, you can run the screenshot tests with the added option `-k test_spectrum_viewer_add_new_roi`

You may want to use a print statements like 
```
class BaseMainWindowView(QMainWindow):

    def __init__(self, parent, ui_file=None):
        self._t0 = time.monotonic()
        super().__init__(parent)

        self._has_shown = False

        print(f"{ui_file}: self.size() = {self.size()}")
        QApplication.instance().setStyle(settings.value('theme_selection'))
        print(f"{ui_file}: self.size() = {self.size()}")

        if ui_file is not None:
            compile_ui(ui_file, self)
```
and in `spectrum_viewer\view.py`:
```
    def show(self):
        print(f"SVWV show start    {self.size()=}")
        super().show()
        self.activateWindow()
        print(f"SVWV show end      {self.size()=}")
```

to see the sizing in the command line.

to check the sizing

### DIFF
--- a/mantidimaging/gui/mvp_base/view.py
+++ b/mantidimaging/gui/mvp_base/view.py
@@ -26,11 +26,10 @@ class BaseMainWindowView(QMainWindow):
         super().__init__(parent)
 
         self._has_shown = False
+        QApplication.instance().setStyle(settings.value('theme_selection'))
 
         if ui_file is not None:
             compile_ui(ui_file, self)
-
-        QApplication.instance().setStyle(settings.value('theme_selection'))
 
     def closeEvent(self, e):
         LOG.debug('UI window closed')


samtygier-stfc (approved):
Thanks. This resolves the issue for me locally.

PR #2186, opened 2024-04-30
NaN warnings activated in the Spectrum Viewer

### LINKED ISSUE #2185: Add NaN detection and warning to Spectrum Viewer
### Summary

For the Spectrum Viewer, if NaNs exist in the data, there are gaps in the Spectrum Plot but no obvious explanation as to why. There needs to be a form of warning to the user that there are NaNs in the current slice or in the data stack itself, as there is for the main window and the Operations window.

### Steps To Reproduce

1. Load Spectrum Viewer with data which has NaNs, e.g. the Fe dataset

### Expected Behaviour

Show a warning if there is a NaN in the current slice or give a list of slice indices with NaNs. If possible, when the warning Icon is hovered over, highlight the position of the NaNs in the slice. 

### Current Behaviour

No NaN detection and non complete plot of the Spectrum .

image

image

MikeSullivan7:
### Issue

Closes #2185

### Description

In the `SpectrumWidget` class, the MainWindow is now passed to the `SpectrumProjectionWidget` and the `BadDataOverlay` mixin class is used in the `SpectrumProjectionWidget`.  The warning icon now appears in the Spectrum viewer when a slice which has a NaN in it is shown on the averaged image (as the NaN will propagate). The warning icon retains the same functionality as in the Main Window where it can be clicked on to open the Operations Window such that NaN removal can be performed.

### Testing 

make check

### Acceptance Criteria 

- Open MI with data which has NaNs in it (e.g. the Fe dataset)
- Open the Spectrum Viewer
- Check that if a slice in the averaged image has NaNs, then the warning icon appears
- Test this by changing the Range Control on the Spectrum plot such that a "complete" part of the plot is in the range, and the warning icon should not show.
- Move the Range Control such that an "empty" part of the plot is in the range, and the warning icon should appear. See screenshots for examples.
- Check that clicking the warning icon has the correct functionality.

image

image

### Documentation

Will add release note

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
@@ -116,12 +116,12 @@ class SpectrumWidget(QWidget):
     spectrum_plot_widget: SpectrumPlotWidget
     image_widget: SpectrumProjectionWidget
 
-    def __init__(self, parent: MainWindowView) -> None:
+    def __init__(self, main_window: MainWindowView) -> None:
         super().__init__()
 
         self.vbox = QVBoxLayout(self)
 
-        self.image_widget = SpectrumProjectionWidget(parent)
+        self.image_widget = SpectrumProjectionWidget(main_window)
         self.image = self.image_widget.image
         self.spectrum_plot_widget = SpectrumPlotWidget()
         self.spectrum = self.spectrum_plot_widget.spectrum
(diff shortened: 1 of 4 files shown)


ashmeigh (dismissed):

mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
```diff
@@ -113,12 +116,12 @@ class SpectrumWidget(QWidget):
     spectrum_plot_widget: SpectrumPlotWidget
     image_widget: SpectrumProjectionWidget
 
-    def __init__(self) -> None:
+    def __init__(self, parent: MainWindowView) -> None:
```
samtygier-stfc, line 119: Can we avoid calling this `parent` here. It has a specific meaning in Qt, and caused issues before. I think best to call it `main_window`.

samtygier-stfc: See https://github.com/mantidproject/mantidimaging/pull/1989

The key bit is not passing it to the Qt widget constructor, but still best to call it `main_window`

samtygier-stfc (changes requested):
See comment

samtygier-stfc (approved):
Looks good now.

Tested with data, shows NaNs and loads correct tools.

PR #2221, opened 2024-06-10
Live viewer: Quick handle notification and defer slow work

### LINKED ISSUE #2204: LiveViewer: Out of Synchronisation with Most Recent Image - Difference Increases with Dataset Size
### Summary

"We have been recording radiography images at an exposure time of 5s/image and find that there is a considerable lag for the Live Viewer images to get updated. It took ~3 minutes to display the new image this morning and there was a lag between the image that was displayed and the image being currently collected (the displayed image was more than 100 images behind the one currently being collected). 
Once the user restarted Mantid, the lag was less, but still around 7 to 8 images behind."

"To add context, the current experiment produces images every 5 seconds, so 12 hours of scan will produce 8000+ images. I think we're targeting the parent RB folders so probably the live viewer struggled to scan through that many images to display the latest one."

### Steps To Reproduce

- Create a dataset of ~8000 tiff files
- Create an output directory to copy images too
- Open Mantid Imaging Live Viewer and select output directory
- Run live viewer simulation script to copy files from fake dataset to output directory
- Watch images to see if they fall out of sync

### Expected Behaviour

Live viewer should always show the newest file written within watched directory.

### Current Behaviour

Live viewer fall out of sync with newest image written to directory after so many images within a large dataset.

### Context

Discovered by IMAT when using live viewer to watch 12 hour of scan will produce 8000+ images. Acquisition time was ~5 seconds.

samtygier-stfc:
### Issue

Closes #2204 

### Description

When a diectory is updated we have to check the age of every file in it, to find the newest (QFileSystemWatcher does not give the changed file name). This can be slow on windows, e.g. a second for a few thousand images. During this time new notifications can arrive.

Previously, each notification as acted when it was received. If the rate of notification was to high, then a queue of signals would build up, and MI would appear stuck.

Now we can quickly handle the notification in `_handle_notified_of_directry_change()`, and defer the work with a timer. Any extra notifications can get aborbed by this, and the actuall work in `_handle_directory_change()` only needs to run once.

Also adds a stress test script. e.g.
`simulate_live_data_stress.py -d \mi_test\live -s 0 -f 1000 -r 0.1 -v`
Creates 1000 files at a rate of one every 0.1 seconds.

### Testing 

Not easy to add an automated test

### Acceptance Criteria 

See bug report for methods to trigger issue. Check that you can trigger it on main. Test that there are not long hangs with changes.

Not that this does nothing to make finding the new file faster. So with a large number of files there will still be a few seconds between a file arriving and being displayed.

### Documentation

release notes

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/model.py
+++ b/mantidimaging/gui/windows/live_viewer/model.py
@@ -6,7 +6,7 @@ import time
 from typing import TYPE_CHECKING
 from pathlib import Path
 from logging import getLogger
-from PyQt5.QtCore import QFileSystemWatcher, QObject, pyqtSignal
+from PyQt5.QtCore import QFileSystemWatcher, QObject, pyqtSignal, QTimer
 
 if TYPE_CHECKING:
     from os import stat_result
@@ -109,7 +109,7 @@ class LiveViewerWindowModel:
         self.image_watcher = ImageWatcher(path)
         self.image_watcher.image_changed.connect(self._handle_image_changed_in_list)
         self.image_watcher.recent_image_changed.connect(self.handle_image_modified)
-        self.image_watcher._handle_directory_change(str(path))
+        self.image_watcher._handle_notified_of_directry_change(str(path))
 
     def _handle_image_changed_in_list(self, image_files: list[Image_Data]) -> None:
         """
@@ -171,7 +171,11 @@ class ImageWatcher(QObject):
         super().__init__()
         self.directory = directory
         self.watcher = QFileSystemWatcher()
-        self.watcher.directoryChanged.connect(self._handle_directory_change)
+        self.watcher.directoryChanged.connect(self._handle_notified_of_directry_change)
+        self.handle_change_timer = QTimer(self)
+        self.handle_change_timer.setSingleShot(True)
+        self.handle_change_timer.timeout.connect(self._handle_directory_change)
+        self.changed_directory = directory
 
         self.recent_file_watcher = QFileSystemWatcher()
         self.recent_file_watcher.fileChanged.connect(self.handle_image_modified)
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -217,15 +221,22 @@ def sort_images_by_modified_time(images: list[Image_Data]) -> list[Image_Data]:
         """
         return sorted(images, key=lambda x: x.image_modified_time)
 
-    def _handle_directory_change(self, directory: str) -> None:
+    def _handle_notified_of_directry_change(self, directory: str) -> None:
+        """
+        Quickly handle the notification event for a change, and start the timer if needed. The means that if files are
+        arriving faster than _handle_directory_change() can process them, we don't end up with a queue of work to do
```
JackEAllen, line 227: ** Potential typo here, maybe change to "If files are arriving faster than _handle_directory_change() can process them, so that we don't end up with a queue of work to do"

I'd also probably add a brief description of the timers purpose here to clear any possibility of misunderstanding it's purpose.

scripts/simulate_live_data_stress.py
```diff
@@ -0,0 +1,68 @@
+# Copyright (C) 2024 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+
```
JackEAllen, line 3: Could we move this whole script into the simulate live data utility script where if no source is specified, we use your implementation to draw new images. This way, we won't grow the number of separate utility scripts we have and keep any live view testing/simulation within one script to keep things simple.

JackEAllen (changes requested):
Overall, looks good, thanks for chasing down this bug and putting a measure in to better handle the build up of queued signals. The implementation of `_handle_notified_of_directry_change` looks good, just a small typo in a docstring. 

The main bit of feedback I have is just to move `scripts/simulate_live_data_stress.py` into `scripts/simulate_live_data.py`. This would not only keep the number of utility scripts we have small, but also improve usability of `scripts/simulate_live_data.py` by no longer requiring users/devs from needing to enter a source directory if we draw an image if no source is provided.

scripts/simulate_live_data_stress.py
samtygier-stfc: I've dropped this out of the PR and I'll do a separate PR with the merged script.

JackEAllen: Thanks, I'll give this another quick review now

JackEAllen (commented):
Thanks for the changes, I've re-tested and all looks good. Files don't fall out of sync as badly when copying a large number of images into a director watched by the live viewer at a high rep rate.

JackEAllen (approved):

PR #2223, opened 2024-06-11
Changed the polyfit to compute function

### LINKED ISSUE #2051: Move operations to compute_function style

ashmeigh:
### Issue

Closes #2051

### Description

This pull request refactors the polyfit_correlation module to transition from the partial function style to the compute_function approach. This change improves code clarity, maintainability, and performance.

Changes
Refactored find_center to use the new compute_function method.
Introduced compute_correlation_error to encapsulate core logic for parallel execution.
Allocated shared memory for arrays to enhance performance.
Updated code to handle shared memory arrays and ensure compatibility.

### Testing 

Verifiy the refactored find_center function by running unit tests.
Ensure that compute_correlation_error correctly computes correlation errors in parallel.
Test shared memory allocation and access in a multiprocessing context.

### Acceptance Criteria 

Ensure all unit tests pass.
Verify that the find_center function performs as expected with different image inputs.
Check that the compute_function handles shared memory correctly without errors.


MikeSullivan7:
> I get the hang when I run locally with `python -m pytest -vs mantidimaging/core/rotation/test/polyfit_correlation_test.py -k test_find_center`
> 
> Also some error
> 
> ```
> mantidimaging/core/rotation/test/polyfit_correlation_test.py::PolyfitCorrelationTest::test_find_center Exception ignored in: <function SharedArray.__del__ at 0x7760ac343910>
> Traceback (most recent call last):
>   File "/home/sam/git/mantidimaging/mantidimaging/core/parallel/utility.py", line 155, in __del__
>     if self.has_shared_memory:
>   File "/home/sam/git/mantidimaging/mantidimaging/core/parallel/utility.py", line 166, in has_shared_memory
>     return self._shared_memory is not None
> AttributeError: 'SharedArray' object has no attribute '_shared_memory'
> Exception ignored in: <function SharedArray.__del__ at 0x711fd2343910>
> ```

This seems to be a Linux specific issue, as I dont get this on Windows. Makes sense with it being related to the Shared Memory.

samtygier-stfc:
I think the problem here is that `SharedArray` objects are being passed to the compute function as `params` where they need to be in using `arrays`. `params` should only contain simple types.

This is needed because `SharedArray`s need to be converted to `SharedArrayProxy`s before being passed to the other processes (see `_check_shared_mem_and_get_data`). If this does not happen then the other process thinks it owns the `SharedArray` and therefore tries to deleted it when it goes out of scope.

mantidimaging/core/rotation/polyfit_correlation.py
```diff
@@ -26,22 +25,40 @@ def do_calculate_correlation_err(store: np.ndarray, search_index: int, p0_and_18
...
+    # Run compute function parallel
+    params = {
+        'shared_projections': shared_projections_proxy,
+        'shared_search_range': shared_search_range_proxy,
+        'image_width': images.width,
+    }
+    ps.run_compute_func(compute_correlation_error,
+                        len(search_range_list), [min_correlation_error],
+                        params,
+                        progress=progress)
```
samtygier-stfc, line 56: Rather than dealing with the proxys manually, you can pass them into `run_compute_func` in the arrays argument:
```
    ps.run_compute_func(compute_correlation_error,
                        len(search_range_list),
                        [min_correlation_error, shared_projections, shared_search_range],
                        params,
                        progress=progress)
```
`params` is just for simple params line numbers or strings, that can safely be passed between processes. `arrays` handle all the difficult bits of sharing large arrays between processes.

mantidimaging/core/rotation/polyfit_correlation.py
```diff
@@ -26,22 +25,40 @@ def do_calculate_correlation_err(store: np.ndarray, search_index: int, p0_and_18
...
     slices = np.arange(images.height)
-    shift = pu.create_array((images.height, ))
-
+    shift = pu.create_array((images.height, ), dtype=np.float32)
     search_range = get_search_range(images.width)
-    min_correlation_error = pu.create_array((len(search_range), images.height))
-    shared_search_range = pu.create_array((len(search_range), ), dtype=np.int32)
-    shared_search_range.array[:] = np.asarray(search_range, dtype=np.int32)
-    _calculate_correlation_error(images, shared_search_range, min_correlation_error, progress)
+    search_range_list = list(search_range)
```
samtygier-stfc, line 34: I don't think there is a need to make a `list` here. Just use `search_range`

samtygier-stfc (changes requested):

mantidimaging/core/rotation/polyfit_correlation.py
ashmeigh: mantidimaging\core\rotation\polyfit_correlation.py:60: error: Argument 2 to "_find_shift" has incompatible type "range"; expected "list[int]"  [arg-type]

samtygier-stfc: you could change the type annotation for `search_range` in `_find_shift()` back to `range`

ashmeigh: lol

mantidimaging/core/rotation/polyfit_correlation.py
```diff
@@ -26,15 +25,33 @@ def do_calculate_correlation_err(store: np.ndarray, search_index: int, p0_and_18
...
+
+    # Copy projections to shared memory
+    shared_projections = pu.create_array((2, images.height, images.width), dtype=np.float32)
+    shared_projections.array[0][:] = images.projection(0)
+    shared_projections.array[1][:] = np.fliplr(images.proj180deg.data[0])
+
+    # Prepare parameters for the compute function
+    params = {
+        'shared_projections': shared_projections.array_proxy,
+        'shared_search_range': shared_search_range.array_proxy,
```
samtygier-stfc, line 47: The proxies don't need to be in the params, because we are passing these in the `arrays` argument.

mantidimaging/core/rotation/polyfit_correlation.py
```diff
@@ -46,45 +63,37 @@ def find_center(images: ImageStack, progress: Progress) -> tuple[ScalarCoR, Degr
...
 
-    arrays = [min_correlation_error, shared_search_range, shared_projections]
-    ps.execute(do_search_partial,
-               arrays,
-               num_operations=min_correlation_error.array.shape[0],
-               progress=progress,
-               msg="Finding correlation on row")
+def compute_correlation_error(index: int, min_correlation_error: np.ndarray, params: dict[str, Any]):
+    shared_projections = params['shared_projections'].array
+    shared_search_range = params['shared_search_range'].array
```
samtygier-stfc, line 76: Second argument is now the list of arrays, so these need to be unpacked like in `rebin.py`

samtygier-stfc (changes requested):

samtygier-stfc:
I've rebased and squashed the commits to reduce the churn.

samtygier-stfc (approved):

PR #2229, opened 2024-06-13
Refactor Codebase by Removing Obsolete Functionality due to Introducing Efficient Compute Functions

ashmeigh:
### Description

This pull request proposes the removal of specific functions and a type alias from our project that have become redundant due to the introduction of more modern and efficient compute functions. This change aims to enhance system performance and code maintainability.

### Functions Removed

- **`inplace2(func, data, i, **kwargs)`**: Executes `func` in-place on two elements from `data` at index `i`. It's typically used for operations that modify these elements based on each other.
- **`inplace1(func, data, i, **kwargs)`**: Applies `func` directly to a single element at index `i` in `data`. This is used for simple, direct modifications of individual items.
- **`return_to_self(func, data, i, **kwargs)`**: Applies `func` to an element at index `i` and updates that same element with the result. It ensures the transformation of individual data points while maintaining their position in the array.
- **`inplace_second_2d(func, data, i, **kwargs)`**: Uses `func` to modify an element at index `i` in one dataset and a whole 2D array in another. This is useful for updating 2D data based on specific 1D index-related values.
- **`return_to_second_at_i(func, data, i, **kwargs)`**: Executes `func` on an element at index `i` from one dataset and stores the result at the same index in a second dataset. It is ideal for operations where outputs are mapped directly to another data structure.

### Testing

Extensive testing has confirmed that these changes do not negatively impact our operations. All tests have been updated to align with the new system, ensuring that it performs reliably and efficiently.

### Acceptance Criteria

1. **Run Unit Tests**: Ensure that all existing unit tests pass successfully. This includes tests that have been updated to reflect the removal of the deprecated functions.
2. **Integration Testing**: Perform integration tests to verify that the new compute functions interact correctly with other parts of the system.
3. **Performance Benchmarking**: Compare the performance of the new compute functions with the old functions to ensure that there are improvements or at least no regressions.
4. **Manual Testing**: Conduct manual testing on critical workflows that previously relied on the removed functions to confirm they operate as expected.

### Documentation

The documentation has been updated to reflect these changes. All relevant modifications have been documented in the appropriate file in `docs/release_notes`.

### DIFF
--- a/mantidimaging/core/parallel/shared.py
+++ b/mantidimaging/core/parallel/shared.py
@@ -2,11 +2,13 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, TYPE_CHECKING
 from collections.abc import Callable
-from numpy import ndarray  # Import ndarray type from numpy
 
 from mantidimaging.core.parallel import utility as pu
+
+if TYPE_CHECKING:
+    from numpy import ndarray
 
 ComputeFuncType = (Callable[[int, list['ndarray'], dict[str, Any]], None]
                    | Callable[[int, 'ndarray', dict[str, Any]], None])
--- a/mantidimaging/core/parallel/shared.py
+++ b/mantidimaging/core/parallel/shared.py
@@ -2,80 +2,11 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 from __future__ import annotations
 
-from functools import partial
-from typing import Any, TYPE_CHECKING
+from typing import Any
 from collections.abc import Callable
+from numpy import ndarray  # Import ndarray type from numpy
 
 from mantidimaging.core.parallel import utility as pu
-
-if TYPE_CHECKING:
-    from numpy import ndarray
-
-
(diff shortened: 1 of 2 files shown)


samtygier-stfc (changes requested):
Can you tweak this to not change the spacing in the docstrings for `create_partial` and `execute`

mantidimaging/core/parallel/shared.py
```diff
@@ -2,81 +2,12 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 from __future__ import annotations
 
-from functools import partial
-from typing import Any, TYPE_CHECKING
+from typing import Any
 from collections.abc import Callable
+from numpy import ndarray  # Import ndarray type from numpy
```
samtygier-stfc, line 7: This should stay inside the `if TYPE_CHECKING:` if `ndarray` is only used here for type annotations

samtygier-stfc (changes requested):

samtygier-stfc (approved):
Test operations and recon.

Thanks. Good to see these removed.

samtygier-stfc (changes requested):
One more thing, could you add a release note to say that this removes shared.execute and associated functions.

samtygier-stfc (approved):

Issue #1150, opened 2021-09-21
Solution to building docs for old versions


samtygier-stfc:
### Summary

Building docs requires importing all the modules in mantidimaging. To get the docs for old releases sphinx-multiversion checks out each release, and builds those docs. So to build the docs we need an environment with that can import all the old versions of mantidimaging.

Its only a matter of time until it becomes impossible to have a single python environment that can import all versions of mantidimaging. Currently we are keeping sarepy and hence the dtasev channel in our environment to allow the old docs to build (  #1149 ).

### Steps To Reproduce

Remove sarepy from the environment and run `python3 ./setup.py docs` 

### Expected Behaviour
Builds versioned docs

### Current Behaviour
```
ModuleNotFoundError: No module named 'sarepy'
```

### Context

Happens when we update the environment.

### Failure Logs

```
Exception occurred:
  File "/tmp/tmpctpwl4hm/c1aa0b13a9d3b8c7fba3cecaedb638dfb05d3b49/mantidimaging/core/operations/remove_all_stripe/remove_all_stripe.py", line 8, in <module>
    from sarepy.prep.stripe_removal_original import remove_all_stripe
ModuleNotFoundError: No module named 'sarepy'
The full traceback has been saved in /tmp/sphinx-err-u0b78ezd.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Traceback (most recent call last):
Running setup.py internal_docs_api
  File "/usr/share/miniconda3/envs/mantidimaging-dev/bin/sphinx-multiversion", line 8, in <module>
    sys.exit(main())
Running setup.py internal_docs
Running sphinx-multiversion
  File "/usr/share/miniconda3/envs/mantidimaging-dev/lib/python3.8/site-packages/sphinx_multiversion/main.py", line 363, in main
    subprocess.check_call(cmd, cwd=current_cwd, env=env)
  File "/usr/share/miniconda3/envs/mantidimaging-dev/lib/python3.8/subprocess.py", line 364, in check_call
    raise CalledProcessError(retcode, cmd)
[...]

PR #2232, opened 2024-06-20
Remove sarepy from dependancies

samtygier-stfc:
### Issue

We migrated from sarepy to algotom in  #1023, but had to keep the dependency to enable building old documentation.

This was resolved in #1150, so now its possible to remove the package

### Description
Remove sarepy from dependancies and remove dtasev channel

### Acceptance Criteria 
Automated testing should all pass

### Documentation

Not needed

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -1,7 +1,6 @@
 name: mantidimaging-dev
 channels:
   - mantidimaging/label/unstable
-  - dtasev
   - astra-toolbox
   - conda-forge
   - ccpi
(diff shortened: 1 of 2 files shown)


JackEAllen (changes requested):
Looks good! I did notice that we need to update some a Docstring and tooltip that still reference Sarepy that we should swap to say using Algotom which make sense to do as part of this PR.

`mantidimaging/core/operations/remove_stripe_filtering/remove_stripe_filtering.filter_func` Line 49: 
```
:param filtering_dim: Whether to use a 1D or 2D low-pass filter. This
   uses different Sarepy methods.
```
and:
`mantidimaging/core/operations/remove_stripe_filtering/remove_stripe_filtering.register_gui` Line 121:
```
tooltip="Whether to use a 1D or 2D low-pass filter. "
    "This uses different Sarepy methods")
```

JackEAllen (approved):
Thanks for swapping the last few references from sarepy to algotom. Looks good!

PR #2226, opened 2024-06-13
Spectrum viewer: use QFormLayout to layout ToF labels and widgets

samtygier-stfc:
### Issue

Closes #2225

### Description

Use QFormLayout to layout ToF labels and widgets

Previously the QGroupBox had no layout set, so the widgets just used absolute positioning.

### Testing 

The screenshot tests should pick up the change and prevent regressions

### Acceptance Criteria 

Open the spectrum viewer, check that all the labels in the left column are visible

### Documentation

Not needed (small fix for something that is new in the release)

PR #2234, opened 2024-06-21
Toggle shuttercount

### LINKED ISSUE #2230: Spectrum Viewer: ShutterCount Front-end Toggle Switch
Add a toggle switch to Spectrum Viewer front-end to toggle application of ShutterCount correction

Image

JackEAllen:
### Issue

 Closes #2230 

### Description

Modify Spectrum Viewer to add toggle interface for ShutterCount normalisation correction and warning icon when ShutterCount correction is toggled on, ShutterCount stacks are not loaded. 

### Testing 

* Manually tested on both Windows and Ubuntu
* Updated Unit tests and verified tests pass
* Update GUI screenshot tests

### Acceptance Criteria 

* New Screenshot baselines for the spectrum viewer look okay 
* Load the following from a ToF dataset into Mantid Imaging:
    * Sample stack
    * Open beam stack
    * Spectra file
* Open the spectrum viewer window
* Resize the default ROI to something smaller and centralised on the sample you loaded in.
* Verify that the ShutterCount checkbox is disabled and that you cannot enable ShutterCount correction.
* Apply normalisation - verify that the ShutterCount checkbox only becomes active once a valid stack has been selected. 
* Apply ShutterCount correction and verify a warning Icon appears to warn that you "Need 2 selected stacks".
* Un-check ShutterCount correction and verify that the error icon is removed from window
* re-check ShutterCount correction and verify warning icon appears
* From the main window, load ShutterCounts for sample stack then check that the warning has changed to ""Need 2 different stacks" and then load the ShutterCounts for the open beam stacks "_File>Load ShutterCounts file for stack..._"
* Verify that the error icon is removed from the spectrum viewer window now that both ShutterCount stacks are loaded.
* Notice how the spectrum scale changes accordingly when toggling ShutterCount correction on and off. 

### Documentation

`docs/release_notes/next/feature-2230-ShutterCount_Correction_CheckBox`

### DIFF
--- a/mantidimaging/gui/ui/spectrum_viewer.ui
+++ b/mantidimaging/gui/ui/spectrum_viewer.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>1001</width>
-    <height>766</height>
+    <height>908</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -80,7 +80,7 @@
           <property name="sizeHint" stdset="0">
            <size>
             <width>20</width>
-            <height>20</height>
+            <height>10</height>
            </size>
           </property>
          </spacer>
@@ -154,7 +154,56 @@
           <property name="sizeHint" stdset="0">
            <size>
             <width>20</width>
-            <height>20</height>
+            <height>10</height>
+           </size>
+          </property>
+         </spacer>
+        </item>
+        <item>
+         <layout class="QHBoxLayout" name="horizontalLayout_2">
+          <item>
+           <widget class="QCheckBox" name="normalise_ShutterCount_CheckBox">
+            <property name="text">
+             <string>ShutterCount Correction</string>
+            </property>
+            <property name="checked">
+                <bool>false</bool>
+            </property>
+           </widget>
+          </item>
+          <item>
+           <widget class="QLabel" name="shuttercountErrorIcon">
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+              <horstretch>0</horstretch>
+              <verstretch>0</verstretch>
+             </sizepolicy>
+            </property>
+            <property name="minimumSize">
+             <size>
+              <width>32</width>
+              <height>32</height>
+             </size>
+            </property>
+            <property name="text">
+             <string/>
+            </property>
+           </widget>
+          </item>
+         </layout>
+        </item>
+        <item>
+         <spacer name="ShuttercountSpacer">
+          <property name="orientation">
+           <enum>Qt::Vertical</enum>
+          </property>
+          <property name="sizeType">
+           <enum>QSizePolicy::Fixed</enum>
+          </property>
+          <property name="sizeHint" stdset="0">
+           <size>
+            <width>20</width>
+            <height>10</height>
            </size>
           </property>
          </spacer>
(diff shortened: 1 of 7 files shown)


mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -250,7 +258,10 @@ def get_shuttercount_normalised_correction_parameter(self) -> float:
         open_shuttercount = self.get_stack_shuttercounts(self._normalise_stack)
         if sample_shuttercount is None or open_shuttercount is None:
             return 1.0  # No shutter count data available so no correction needed
-        normalised_shuttercounts = sample_shuttercount / open_shuttercount
+        normalised_shuttercounts = np.divide(sample_shuttercount.astype(np.float32),
+                                             open_shuttercount.astype(np.float32),
+                                             out=np.ones_like(sample_shuttercount, dtype=np.float32),
+                                             where=open_shuttercount != 0)
```
JackEAllen, line 264: I noticed we actually needed to use a safe divide here to better handle zero values in ShutterCounts and remove warning from logs about invalid division.

samtygier-stfc: Or could just divide the first values in the array, and give an error if that is not a number.

JackEAllen: I want to divide the whole stacks as there's a chance we could still be required to use all available values to apply ShutterCount correction going forward when offering different levels of correction.  Happy to remove if we find that there's a low likelihood of the other values being useful for the correction.

mantidimaging/gui/ui/spectrum_viewer.ui
```diff
@@ -407,36 +456,58 @@
...
+             <x>80</x>
+             <y>50</y>
+             <width>211</width>
+             <height>19</height>
+            </rect>
+           </property>
+           <property name="suffix">
+            <string/>
+           </property>
+          </widget>
```
samtygier-stfc, line 510: Looks like this is undoing the fix in #2226

JackEAllen: yeah, was aware of this, but hadn't got around to fixing yet.

JackEAllen: Fixed locally, will push up when i next rebase

JackEAllen:
This PR changes the spectrum viewer GUI and therefore I have update the screenshot test baselines

mantidimaging/gui/windows/main/view.py
```diff
@@ -331,6 +331,7 @@ def load_shuttercounts_dialog(self):
         QMessageBox.information(
             self, "Load complete", f"{selected_file} was loaded as a shutter count file into "
             f"{stack_to_add_shuttercounts_to}.")
+        self.spectrum_viewer.handle_shuttercount_change()
```
samtygier-stfc, line 334: This fails if the spectrum viewer has not been opened. It ought to emit a signal here, and the spectrum viewer should hook that up to `handle_shuttercount_change`

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -208,6 +208,17 @@ def normalise_issue(self) -> str:
             return "Stack shapes must match"
         return ""
 
+    def shuttercount_issue(self) -> str:
+        """
+        Return an error message if there is an issue with the shutter count data.
+        """
+        assert self._stack and self._normalise_stack is not None
```
samtygier-stfc, line 215: Should that be `assert self._stack is not None and self._normalise_stack is not None`

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -317,6 +331,17 @@ def has_stack(self) -> bool:
         """
         return self._stack is not None
 
+    def has_normalise_stack(self) -> bool:
+        """
+        Check if data is available to export
+        @return: True if data is available to export and False otherwise
+        """
+        return self._normalise_stack is not None
```
samtygier-stfc, line 339: I can't see this being used anywhere

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -317,6 +331,17 @@ def has_stack(self) -> bool:
...
+
+    def apply_shuttercount_correction(self) -> bool:
+        """Return True if there are no shuttercount issues and False otherwise"""
+        return self.shuttercount_issue() == ""
```
samtygier-stfc, line 343: This needs renaming, the current name makes it sound like it applies the correction.

Also not sure why it is used when exporting to a file. Shouldn't those use `self.view.shuttercount_norm_enabled()`

samtygier-stfc (changes requested):

mantidimaging/gui/windows/spectrum_viewer/model.py
JackEAllen: Thanks for catching this! It was used at one point, but then I managed to get rid of the need for it after a few revisions so will drop it out now

mantidimaging/gui/windows/spectrum_viewer/model.py
JackEAllen: Looked into this more can i can actually drop this now, made sense before i refactored a lot, but taking a step back, i can actually just use `self.view.shuttercount_norm_enabled()` like you suggest

mantidimaging/gui/windows/main/view.py
JackEAllen: Just added an if statement to check if self.spectrum exists which matches the style lower down in the view.

mantidimaging/gui/windows/spectrum_viewer/view.py
```diff
@@ -108,6 +110,9 @@ def __init__(self, main_window: MainWindowView):
         self.normaliseCheckBox.stateChanged.connect(self.normaliseStackSelector.setEnabled)
         self.normaliseCheckBox.stateChanged.connect(self.presenter.handle_enable_normalised)
         self.normaliseCheckBox.stateChanged.connect(self.presenter.handle_button_enabled)
+        self.normalise_ShutterCount_CheckBox.stateChanged.connect(self.normalise_ShutterCount_CheckBox.setDisabled)
```
samtygier-stfc, line 113: Is this still needed? Looks like `handle_button_enabled` calls `self.normalise_ShutterCount_CheckBox.setEnabled` and will overwrite it.

samtygier-stfc (changes requested):

mantidimaging/gui/windows/spectrum_viewer/view.py
JackEAllen: you're right, this is no longer needed

samtygier-stfc:
A small thing if its easy to fix.
* With a dataset that does not have shutter counts
* Open spectrum viewer, check "Normalise to open beam" and set norm stack
* Check "ShutterCount Correction", the error should now be displayed
* Change the Norm stack to the sample, this should the error for the norm to open beam, hides the shuttercount error, disables the shuttercount checkbox.
* Change the Norm stack back to the open beam
* Now the shutter count check box is enabled and checked, but not showing the error message.

The fix might be to uncheck the shuttercount check box when it is disabled.

samtygier-stfc (approved):
Looking good. I've gone through steps on the issue plus some extra playing around.

JackEAllen:
> Looking good. I've gone through steps on the issue plus some extra playing around.

Thanks for the thorough review 👍

Issue #2213, opened 2024-05-28
Spectrum Viewer: Improve Testing of Stack and Unit Mode Change


JackEAllen:
### Desired Behaviour

A bug was recently found in the spectrum viewer where tof_mode was not set correctly on initialisation when opening with a new dataset which included a spectra file (SEE PR: #2212) Tests did not fail meaning this bug was only found by chance during usage.

We should add some more tests surrounding swapping sample stacks and unit modes where the sample being switched too and from may or may not have a spectra log file loaded.

Example tests we should add:

- [x]  Switching stack with and without data behave correctly
- [x] Opening spectrum viewer first time without spectra data
- [x] Opening spectrum viewer with spectra data
- [x] Switching from a stack with a spectra file to a stack without a spectra file
- [x] Switching from a stack without a spectra file to a stack with a spectra file
- [x] Switching form a stack without a spectra file to another stack without a spectra file
- [x] Switching from a stack with a spectra file to another stack with a spectra file for each ToF mode, ensuring mode stays the same when switching stacks. 
- [ ] ToF mode should never be None. if None, an error should be returned

### Current Behaviour

There are no tests to check swapping between samples and/or unit modes.

samtygier-stfc:
Once #2239 is merged I am happy that we have enough type checking to be confident that the TOF mode should never get set to None.

Issue #2032, opened 2024-02-06
Update pytests for Spectrum Viewer


MikeSullivan7:
### Desired Behaviour

Update pytest scripts for the recently added features, i.e. the adjustable ROI (#2031) and resizable ToF spectrum (#2026)  

### Current Behaviour

No unit tests exist for these features yet

PR #2235, opened 2024-06-21
Stack change units tests added to Spectrum Viewer

MikeSullivan7:
### Issue

Progresses #2213 and #2032

### Description

Added unit tests to check the tof mode is changed correctly when:

- switching from stack with no spectra file to stack with spectra file
- switching from stack with spectra file to stack with spectra file (for tof modes `ToFUnitMode.WAVELENGTH`, `ToFUnitMode.ENERGY`, `ToFUnitMode.TOF_US`)
- switching from stack with no spectra file to stack with no spectra file
- switching from stack with spectra file to stack with no spectra file

These tests check that the tof mode is changed as expected but also if the right click units menu has the correct behaviour (i.e. disabling when there is no spectra file and enabling when there is a spectrum file).

A test has also been added to check that when there is no stack available in the spectrum viewer (i.e. it has been deleted in the Main Window), then the tof mode and right click menu behaves appropriately.

### Acceptance Criteria 

Test that all tests pass with `make check`

### Documentation

Release note

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/test/presenter_test.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/test/presenter_test.py
@@ -316,20 +316,14 @@ class SpectrumViewerWindowPresenterTest(unittest.TestCase):
         self.assertEqual(self.presenter.model.tof_mode, expected_tof_mode)
 
     @parameterized.expand([
-        (None, "Image Index", ToFUnitMode.IMAGE_NUMBER, np.arange(1, 10), [mock.call(False),
-                                                                           mock.call(True)], ToFUnitMode.WAVELENGTH),
-        (np.arange(1, 10), "Wavelength", ToFUnitMode.WAVELENGTH, None, [mock.call(True),
-                                                                        mock.call(False)], ToFUnitMode.IMAGE_NUMBER),
-        (None, "Image Index", ToFUnitMode.IMAGE_NUMBER, None, [mock.call(False),
-                                                               mock.call(False)], ToFUnitMode.IMAGE_NUMBER),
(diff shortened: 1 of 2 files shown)


MikeSullivan7:
Note: I have said that it progresses the issues in case we want to add some more tests to the PR :)

mantidimaging/gui/windows/spectrum_viewer/test/presenter_test.py
```diff
@@ -299,16 +300,60 @@ def test_WHEN_tof_unit_selected_THEN_model_mode_changes(self, mode_text, expecte
...
-        get_stack_time_of_flight.return_value = None
+        get_stack_time_of_flight.return_value = tof_data
         self.view.tof_units_mode = "Wavelength"
         self.presenter.refresh_spectrum_plot = mock.Mock()
         self.presenter.handle_sample_change(uuid.uuid4())
-        self.assertEqual(self.presenter.model.tof_mode, ToFUnitMode.IMAGE_NUMBER)
+        self.assertEqual(self.presenter.model.tof_mode, expected_tof_mode)
+
+    @parameterized.expand([
+        (None, "Image Index", ToFUnitMode.IMAGE_NUMBER, np.arange(1, 10), [mock.call(False),
```
samtygier-stfc, line 319: A small tweak that reduces the wrapping. If you change `[mock.call(False), mock.call(True)],` to `[False, True]`, then you can turn the into `call`s in the method
`expected_calls = [mock.call(b) for b in expected_calls]`

samtygier-stfc (approved):
Looks good and passes tests.

PR #2238, opened 2024-06-26
Type checking in TableModel

samtygier-stfc:
### Issue

Some type checking for #2213

### Description

This adds some type annotations to `TableModel`. This is slightly awkward because the data rows are a heterogenous list, and there is not way to strictly type its elements. We can't switch to a tuple, because they need to be mutable (maybe we can. but it is more code change).

So add `ElementType` and `RowType`, which limit the types, but don't understand the positions. and add an `overload`ed `get_element()` that fixes the type based on the literal value of the column argument.

Also some changes in the view to access things though `get_element()` and ` set_element()`

### Acceptance Criteria 

Existing tests should all still pass
Play around with adding, removing and renaming ROIs

### Documentation

Not needed


JackEAllen (approved):
Nice use of overloading

PR #2239, opened 2024-06-26
Type checking for view, widget and presenter

### LINKED ISSUE #2213: Spectrum Viewer: Improve Testing of Stack and Unit Mode Change

samtygier-stfc:
### Issue

Closes #2213

### Description

Type annotations for spectrum viewer, in the view, widget and presenter

### Acceptance Criteria 
Everything should still pass

### Documentation
Not needed

### DIFF
--- a/mantidimaging/gui/mvp_base/view.py
+++ b/mantidimaging/gui/mvp_base/view.py
@@ -36,21 +36,21 @@ class BaseMainWindowView(QMainWindow):
         self.cleanup()
         super().closeEvent(e)
 
-    def cleanup(self):
+    def cleanup(self) -> None:
         """
         Runs when the window is closed.
         """
         pass
 
-    def show_error_dialog(self, msg: str):
+    def show_error_dialog(self, msg: str) -> None:
         LOG.error(f"show_error_dialog(): {msg}")
         QMessageBox.critical(self, "Error", str(msg))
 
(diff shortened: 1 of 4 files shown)


samtygier-stfc:
With #2237 #2238 and #2239

From
image

To
image

JackEAllen (approved):

PR #2243, opened 2024-07-01
Correct type annotations in the MainWindow, and related classes

ashmeigh:
### Description

Added and corrected type annotations in MainWindow, and related classes.
Ensured that type annotations match the expected return types and parameters.
Fixed type-related issues by converting IDs to appropriate types in MainWindow properties and methods.
Testing

### Testing 

Ran type checks to confirm that type annotations are correctly applied and no type errors remain.
Verified that the updated type annotations do not introduce any runtime errors.

### Acceptance Criteria 

Reviewer should run type checks to ensure all type annotations are correct and no type errors exist.
Confirm that the application functions as expected with the updated type annotations.

### DIFF
--- a/mantidimaging/gui/windows/main/view.py
+++ b/mantidimaging/gui/windows/main/view.py
@@ -114,6 +114,9 @@ class MainWindowView(BaseMainWindowView):
 
     default_theme_enabled: int = 1
 
+    welcome_window: WelcomeScreenPresenter | None = None
+    wizard: WizardPresenter | None = None
+
     def __init__(self, open_dialogs: bool = True):
         super().__init__(None, "gui/ui/main_window.ui")
 
@@ -180,7 +183,7 @@ class MainWindowView(BaseMainWindowView):
             perf_logger.info(f"Mantid Imaging ready in {time.monotonic() - process_start_time}")
         super()._window_ready()
 
-    def setup_shortcuts(self):
+    def setup_shortcuts(self) -> None:
         self.actionLoadDataset.triggered.connect(self.show_image_load_dialog)
         self.actionLoadImages.triggered.connect(self.load_image_stack)
         self.actionLoadNeXusFile.triggered.connect(self.show_nexus_load_dialog)
@@ -208,7 +211,7 @@ class MainWindowView(BaseMainWindowView):
 
         self.model_changed.connect(self.update_shortcuts)
 
-    def populate_image_menu(self):
+    def populate_image_menu(self) -> None:
         self.menuImage.clear()
         current_stack = self.current_showing_stack()
         if current_stack is None:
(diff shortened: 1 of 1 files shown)


mantidimaging/gui/windows/main/view.py
```diff
@@ -559,31 +580,36 @@ def closeEvent(self, event):
             # Ignore the close event, keeping window open
             event.ignore()
 
-    def cleanup(self):
+    def cleanup(self) -> None:
```
ashmeigh, line 583: @samtygier-stfc The shared_array attribute to None directly is not allowed because the shared_array property is read-only. I suggest a clear_shared_array method in the ImageStack class that clears the shared memory.

mantidimaging/gui/windows/main/view.py
```diff
@@ -462,18 +483,18 @@ def show_live_viewer(self, live_data_path: Path) -> None:
             self.live_viewer.show()
 
     @property
-    def stack_list(self):
+    def stack_list(self) -> list[StackVisualiserView]:
         return self.presenter.stack_visualiser_list
 
     @property
-    def strict_dataset_list(self):
```
ashmeigh, line 469: @samtygier-stfc In the MainWindow class, properties such as stack_list and strict_dataset_list are expected to return lists of StackVisualiserView and StrictDataset respectively. However, the corresponding properties in the MainWindowPresenter class return lists of StackId and DatasetId. This mismatch leads to type errors because the expected return types in MainWindow do not match the actual return types from the presenter.

samtygier-stfc: Looks like it should be `list[DatasetId]` here, and that needs to be propagated into a few other places

mantidimaging/gui/windows/main/view.py
samtygier-stfc: May need to use an ignore here:
`stack.shared_array = None # type: ignore # Only happens when cleaning up`

samtygier-stfc (dismissed):
Changes look good.

samtygier-stfc:
Looks like one of the changes breaks a screenshot test.

```
 =================================== FAILURES ===================================
___________ CompareImagesWindowTest.test_compare_images_window_opens ___________

self = <mantidimaging.eyes_tests.compare_images_window_test.CompareImagesWindowTest testMethod=test_compare_images_window_opens>
multi_stack_select = <MagicMock name='MultipleStackSelect' id='140163763533520'>

    @mock.patch("mantidimaging.gui.windows.main.view.MultipleStackSelect")
    def test_compare_images_window_opens(self, multi_stack_select):
        multi_stack_select.return_value.exec.return_value = QDialog.DialogCode.Accepted
        test_stack = generate_images(seed=2021)
        test_stack.name = "Stack 1"
        self.imaging.presenter.get_stack = mock.MagicMock()
        self.imaging.presenter.get_stack.return_value = test_stack
    
        stack_compare = self.imaging.show_stack_select_dialog()
    
>       self.check_target(widget=stack_compare.view)
E       AttributeError: 'NoneType' object has no attribute 'view'

mantidimaging/eyes_tests/compare_images_window_test.py:25: AttributeError
```

MikeSullivan7 (approved):
Everything looks good, all type checks pass. I ran MI through some typical operations and no errors were raised.

samtygier-stfc:
Lets hold merging until after #2266, so that the version is right on the unstable packages

PR #2248, opened 2024-07-02
Don't duplicate ROI names after stack modification

### LINKED ISSUE #2246: Duplicate ROI names
### Summary

It should not be possible to create ROIs with duplicate names. IMAT have found a case where duplicates do happen, which then cause further issues.

### Steps To Reproduce

Open an imagestack in the spectrum viewer.
Create some ROIs
Open the Operations window
Apply an operation to the image stack (e.g. Gaussian)
Back in the spectum view add some more ROIs

### Expected Behaviour

Image in spectrum viewer is updated by operation (e.g. it is blured)
Names for new ROIs are still unique

### Current Behaviour

Naming resets, and duplicate names are created.

### Context
Current main

samtygier-stfc:
### Issue

Closes #2246 

### Description
Move the reset of `_roi_id_counter` so that it only happens when all the ROIs are cleared.
Add a check in the presenter so that this would be caught earlier.
Add some doc strings and rename a method, because it was not clear what the difference was,
Add a test.

### Testing 

Includes a new test

### Acceptance Criteria 

Use the test case on #2246

### Documentation

Release notes

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -59,6 +59,9 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         self.main_window.stack_changed.connect(self.handle_stack_changed)
 
     def handle_stack_changed(self) -> None:
+        """
+        Called when an image stack is modified somewhere else in MI, for example in the operations window
+        """
         if self.current_stack_uuid:
             self.model.set_stack(self.main_window.get_stack(self.current_stack_uuid))
         else:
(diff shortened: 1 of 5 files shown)


mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -318,6 +324,8 @@ def do_add_roi(self) -> None:
         Add a new ROI to the spectrum
         """
         roi_name = self.model.roi_name_generator()
+        if roi_name in self.view.spectrum_widget.roi_dict:
```
JackEAllen, line 327: Can we add a test to check we can raise this exception?

JackEAllen (changes requested):
Looks good! Thanks for adding this in! Seems to resolve the issue for me on Windows, I can't test on ubuntu as my machine lacks the memory to test this. 
Only suggestion is to add a test to validate the exception you added where an ROI name already exists. Though we should in theory never hit this, probably best to test this.

JackEAllen (approved):
Tested locally and happy with this so will approve.

Issue #2250, opened 2024-07-03
Applying Median Filter fails


MikeSullivan7:
### Summary

When applying the Median filter with default parameters on a dataset, the operation fails.

### Steps To Reproduce

1.  Open MI
2. Load in a dataset
3. Open Operations window
4. Select the Median Filter and apply

### Expected Behaviour

The Median filter should be applied to the dataset

### Current Behaviour

The Operation fails

### Failure Logs

```
2024-07-03 16:50:34,507 [mantidimaging.gui.dialogs.async_task.task:L53] ERROR: Failed to execute task: cannot pickle '_thread.lock' object
Traceback (most recent call last):
  File "C:\Users\ddb29996\mantidimaging\mantidimaging\gui\dialogs\async_task\task.py", line 50, in run
    self.result = call_with_known_parameters(self.task_function, **self.kwargs)
  File "C:\Users\ddb29996\mantidimaging\mantidimaging\core\utility\func_call.py", line 12, in call_with_known_parameters
    return func(**ka)
  File "C:\Users\ddb29996\mantidimaging\mantidimaging\gui\windows\operations\model.py", line 118, in apply_to_stacks
    self.apply_to_images(stack, progress=progress)
  File "C:\Users\ddb29996\mantidimaging\mantidimaging\gui\windows\operations\model.py", line 130, in apply_to_images
    exec_func(images)
  File "C:\Users\ddb29996\mantidimaging\mantidimaging\core\operations\median_filter\median_filter.py", line 95, in filter_func
    ps.run_compute_func(MedianFilter.compute_function, data.data.shape[0], data.shared_array, params)
  File "C:\Users\ddb29996\mantidimaging\mantidimaging\core\parallel\shared.py", line 108, in run_compute_func
    pu.run_compute_func_impl(worker_func, num_operations, all_data_in_shared_memory, progress)
  File "C:\Users\ddb29996\mantidimaging\mantidimaging\core\parallel\utility.py", line 137, in run_compute_func_impl
    for _ in pm.pool.imap(worker_func, indices_list, chunksize=calculate_chunksize(pm.cores)):
  File "C:\Users\ddb29996\AppData\Local\miniforge3\envs\mantidimaging-dev\lib\multiprocessing\pool.py", line 873, in next
    raise value
  File "C:\Users\ddb29996\AppData\Local\miniforge3\envs\mantidimaging-dev\lib\multiprocessing\pool.py", line 540, in _handle_tasks
    put(task)
[...]

PR #2253, opened 2024-07-04
Catch errors raised by operations in the system tests

samtygier-stfc:
With a filter for messages that are expected
* Crop coordinates after crop is applied
* Flat fielding causes negatives with test data

### Issue

Can be applied once #2250 is fixed.

### Description

Some of the operations raise exception under normal circumstances (e.g. after running a crop, the same parameters are used to redraw the preview, but fail because the input dimensions are different). So the operations window catches and displays the error. Unfortunately this hides actual errors from the system tests.

This change introduces a small allow list of error messages. Any thing else will count as a test failure.

Add a `raise RuntimeError("hello")` into one of the operations and check that the system-test fails

### Documentation

Release notes

### DIFF
--- a/docs/release_notes/next/dev-2250-catch-operations-errors
+++ b/docs/release_notes/next/dev-2250-catch-operations-errors
@@ -1,0 +1,2 @@
+#2250 : Make systems tests stricter to catch operations errors
+
--- a/mantidimaging/gui/test/gui_system_operations_test.py
+++ b/mantidimaging/gui/test/gui_system_operations_test.py
@@ -46,6 +46,12 @@ OP_LIST = [
     ("Rotate Stack", []),
 ]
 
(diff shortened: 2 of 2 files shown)


samtygier-stfc:
This now catches the error https://github.com/mantidproject/mantidimaging/actions/runs/9790691503/job/27032847828?pr=2253
image

samtygier-stfc:
Rebased on fix so should pass now.

JackEAllen (approved):
Looks good. I have tested system tests and can verify the expected result so i am happy to approve. My only suggestion is that we update our developer documentation to reference why we need to allow some errors and not others, and also why we don't test the stripe removal operations and when we may want to add to both lists when adding future operations or for supporting existing operations. - this can be added as part of a separate PR.

PR #2257, opened 2024-07-08
Don't fail if nothing to clean

samtygier-stfc:
### Issue

When building docs on a new release branch for the first time, the step to clean the existing contents fails, because there is none.

### Description

Add an `|| true`, so that it does not matter if the `git rm` command finds nothing to remove

### Testing 

After merging, the docs should build and publish


samtygier-stfc:
Deploy is failing here because it should not be running. Its running because the branch name starts with `release`, I close this PR and use a better branch name.

Issue #2272, opened 2024-07-10
UnavailableInvalidChannel: HTTP 403 FORBIDDEN for channel intel


samtygier-stfc:
### Summary

When installing the following error is shown.

```
Collecting package metadata (repodata.json): failed

UnavailableInvalidChannel: HTTP 403 FORBIDDEN for channel intel <https://conda.anaconda.org/intel>

The channel is not accessible or is invalid.
```

or 

```
RuntimeError: Multi-download failed. Reason: Transfer finalized, status: 403 [https://anaconda.org/] 6525 bytes
```

This is an issue with the conda repository operated by intel. It has been failing on July 3rd and 10th.

Also being tracked at https://github.com/TomographicImaging/CIL/issues/1856 and https://community.intel.com/t5/Intel-Integrated-Performance/Problems-installing-with-conda-HTTP-403-FORBIDDEN/m-p/1611876/highlight/true#M28732

Current work around is try again tomorrow.

### Steps To Reproduce

mamba env create -f https://raw.githubusercontent.com/mantidproject/mantidimaging/stable/environment.yml

### Expected Behaviour

Successful install

### Current Behaviour

403 Forbidden error

samtygier-stfc:
An official response: https://community.intel.com/t5/Intel-Integrated-Performance/Problems-installing-with-conda-HTTP-403-FORBIDDEN/m-p/1619861/highlight/true#M28787

PR #2275, opened 2024-07-11
Workaround broken intel repo

samtygier-stfc:
### Issue

Workaround for #2272

### Description

The intel conda mirror has been unreliable recently. Switch to a mirror as a workaround until this is resolved.

Use a mirror populated with packages files from our local caches.

### Testing 

If the tests pass then this as worked

### Documentation

Not needed

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -5,7 +5,8 @@ channels:
   - conda-forge
   - ccpi
   - algotom
-  - intel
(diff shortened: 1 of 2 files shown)


MikeSullivan7 (approved):
I rebuilt the dev env with `python setup.py create_dev_env` and then ran all tests and everything passed.

PR #2282, opened 2024-07-16
Type Annotation Enhancement in mantidimaging/gui/ Directory

ashmeigh:
### Description

Added type annotations to various files within the mantidimaging/gui/ directory to enhance type safety and improve code maintainability. These annotations ensure clearer function signatures and better compatibility with Python type checkers like mypy.

### Testing 

Type annotations do not introduce runtime errors.
Functionality across affected modules remains intact.
All affected code paths are adequately covered by existing unit tests.

### Acceptance Criteria 

No linting errors related to type annotations.
No type-related runtime errors occur during testing.
All affected code paths are adequately covered by existing tests. understand this write on for mantidimaging/gui/ in general

### DIFF
--- a/mantidimaging/gui/dialogs/cor_inspection/model.py
+++ b/mantidimaging/gui/dialogs/cor_inspection/model.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 from dataclasses import replace
 from logging import getLogger
 
+import numpy as np
+
 from mantidimaging.core.data import ImageStack
 from mantidimaging.core.reconstruct import get_reconstructor_for
 from mantidimaging.core.utility.data_containers import ScalarCoR, ReconstructionParameters
@@ -26,7 +28,7 @@ class CORInspectionDialogModel:
         # Initial parameters
         if iters_mode:
             self.centre_value: int | float = INIT_ITERS_CENTRE_VALUE
-            self.step = INIT_ITERS_STEP
+            self.step: float = INIT_ITERS_STEP
             self.initial_cor = initial_cor
             self._recon_preview = self._recon_iters_preview
             self._divide_step = self._divide_iters_step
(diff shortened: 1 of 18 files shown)


mantidimaging/gui/dialogs/cor_inspection/model.py
```diff
@@ -4,6 +4,8 @@
 from dataclasses import replace
 from logging import getLogger
 
+import numpy as np
```
MikeSullivan7, line 7: If we are just using `numpy` when doing type annotation then we should probably use 

```suggestion
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    import numpy as np
```
to prevent from unnecessary imports

mantidimaging/gui/utility/qt_helpers.py
```diff
@@ -46,20 +46,20 @@ def __exit__(self, *args):
...
+def select_directory(field: QWidget, caption: str) -> None:
     assert isinstance(field, QLineEdit), (f"The passed object is of type {type(field)}. This function only works with "
                                           "QLineEdit")
 
     # open file dialog and set the text if file is selected
     field.setText(QFileDialog.getExistingDirectory(caption=caption))
 
 
-def get_value_from_qwidget(widget: QWidget):
+def get_value_from_qwidget(widget: QWidget) -> Any:
```
MikeSullivan7, line 62: its a bit verbose but we could narrow this down from `Any` with 

```suggestion
def get_value_from_qwidget(widget: QWidget) -> str | float | bool | None:
```

MikeSullivan7: Might need to add `return None` at the end of the function to help out MyPy

MikeSullivan7 (changes requested):
Ran tests and loaded data into MI, ran though some usual dataflow by performing operations and reconstructions. Everything functions as usual with no errors. MyPy is happy with no complaints.
Just some very small changes but after that I'm happy to approve.

MikeSullivan7 (approved):
Thanks for the changes :)

Issue #2199, opened 2024-05-09
Unify Datasets


samtygier-stfc:
Can we remove the distinction between `MixedDataset` and `StrictDataset`

PR #2288, opened 2024-07-22
Unify datasets part 1

samtygier-stfc:
### Issue

Work on #2199

### Description

This is the first step in unifying `MixedDataset` and `StrictDataset`. This first step moves all the implementation of `MixedDataset` into `BaseDataset`. 

This required some changes that result in a small changes to the rest of the codebase
Forcing constructor arguments to named, because the current classes use different argument ordering
Forcing access to the recons through a property.

### Testing 
I have moved the  `MixedDataset` and `StrictDataset` tests into core.
I have added a DatasetTest, currently for BaseDataset. This will duplicate tests from the other test classes, and eventually replace them.
I have left the  `MixedDataset` and `StrictDataset` alone (apart from small API changes), to ensure that the classes behaviour is not changed.

### Acceptance Criteria 

Test loading a dataset and an image stack

### Documentation

Not needed yet


mantidimaging/core/data/dataset.py
```diff
@@ -25,11 +25,14 @@ def remove_nones(image_stacks: list[ImageStack | None]) -> list[ImageStack]:
...
 
-    def __init__(self, *, name: str = ""):
+    def __init__(self, *, name: str = "", stacks: list[ImageStack] | None = None) -> None:
         self._id: uuid.UUID = uuid.uuid4()
-        self._recons = ReconList()
         self.name = name
+
+        self._recons = ReconList()
         self._sinograms: ImageStack | None = None
+        stacks = [] if stacks is None else stacks
```
JackEAllen, line 34: If i'm understanding this correctly, could we set the default value of stacks in the initialiser as an empty list instead of None to save us needing to do this if no argument is passed?

JackEAllen: If there is a chance we could be calling with None, does this need to be the case or can it be avoided to allow for just setting a default return value of an empty list?

samtygier-stfc: We can't use an empty list as the default argument as it is mutable and would end up getting shared by multiple instances. ( https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments )

mantidimaging/core/data/dataset.py
```diff
@@ -45,13 +48,17 @@ def sinograms(self, sino: ImageStack | None) -> None:
 
     @property
     def all(self) -> list[ImageStack]:
-        return self.recons.stacks + remove_nones([self._sinograms])
+        return self.recons.stacks + self._stacks + remove_nones([self._sinograms])
 
     def delete_stack(self, images_id: uuid.UUID) -> None:
         for recon in self.recons:
```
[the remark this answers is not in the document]
samtygier-stfc: There is already a 'return', so the loop will stop once a match is found.

This will all get a cleanup once the rest of the merging happens

JackEAllen (approved):
Having now tested a few times and gone over the code a few times, I'm happy to approve this. If you feel that a second courtesy review is needed, I would also agree with this given the code we are touching, otherwise can just merge in :)

samtygier-stfc:
I think its all going to get a bunch more testing over the next few PRs, so I'd suggest merging now

Issue #2267, opened 2024-07-10
Update Python to 3.12


samtygier-stfc:
### Desired Behaviour

Update to the latest python version

MikeSullivan7:
Note: Upgrading to `python 3.11` introduces the following test failures

```
============================================================ FAILURES =============================================================
_____________________________________________ CommandLineArgumentsTest.test_tracking ______________________________________________
[gw6] win32 -- Python 3.11.9 C:\Users\ddb29996\AppData\Local\miniforge3\envs\mantidimaging-dev\python.exe

self = <mantidimaging.core.utility.test.leak_tracker_test.CommandLineArgumentsTest testMethod=test_tracking>

    def test_tracking(self):
        self.assertEqual(self.leak_tracker.count(), 0)
        ob1 = ExampleObject()
        self.leak_tracker.add(ob1, "created_in_test_tracking")
        container = ExampleContainer(ob1)  # noqa: F841

        self.assertEqual(self.leak_tracker.count(), 1)
        del ob1
        self.assertEqual(self.leak_tracker.count(), 1)

        def check_output(message, **kwargs):
            track_output = StringIO()
            self.leak_tracker.pretty_print(output=track_output, **kwargs)
            track_output_value = track_output.getvalue()
            self.assertIn(message, track_output_value)

        # check that the object type and message are in the output
        check_output("leak_tracker_test.ExampleObject")
        check_output("created_in_test_tracking")

        # Check that this file and the line of code are listed in the output
        check_output(__file__, debug_init=True)
        check_output('self.leak_tracker.add(ob1, "created_in_test_tracking")', debug_init=True)

        # Check that container holding the reference is listed in the output
        check_output("leak_tracker_test.ExampleContainer", debug_owners=True)
>       check_output("held_reference", debug_owners=True)

mantidimaging\core\utility\test\leak_tracker_test.py:87:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
mantidimaging\core\utility\test\leak_tracker_test.py:75: in check_output
    self.assertIn(message, track_output_value)
[...]

PR #2291, opened 2024-07-23
Show attribute names in leak tracker

samtygier-stfc:
Needed for 3.11 classes with lazy dictionaries

### Issue
Needed for python 3.11 update ( see #2267 ), but works fine with 3.10

### Description

In python 3.11 classes only create the dictionary object if needed ( https://github.com/faster-cpython/ideas/issues/72 ).

When finding owners for objects in the leak tracker, and having an object held by a class like:
```python
class ExampleObject:
    pass

class ExampleContainer:
    def __init__(self, obj):
        self.held_reference = obj

ob1 = ExampleObject()
container = ExampleContainer(ob1)
```
previously we `gc.get_referrers(ob1)` would give us the `dict` (i.e. `container.__dict__`) and from that we would find the key. Now `gc.get_referrers(ob1)` gives us the instance, and it is not clear which attribute it is.

The change add a case for a class, and adds the attribute name to `extra_info`. The act of looking at the `__dict__` causes it to be created, so that we can get the name.

Also fix the test class name.

### Testing 

This fixes the leaktracker test when running with python 3.11 and above

### Acceptance Criteria 

Tests should pass

### Documentation

release notes

### DIFF
--- a/docs/release_notes/next/dev-2267-leak-tracker
+++ b/docs/release_notes/next/dev-2267-leak-tracker
@@ -1,0 +1,1 @@
+#2267: Show attribute names in leak tracker
--- a/mantidimaging/core/utility/leak_tracker.py
+++ b/mantidimaging/core/utility/leak_tracker.py
@@ -25,6 +25,11 @@ def obj_to_string(obj, relative=None) -> str:
         keys = {k for k, v in obj.items() if v is relative}
         extra_info += f" keys={keys}"
 
+    if hasattr(obj, '__dict__') and relative is not None:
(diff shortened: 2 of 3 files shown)


MikeSullivan7:
Using `Python 3.12`, the error relating to the `leak tracker` has been fixed and isn't showing anymore. I do get errors relating to `mantidimaging/core/data/test/image_stack_test.py::ImageStackTest::test_cant_change_id ` and `mantidimaging/gui/windows/main/test/strictdataset_test.py::StrictDatasetTest::test_cant_change_dataset_id ` but these are fixed by #2292

MikeSullivan7 (approved):
See comment above!

PR #2293, opened 2024-07-24
Unify datasets part 2

samtygier-stfc:
### Issue
Work on #2199.

~~Needs rebase after #2288~~

### Description

This PR forces named parameters for 'StrictDataset'. That touches many files, so I'll keep it in its own PR.

### Testing 

Only test changes are to add the parameter names

### Acceptance Criteria 

Test loading a dataset and an image stack

### Documentation

Not needed yet

### DIFF
--- a/mantidimaging/gui/windows/main/test/presenter_test.py
+++ b/mantidimaging/gui/windows/main/test/presenter_test.py
@@ -489,9 +489,9 @@ class MainWindowPresenterTest(unittest.TestCase):
         self.view.model_changed.emit.assert_called_once()
 
     def test_dataset_list(self):
-        dataset_1 = StrictDataset(generate_images())
+        dataset_1 = StrictDataset(sample=generate_images())
         dataset_1.name = "dataset-1"
-        dataset_2 = StrictDataset(generate_images())
+        dataset_2 = StrictDataset(sample=generate_images())
         dataset_2.name = "dataset-2"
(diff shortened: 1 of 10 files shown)


mantidimaging/gui/windows/main/test/presenter_test.py
```diff
@@ -565,7 +565,12 @@ def test_tabify_stack_window_to_item_in_list(self):
         self.view.tabifyDockWidget.assert_called_once_with(other_stack, new_stack)
 
     def test_create_strict_dataset_tree_view_items(self):
-        dataset = StrictDataset(*generate_images_with_filenames(5))
+        stacks = generate_images_with_filenames(5)
+        dataset = StrictDataset(sample=stacks[0],
```
JackEAllen, line 569: NITPICK, not that this really need condensing, but we could maybe get away with unpacking this instead:
`dataset = StrictDataset(sample=stacks[0], *stacks[1:5])`

samtygier-stfc: At least during the migration I want to force named arguments to avoid any ambiguity with the constructors.

JackEAllen: Definitely! I appreciate you adding * to do this to the initialisers

JackEAllen (approved):
Happy to approve this after testing. Apologies for taking a while to get around to it!

PR #1967, opened 2023-11-01
Move to miniforge

samtygier-stfc:
### Issue
Closes #1949

### Description

Previously we used Mambaforge to create a conda setup which contained the `mamba` command. Recently the conda-forge devs have unified their Miniforge and Mambaforge installers, so the are now identical.

This PR:
* updates the docs to references Miniforge
* use Miniforge in the github workflows
* swap some old uses of the `conda` command to `mamba` (there are only a small number of cases where `conda` is still needed)

### Testing & Acceptance Criteria 

Check that the doc changes make sense

Check that the automated tests have successfully completed

### Documentation

Not needed

PR #2344, opened 2024-10-15
Switch from Mambaforge to Miniforge in GitHub Actions workflow

ashmeigh:
### Issue

Closes #1967

### Description

This PR resolves the GitHub Actions warning about the deprecation of Mambaforge by switching to Miniforge. It updates the environment setup to use Miniforge while keeping Mamba for faster environment management. The Coveralls step is also restored for test coverage reporting.

### Testing 

Verified that the workflow runs without Mambaforge warnings.
Confirmed test coverage reports are uploaded to Coveralls.
Ensured environment setup works with and without cache.

### Acceptance Criteria 

Workflow should run without Mambaforge warnings.
Caching and Coveralls should function correctly.

### Documentation

No user-facing changes.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -27,7 +27,7 @@ jobs:
           echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
         shell: bash
 
-      - name: Cache Mambaforge and Pip packages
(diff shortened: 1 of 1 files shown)


samtygier-stfc (dismissed):
Looks like that fixed it.

I'll squash to avoid the spacing change

samtygier-stfc (approved):

PR #2346, opened 2024-10-16
Unify datasets part 8 

samtygier-stfc:
### Issue
More on #2199
~~Needs rebasing after #2341~~

### Description

Close follow up of #2341, continues the work of unifying and simplifying the MainWindowPresenter.

Refactor `_add_images_to_existing_dataset` to `handle_add_images_to_existing_dataset_from_dialog`, and split out useful parts to `add_images_to_existing_dataset`.

Refactor methods for adding stacks to use common code
`_add_recon_to_dataset`, `add_sinograms_to_dataset_and_update_view`, `add_180_deg_file_to_dataset`, `add_alternative_180_if_required`.

Remove several methods that are not longer needed.

### Testing 

Various tests were removed or updated

### Acceptance Criteria 

Test loading datasets, adding additional images, adding a 180, converting projections to sinograms
### Documentation

not yet

### DIFF
--- a/mantidimaging/core/data/dataset.py
+++ b/mantidimaging/core/data/dataset.py
@@ -184,6 +184,8 @@ class Dataset:
             self.add_recon(image_stack)
         elif file_type_name == "IMAGES":
             self.stacks.append(image_stack)
+        elif file_type_name == "SINOGRAMS":
+            self._sinograms = image_stack
         else:
             file_type = getattr(FILE_TYPES, file_type_name)
             self.set_stack(file_type, image_stack)
--- a/mantidimaging/core/data/dataset.py
+++ b/mantidimaging/core/data/dataset.py
@@ -220,15 +220,14 @@ def _get_stack_data_type(stack_id: uuid.UUID, dataset: Dataset) -> str:
         return "Recon"
     if stack_id in [stack.id for stack in dataset._stacks]:
         return "Images"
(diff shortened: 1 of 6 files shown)


mantidimaging/core/data/dataset.py
```diff
@@ -218,15 +220,14 @@ def _get_stack_data_type(stack_id: uuid.UUID, dataset: Dataset) -> str:
         return "Recon"
     if stack_id in [stack.id for stack in dataset._stacks]:
```
JackEAllen, line 221: NITPICK: While we're touching this area of code, we should probably refactor how we access the `_stacks` to go through `stacks` to maintain encapsulation, understanding that is is a subclass of `Dataset`.

JackEAllen (approved):
The changes made in this PR as part of the dataset unification look good. I've tested manually with a few different datasets including adding additional images, 180 deg and converting to and from sinogram. All seems to works as expected. 
I have left a NITPICK relating to access of `_stacks` directly within `_get_stack_data_type()`, but otherwise this PR is good to merge. I don't feel strongly towards the suggestion so will approve for now and leave the decision up to you 🙂

mantidimaging/core/data/dataset.py
samtygier-stfc: I might save that for a future unification of how different stack types are handled.

PR #2349, opened 2024-10-22
2348 Pyright

### LINKED ISSUE #2348: Test Pyright Implementation
Add Pyright to list of dependencies to see what is picked up that MyPy may have missed and resolve any errors picked up.

JackEAllen:
### Issue
Closes #2348 

### Description

* Add pyright as a dependancy to the project
* Add Pyright to conda and windows workflows
* Add Pyright pre-commit check to project
* Configure basic ruleset so Pyright does not initially overwhelm us with too many warnings
* Resolve some of the basic Pyright issues blagged up in the basic configuration to get us started with refactorign the project to be Pyright compliant.

This PR once merged in, provides a good starting point to begin disabling rules currently set to false within `pyproject.toml` to eventually allow for the typeChecking mode to be increased from "basic" to "Standard" and eventually to "strict".

### Testing 

* Tested workflows fail if Pyright issues exist within the project
* Tested pre-commit fails if Pyright issues exist in any staged files.
* Tested Mantid Imaging basic Workflows
* Ran unit tests

### Acceptance Criteria 

- [ ] Rebuilding a developer environmnet inlcudes Pyright in the list of dependancies
- [ ] Conda and Windows Pyright workflows will fail if a Pyright issue exists in the project
- [ ] Pyright pre-commit check will block a commit if any of the staged files do not pass Pyright checks
- [ ] Mantid Imaging still performs as expected:
    - [ ] Run through operations, reconstructions and test spectrum viewer all still work as expected.

### Documentation

See `docs/release_notes/next/dev-2348-Pyright`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,16 @@ repos:
         files: ^mantidimaging/
         args: [--ignore-missing-imports]
         additional_dependencies: [types-docutils, types-PyYAML, types-requests]
+-   repo: local
+    hooks:
+    -   id: pyright
+        name: pyright
+        description: Checks that all files pass pyright checks
+        entry: pyright
+        language: system
+        types: [python]
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.3.3
     hooks:
     -   id: ruff
+
--- a/mantidimaging/core/utility/histogram.py
+++ b/mantidimaging/core/utility/histogram.py
@@ -17,6 +17,7 @@ def set_histogram_log_scale(histogram: HistogramLUTItem) -> None:
     :param histogram: The HistogramLUTItem of an image.
     """
     x_data, y_data = histogram.plot.getData()
+    assert isinstance(x_data, np.ndarray) and isinstance(y_data, np.ndarray)
     histogram.plot.setData(x_data, np.log(y_data + 1))
 
 
--- a/mantidimaging/core/utility/memory_usage.py
+++ b/mantidimaging/core/utility/memory_usage.py
@@ -28,26 +28,20 @@ def system_free_memory():
     return Value(meminfo.available - meminfo.total * MEMORY_CAP_PERCENTAGE)
 
 
-def get_memory_usage_linux(kb=False, mb=False):
+def get_memory_usage_linux() -> float:
     """
-    :param kb: Return the value in Kilobytes
-    :param mb: Return the value in Megabytes
+    Get the memory usage of the system in bytes
     """
     import psutil
 
     meminfo = psutil.virtual_memory()
-    tuple_to_return = ()  # start with empty tuple
     # meminfo.used gives the size in bytes
-    if kb:
-        tuple_to_return += (meminfo.used / 1024, )
-
-    if mb:
-        tuple_to_return += (meminfo.used / 1024 / 1024, )
-    return tuple_to_return
+    return meminfo.used
 
 
 def get_memory_usage_linux_str():
-    memory_in_kbs, memory_in_mbs = get_memory_usage_linux(kb=True, mb=True)
+    memory_in_kbs = get_memory_usage_linux() / 1024
+    memory_in_mbs = memory_in_kbs / 1024
     # handle caching
     memory_string = f"{memory_in_kbs} KB, {memory_in_mbs} MB"
 
--- a/mantidimaging/eyes_tests/base_eyes.py
+++ b/mantidimaging/eyes_tests/base_eyes.py
@@ -57,6 +57,7 @@ QApplication.setFont(QFont("Sans Serif", 10))
 @start_qapplication
 class BaseEyesTest(unittest.TestCase):
     eyes_manager: EyesManager
+    app: QApplication
 
     @classmethod
     def setUpClass(cls) -> None:
@@ -130,8 +131,8 @@ class BaseEyesTest(unittest.TestCase):
 
         return new_dataset
 
-    def _get_top_level_widget(cls, widget_type):
-        for widget in cls.app.topLevelWidgets():
+    def _get_top_level_widget(self, widget_type):
+        for widget in self.app.topLevelWidgets():
             if isinstance(widget, widget_type):
                 return widget
         raise ValueError(f"Could not find top level widget of type {widget_type.__name__}")
--- a/mantidimaging/gui/widgets/mi_image_view/view.py
+++ b/mantidimaging/gui/widgets/mi_image_view/view.py
@@ -19,9 +19,10 @@ from mantidimaging.gui.widgets.auto_colour_menu.auto_color_menu import AutoColor
 from mantidimaging.gui.widgets.mi_image_view.presenter import MIImagePresenter
 from mantidimaging.gui.widgets.bad_data_overlay.bad_data_overlay import BadDataOverlay
 
+import numpy as np
+
 if TYPE_CHECKING:
     from pyqtgraph import HistogramLUTItem
-    import numpy as np
     from mantidimaging.core.utility.data_containers import ProjectionAngles
 
 
@@ -202,6 +203,7 @@ class MIImageView(ImageView, BadDataOverlay, AutoColorMenu):
         self._refresh_message()
 
     def _update_roi_region_avg(self) -> SensibleROI | None:
+        assert self.image is not None
         if self.image.ndim != 3:
             return None
         roi_pos, roi_size = self.get_roi()
@@ -264,7 +266,7 @@ class MIImageView(ImageView, BadDataOverlay, AutoColorMenu):
         if self.image.ndim == 3:
             x = clip(pt.x, 0, self.image.shape[2] - 1)
             y = clip(pt.y, 0, self.image.shape[1] - 1)
-            value = self.image[self.currentIndex, y, x]
+            value = self.image[self.currentIndex, y, x] if self.image is not None else 0
             msg = f"x={y}, y={x}, z={self.currentIndex}, value={value :.6f}"
             if self.angles:
                 angle = degrees(self.angles.value[self.currentIndex])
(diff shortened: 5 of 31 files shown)


mantidimaging/gui/widgets/mi_image_view/view.py
```diff
@@ -211,12 +212,12 @@ def _update_roi_region_avg(self) -> SensibleROI | None:
 
         if self.roi.isVisible():
             z_value = int(self.timeLine.value())
-            mean_val = self.image[z_value, top:bottom, left:right].mean()
+            mean_val = self.image[z_value, top:bottom, left:right].mean() if self.image is not None else 0
```
samtygier-stfc, line 215: Interesting that it warns about this, but not line 206. Would it is work to put a `assert self.image is not None` at the top of the method? or can this be called without the image being set?

mantidimaging/core/utility/memory_usage.py
```diff
@@ -28,22 +28,25 @@ def mb(self):
...
-        tuple_to_return += (meminfo.used / 1024 / 1024, )
-    return tuple_to_return
+    if kb and mb:
+        return (meminfo.used / 1024, meminfo.used / 1024 / 1024)
+    elif kb:
+        return (meminfo.used / 1024, 0.0)
+    elif mb:
+        return (0.0, meminfo.used / 1024 / 1024)
+    else:
+        return (0.0, 0.0)
```
samtygier-stfc, line 49: I think the original function here is over complex. Maybe it should just return a value in bytes, and the place using it can convert to kb or mb as needed.

mantidimaging/eyes_tests/base_eyes.py
```diff
@@ -64,6 +64,7 @@ def setUpClass(cls) -> None:
         cls.eyes_manager.set_batch(APPLITOOLS_BATCH_ID)
 
     def setUp(self):
+        self.app = QApplication.instance() or QApplication([])
```
samtygier-stfc, line 67: This might duplicate what is happening in `@start_application`. Maybe it could be a class level type annotation, near line 59.
`app: QApplication`

mantidimaging/core/utility/memory_usage.py
JackEAllen: Seems to only be called in one place in the same class by `get_memory_usage_linux_str()`so I've simplified to only return bytes and allow `get_memory_usage_linux_str()` to convert into kb and mb

.pre-commit-config.yaml
```diff
@@ -23,7 +23,16 @@ repos:
         files: ^mantidimaging/
         args: [--ignore-missing-imports]
         additional_dependencies: [types-docutils, types-PyYAML, types-requests]
+-   repo: local
+    hooks:
+    -   id: pyright
+        name: pyright
+        description: Checks that all files paas pyright checks
```
samtygier-stfc, line 30: `paas` -> `pass`

JackEAllen: Thanks for spotting this!

mantidimaging/gui/windows/recon/presenter.py
```diff
@@ -396,7 +396,7 @@ def completed(task: TaskWorkerThread) -> None:
                     f"\n\n Suggestion: Use crop coordinates to resize the 180 degree projection to "
                     f"({selected_stack.height}, {selected_stack.width})")
             else:
-                cor, tilt = task.result
+                cor, tilt = task.result if task.result is not None else (None, None)
```
samtygier-stfc, line 400: I'd prefer an assert here. We'd only reach this if both `task.result` and `task.error` were `None`, which would mean something bad was wrong and we'd want to raise that to an error message.

Would it work to restructure the `if` as
```
if task.error is not None:
  ...
elif task.result is not None:
  ...
else:
  raise AssertionError("task in inconsistent state")
```
It would be a bit more readable maybe for humans and type checkers.

mantidimaging/gui/widgets/mi_image_view/view.py
```diff
@@ -130,7 +130,7 @@ def histogram(self) -> HistogramLUTItem:
 
     @property
     def image_data(self) -> np.ndarray:
-        return self.image
+        return self.image if self.image is not None else np.array([])
```
samtygier-stfc, line 133: I think it would be better to change the type signature to `np.ndarray|None` than the behaviour.

I found `ImageViewLineROI._image_data_exists()`, uses this to check if there is an image.

mantidimaging/gui/windows/nexus_load_dialog/presenter.py
```diff
@@ -133,15 +133,17 @@ def _read_rotation_angles(self, image_key: int, before: bool | None = None) -> n
         """
         assert self.image_key_dataset is not None and self.rotation_angles is not None
         if before is None:
-            rotation_angles = self.rotation_angles[np.where(self.image_key_dataset[...] == image_key)]
+            image_key_array = np.array(self.image_key_dataset)
+            rotation_angles = np.array(self.rotation_angles)[np.where(image_key_array == image_key)]
```
samtygier-stfc, line 137: Can these be avoided by setting the types as `np.ndarray|None` in the constructor? That would allow the type check to catch anything assigning a different type to them.

mantidimaging/gui/windows/nexus_load_dialog/presenter.py
```diff
@@ -220,7 +222,10 @@ def _look_for_recon_entries(self):
         assert self.nexus_file is not None
         for key in self.nexus_file.keys():
             if DEFINITION in self.nexus_file[key].keys():
-                if np.array(self.nexus_file[key][DEFINITION]).tobytes().decode("utf-8") == NXTOMOPROC:
+                definition_data = self.nexus_file[key].get(DEFINITION)
+                if definition_data is None:
+                    continue
```
samtygier-stfc, line 227: I think we could use cast here. We know the `DEFINITION` is there, and I assume from the nexus spec it must be a string. Otherwise I'd rather have an assertion so that we get an error rather than silently continuing.

mantidimaging/gui/windows/nexus_load_dialog/presenter.py
```diff
@@ -274,12 +279,15 @@ def _get_images(self, image_key_number: ImageKeys, before: bool | None = None) -
...
+            indices = np.array(self.image_key_dataset) == image_key_number.value
         else:
             if before:
-                indices = self.image_key_dataset[:self.image_key_dataset.size // 2] == image_key_number.value
+                image_key_array = np.array(self.image_key_dataset)
+                indices = image_key_array[:image_key_array.size // 2] == image_key_number.value
             else:
+                # indices = np.array(self.image_key_dataset) == image_key_number.value
                 indices = self.image_key_dataset[:] == image_key_number.value
+
```
samtygier-stfc, line 290: Might not need these if the type of `self.image_key_dataset` is specified in the constructor.

mantidimaging/gui/widgets/mi_mini_image_view/view.py
```diff
@@ -72,15 +72,15 @@ def histogram(self) -> HistogramLUTItem:
...
+        return self.hist.region.getRegion()  # type: ignore
 
     @histogram_region.setter
     def histogram_region(self, new_region: tuple[int | list[int], int | list[int]]) -> None:
         self.hist.region.setRegion(new_region)
 
     @property
     def image_data(self) -> np.ndarray:
-        return self.im.image
+        return self.im.image  # type: ignore
```
samtygier-stfc, line 83: We should change the signature to that callers have to handle the None case.

mantidimaging/core/utility/histogram.py
```diff
@@ -17,7 +17,8 @@ def set_histogram_log_scale(histogram: HistogramLUTItem) -> None:
     :param histogram: The HistogramLUTItem of an image.
     """
     x_data, y_data = histogram.plot.getData()
-    histogram.plot.setData(x_data, np.log(y_data + 1))
+    if y_data is not None:
+        histogram.plot.setData(x_data, np.log(y_data + 1))
```
samtygier-stfc, line 21: Maybe `assert isinstance(x_data, np.ndarray) and isinstance(y_data, np.ndarray)`. To be more specific and raise an error if assumption is not true.

mantidimaging/gui/widgets/mi_image_view/view.py
JackEAllen: I can change the type. signature only if I also change the type signature of `image_data` in `AutoColorMenu` otherwise `image_data` will be incompatible with the supertype `AutoColorMenu`

samtygier-stfc: I think that's the right thing to do.

mantidimaging/gui/widgets/mi_mini_image_view/view.py
JackEAllen: Played around with this and have set the return to `np.ndarray | None` and added an assertion in the spectrum viewer which relied on `self.im.image` not being None. If I threw in an assert in `image_data` instead of changing the return type, We get a lot of errors due to there being instances where `self.im.image` is None.

mantidimaging/gui/windows/nexus_load_dialog/presenter.py
```diff
@@ -131,18 +131,20 @@ def _read_rotation_angles(self, image_key: int, before: bool | None = None) -> n
...
+                image_key_array = np.array(self.image_key_dataset)[:first_sample_image_index]
+                rotation_angles = np.array(
+                    self.rotation_angles)[:first_sample_image_index][np.where(image_key_array == image_key)]
             else:
-                rotation_angles = self.rotation_angles[first_sample_image_index:][np.where(
-                    self.image_key_dataset[first_sample_image_index:] == image_key)]
+                rotation_angles = np.array(self.rotation_angles)[first_sample_image_index:][np.where(
+                    np.array(self.image_key_dataset)[first_sample_image_index:] == image_key)]
 
+        rotation_angles = np.array(rotation_angles)
```
samtygier-stfc, line 147: Are these `np.array()`s still needed? between the type annotation and the assert I think that should be safe.

JackEAllen: These can now be removed 👍

samtygier-stfc (approved):
Thanks. A good set of improvements.

PR #2398, opened 2024-11-13
Multiple Live Viewers can be opened simultaneously

### LINKED ISSUE #2361: Allow opening multiple live viewers
IMAT's usage of the live viewer ecnapsulates two different workflows:

Initial setup using genie python snap method which places a singular image in a dedicated snap directory which will always have the same path
Sample scan/running experiment where directory can change and many subfolders can be generated
IMAT have voiced that they would they would like to easily view two directories at a time, either within the same window or by opening two separate live viewer windows.

Allowing for two separate live viewer windows would be the easiest to implement

MikeSullivan7:
### Issue

Closes #2361

### Description

Removed the reraising of the currently opened Live Viewer to allow a second (or more) Live Viewer to be opened, each with a specified path.

### Testing 

`make check`
`make system-test`

### Acceptance Criteria 

Open multiple Live Viewer windows and point them to different directories. Check that the LVs behave independently of each other and behave as expected. You should be able to use multiple terminals with the `simulate_live_data.py` script to check that MI can process the various LVs simultaneously (see screenshot below)

### Documentation

Add release note

image

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -28,9 +28,9 @@ class LiveViewerWindowView(BaseMainWindowView):
 
     def __init__(self, main_window: MainWindowView, live_dir_path: Path) -> None:
         super().__init__(None, 'gui/ui/live_viewer_window.ui')
-        self.setWindowTitle("Mantid Imaging - Live Viewer")
         self.main_window = main_window
         self.path = live_dir_path
+        self.setWindowTitle(f"Mantid Imaging - Live Viewer - {str(self.path)}")
         self.presenter = LiveViewerWindowPresenter(self, main_window)
         self.live_viewer = LiveViewWidget()
         self.imageLayout.addWidget(self.live_viewer)
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -88,11 +88,8 @@ class LiveViewerWindowView(BaseMainWindowView):
 
     def closeEvent(self, e) -> None:
         """Close the window and remove it from the main window list"""
-        for live_viewer in self.main_window.live_viewer_list:
-            if live_viewer.path == self.path:
-                if live_viewer.presenter:
-                    live_viewer.presenter.close()
-                live_viewer.close()
+        self.main_window.live_viewer_list.remove(self)
+        self.presenter.close()
         self.live_viewer.handle_deleted()
         super().closeEvent(e)
         self.presenter = None  # type: ignore # View instance to be destroyed -type can be inconsistent
--- a/mantidimaging/gui/windows/live_viewer/view.py
+++ b/mantidimaging/gui/windows/live_viewer/view.py
@@ -88,8 +88,11 @@ class LiveViewerWindowView(BaseMainWindowView):
 
     def closeEvent(self, e) -> None:
         """Close the window and remove it from the main window list"""
-        self.main_window.live_viewer = None
-        self.presenter.close()
+        for live_viewer in self.main_window.live_viewer_list:
+            if live_viewer.path == self.path:
+                if live_viewer.presenter:
+                    live_viewer.presenter.close()
+                live_viewer.close()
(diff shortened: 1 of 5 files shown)


mantidimaging/gui/windows/main/view.py
```diff
@@ -474,13 +474,10 @@ def live_view_choose_directory(self) -> None:
...
-            self.live_viewer = LiveViewerWindowView(self, live_data_path)
-            self.live_viewer.show()
-        else:
-            self.live_viewer.activateWindow()
-            self.live_viewer.raise_()
-            self.live_viewer.show()
+        self.live_viewer = LiveViewerWindowView(self, live_data_path)
+        self.live_viewer.activateWindow()
+        self.live_viewer.raise_()
+        self.live_viewer.show()
```
samtygier-stfc, line 480: This will need to be made into a list of liveviewers, so that `closeEvent()` can close all of them. Otherwise there are issues if you close the main window with multiple live viewers open.

Also in `LiveViewerWindowView.closeEvent()` it will need to remove itself from this list.

samtygier-stfc (changes requested):

mantidimaging/gui/windows/live_viewer/view.py
```diff
@@ -88,8 +88,11 @@ def set_image_index(self, index: int) -> None:
 
     def closeEvent(self, e) -> None:
         """Close the window and remove it from the main window list"""
-        self.main_window.live_viewer = None
-        self.presenter.close()
+        for live_viewer in self.main_window.live_viewer_list:
+            if live_viewer.path == self.path:
+                if live_viewer.presenter:
+                    live_viewer.presenter.close()
+                live_viewer.close()
```
samtygier-stfc, line 95: Something like:
```python
self.main_window.live_viewer_list.remove[self]
self.presenter.close()
```
should work here.

mantidimaging/gui/windows/main/view.py
```diff
@@ -474,13 +474,11 @@ def live_view_choose_directory(self) -> None:
...
-            self.live_viewer.show()
-        else:
-            self.live_viewer.activateWindow()
-            self.live_viewer.raise_()
-            self.live_viewer.show()
+        live_viewer = LiveViewerWindowView(self, live_data_path)
+        self.live_viewer_list.append(live_viewer)
+        self.live_viewer_list[-1].activateWindow()
+        self.live_viewer_list[-1].raise_()
+        self.live_viewer_list[-1].show()
```
samtygier-stfc, line 481: I don't think we need these, as we no longer have the case of raising an existing live window

mantidimaging/gui/windows/main/view.py
```diff
@@ -474,13 +474,11 @@ def live_view_choose_directory(self) -> None:
...
-            self.live_viewer = LiveViewerWindowView(self, live_data_path)
-            self.live_viewer.show()
-        else:
-            self.live_viewer.activateWindow()
-            self.live_viewer.raise_()
-            self.live_viewer.show()
+        live_viewer = LiveViewerWindowView(self, live_data_path)
+        self.live_viewer_list.append(live_viewer)
+        # self.live_viewer_list[-1].activateWindow()
+        # self.live_viewer_list[-1].raise_()
```
samtygier-stfc, line 480: The commented code can be removed.

mantidimaging/gui/windows/main/view.py
```diff
@@ -563,8 +561,9 @@ def closeEvent(self, event) -> None:
             # Close additional windows which do not have the MainWindow as parent
             if self.recon:
                 self.recon.close()
-            if self.live_viewer:
-                self.live_viewer.close()
+            if self.live_viewer_list:
+                while self.live_viewer_list:
+                    self.live_viewer_list[-1].close()
```
samtygier-stfc, line 566: You could remove the `if self.live_viewer_list:`, the while will do nothing if the list is empty.

samtygier-stfc (changes requested):
Looking good. A couple of small tweaks.

samtygier-stfc (approved):
Looks good. Tested locally.

PR #2418, opened 2024-11-26
Reduce memory use during tests and relax memory limit

### LINKED ISSUE #2417: Pytest fails due to "not enough physical memory available"
### Summary

Sometimes pytest will fail unexpectedly with an error about there not being enough memory to allocate arrays for images. While it is a genuine message about there being limited memory available, it occurs when there is still reasonable amount of free ram.

### Steps To Reproduce

Run pytest on machine where memory about 80% used.

### Expected Behaviour

Tests should not use much memory. And should therefore run when there are a few GB free.

### Current Behaviour

The tests use a significant amount of memory because they start a python process for every CPU core, and because `system_free_memory()` will not use the last 12.5% of RAM. 

### Context

Current main, has been happening for ages

samtygier-stfc:
### Issue

Closes #2417

### Description
Prevent a large number of python processes being started when running pytest

Previous Pytest would start a worker for each physical cpu core (though `pytest-xdist`), each of these could start a multiprocessing pool with a process for each logical core (hyperthread). On a 8-core 16-thread, this could launch many python processes and use 8GB of RAM.

This PR does several things
* Limit size of multiprocessing pool to 2 during tests (it is controlled by a setting for the application)
* Limit number of pytest workers to 4
* Keep all tests that use multiprocess in a single pytest worker (using [xdist loadgroup](https://pytest-xdist.readthedocs.io/en/stable/distribution.html))

### Testing 

### Acceptance Criteria 

### Documentation

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -110,7 +110,7 @@ jobs:
         timeout-minutes: 10
         shell: bash -l {0}
         run: |
-          xvfb-run --auto-servernum python -m pytest --cov --cov-report=xml -n auto --maxprocesses=4 -o log_cli=true --ignore=mantidimaging/eyes_tests --durations=10
+          xvfb-run --auto-servernum python -m pytest --cov --cov-report=xml -n auto --maxprocesses=4 --dist loadgroup -o log_cli=true --ignore=mantidimaging/eyes_tests --durations=10
 
       - name: Get test data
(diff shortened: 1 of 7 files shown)


ashleymeigh2 (commented):
Changes look good! Limiting multiprocessing and pytest workers is a sensible solution , and everything seems in order. 👍

PR #2425, opened 2024-12-03
Live Viewer shows spectrum of images via caching recent images

### LINKED ISSUE #2421: Use caching of images in the Live Viewer for most recent images
### Desired Behaviour

Following from the work done with Dask

MikeSullivan7:
### Issue

Closes #2421

### Description

An ImageCache has been implemented in the Live Viewer which holds a dictionary of a given size, which image data is stored in. When loaded through the ImageCache, if the image is already in cache, the data is returned without further computation. This provides a performance improvement with compromise of using RAM. 

The Live Viewer now has the option to show the spectrum of live data via a right click menu. If the ROI is not moved, the mean spectrum of the newest image is added to the plotted spectrum. If the ROI is moved, all previous images are loaded in a loop, where the mean is stored and plotted. This is done in a separate thread to prevent loss of interaction with the GUI. The calculation of the mean also uses the ImageCache to speed up calculation for those images stored in the cache.

### Testing 

make check

### Acceptance Criteria 

Open MI and the Live Viewer Window
Right click and click Operations -> Calculate Spectrum
Move images into LV directory via a script and check that the Spectrum is calculated as expected.
While images are incoming, move the ROI and check that the Spectrum recalculates and plots with no interruption of the GUI.

### Documentation

Release Note


mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -114,8 +177,20 @@ def path(self, path: Path) -> None:
...
+    @property
+    def images(self):
+        return self._images if self._images is not None else None
+
+    @images.setter
+    def images(self, images):
+        self._images = images
+
+    def set_roi(self, roi: SensibleROI):
+        self.roi = roi
```
samtygier-stfc, line 192: I don't think these are needed. Could just let things directly access `images` and `roi` as attributes.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -137,6 +214,46 @@ def close(self) -> None:
             self.image_watcher = None
         self.presenter = None  # type: ignore # Model instance to be destroyed -type can be inconsistent
 
+    def add_mean(self, image_data_obj: Image_Data, image_array: np.ndarray) -> None:
+        if self.roi and (self.roi.left, self.roi.top, self.roi.right, self.roi.bottom) != (0, 0, 0, 0):
```
samtygier-stfc, line 218: Can this be `if self.roi is not None:`. Shouldn't self.roi either be a valid value or None?

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -137,6 +214,46 @@ def close(self) -> None:
...
+        if self.images is not None:
+            nanInds = np.argwhere(np.isnan(self.mean))
+            if self.roi:
+                left, top, right, bottom = self.roi
+            else:
+                left, top, right, bottom = (0, 0, -1, -1)
+            if nanInds.size > 0:
+                for ind in range(len(nanInds) - 1, len(nanInds) - 1 - chunk_size, -1):
+                    if ind < 0:
+                        break
```
samtygier-stfc, line 249: Would `for ind in nanInds[-1:-chunk_size:-1]:` work?
Or `for ind in nanInds.flip()[:chunk_size]:`

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -161,7 +278,9 @@ class ImageWatcher(QObject):
         Sort the images by modified time.
     """
     image_changed = pyqtSignal(list)  # Signal emitted when an image is added or removed
+    update_spectrum = pyqtSignal(np.ndarray)  # Signal emitted to update the Live Viewer Spectrum
     recent_image_changed = pyqtSignal(Path)
+    create_delayed_array: bool = False
```
samtygier-stfc, line 283: Not used

mantidimaging/gui/windows/live_viewer/model.py
MikeSullivan7: The problem is that we want to calculate and add the mean to the spectrum in `Presenter.update_image_list()` but the first mean is calculated before the image is displayed in the image view so the roi is returned as (0,0,0,0), this is to cover that situation

mantidimaging/gui/windows/live_viewer/model.py
MikeSullivan7: `for ind in nanInds[-1:-chunk_size:-1]` works!

mantidimaging/gui/windows/live_viewer/model.py
samtygier-stfc: Is that because `LiveViewWidget.get_roi()` can return `SensibleROI()`? Would it work for it to return `None` in this case? `LiveViewerWindowModel.roi` is already `SensibleROI | None`, so everywhere need to be able to handle `None`

MikeSullivan7: Yes I suppose so, I could change it so it it returns `None` rather than a default `SensibleROI`

mantidimaging/gui/windows/live_viewer/live_view_widget.py
```diff
@@ -48,3 +58,42 @@ def handle_deleted(self) -> None:
...
+        roi = SensibleROI.from_list([0, 0, width, height])
+        self.roi_object = SpectrumROI('roi', roi, rotatable=False, scaleSnap=True, translateSnap=True)
+        self.roi_object.colour = (255, 194, 10, 255)
+        self.roi_object.hoverPen = mkPen(self.roi_object.colour, width=3)
+        self.roi_object.roi.sigRegionChangeFinished.connect(self.roi_changed.emit)
+        self.roi_object.roi.sigRegionChangeStarted.connect(self.roi_changed_start.emit)
+        self.image.vb.addItem(self.roi_object.roi)
+
+    def set_image_shape(self, shape: tuple) -> None:
+        self.image_shape = shape
```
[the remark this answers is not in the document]
samtygier-stfc: I think `set_image_shape()` can be removed now.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -137,6 +207,50 @@ def close(self) -> None:
...
+    def calc_mean_chunk(self, chunk_size: int) -> None:
+        if self.images is not None:
+            nanInds = np.argwhere(np.isnan(self.mean))
+            if self.roi:
+                left, top, right, bottom = self.roi
+            else:
+                left, top, right, bottom = (0, 0, -1, -1)
+            if nanInds.size > 0:
+                for ind in nanInds[-1:-chunk_size:-1]:
+                    if ind[0] < 0:
```
samtygier-stfc, line 243: Do you still need `if nanInds.size > 0` and `if ind[0] < 0`?

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -6,15 +6,79 @@
...
+                if self.max_cache_size <= len(self.cache_dict):
+                    self.remove_oldest_image()
+            self.cache_dict[image] = (image_array, image.image_modified_time)
+
+    def remove_from_cache(self, image: Image_Data):
+        if image in self.cache_dict:
+            del self.cache_dict[image]
+
+    def get_oldest_image(self):
+        time_ordered_cache = sorted(self.cache_dict.items(), key=lambda item: item[1][-1])
```
samtygier-stfc, line 62: You can used the modification time from the dictionary key to simplify this
```
time_ordered_cache = sorted(self.cache_dict.keys(), key=lambda item: item.image_modified_time)
return time_ordered_cache[0]
```

Then the dictionary does not need the mod time to be put in the value and can be `cache_dict: dict[Image_Data, np.ndarray]`.

You can also use `key=attrgetter('image_modified_time')`, see https://docs.python.org/3/howto/sorting.html#operator-module-functions-and-partial-function-evaluation

Also you can use `min` instead of `sorted`, which is faster

`return min(self.cache_dict.keys(), key=attrgetter('image_modified_time'))`

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -6,15 +6,79 @@
...
+    if image_path.suffix.lower() in [".tif", ".tiff"]:
+        with tifffile.TiffFile(image_path) as tif:
+            image_data = tif.asarray()
+    elif image_path.suffix.lower() == ".fits":
+        with fits.open(image_path.__str__()) as fit:
+            image_data = fit[0].data
+    return image_data
+
+
+class ImageCache:
```
samtygier-stfc, line 39: All the methods in the class should have type annotations for return values.

And I don't think `load_image()` ever returns `None`.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -6,15 +6,79 @@
...
+    def add_to_cache(self, image: Image_Data, image_array: np.ndarray):
+        if image not in self.cache_dict.keys():
+            if self.max_cache_size is not None:
+                if self.max_cache_size <= len(self.cache_dict):
+                    self.remove_oldest_image()
+            self.cache_dict[image] = (image_array, image.image_modified_time)
+
+    def remove_from_cache(self, image: Image_Data):
+        if image in self.cache_dict:
+            del self.cache_dict[image]
```
samtygier-stfc, line 59: This is only used in one test, so it and the test could be removed.

mantidimaging/gui/windows/live_viewer/model.py
samtygier-stfc: Also, as `load_image()` is the only method that we expect to be used from outside the class, we could mark the others as 'private' with a leading underscore, e.g. `_add_to_cache()`

samtygier-stfc (changes requested):

mantidimaging/gui/windows/live_viewer/model.py
MikeSullivan7: `load_image` can return `None` if an exception occurs, it doesn't reach a return statement

samtygier-stfc: If it raises an exception, then it does not count as a return, so `None` is not needed in the annotation.

PR #2434, opened 2024-12-11
Use instance=True in create_autospec

### LINKED ISSUE #2433: Work around create_autospec slowness
### Summary

`mock.create_autospec()` is slow because it must inspect the class in order to create a mock. This is worth doing, but it would be nice to not take such a performance hit from it.

### Steps To Reproduce
For example, by profiling a test,
```
python -m cProfile -o profile -m pytest mantidimaging/gui/windows/main/test/presenter_test.py
python -c "import pstats; pstats.Stats('profile').strip_dirs().sort_stats('cumtime').print_stats(50)"

```
we can see
```
         37228915 function calls (35971513 primitive calls) in 38.344 seconds

   Ordered by: cumulative time
   List reduced from 9195 to 50 due to restriction <50>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   2241/1    0.095    0.000   38.353   38.353 {built-in method builtins.exec}
        1    0.000    0.000   38.353   38.353 <string>:1(<module>)
        1    0.000    0.000   38.353   38.353 <frozen runpy>:201(run_module)
        1    0.000    0.000   38.117   38.117 <frozen runpy>:65(_run_code)
        2    0.000    0.000   38.117   19.058 __main__.py:1(<module>)
        1    0.000    0.000   38.117   38.117 __init__.py:185(console_main)
        1    0.000    0.000   38.117   38.117 __init__.py:135(main)
...
       83    0.004    0.000   33.244    0.401 presenter_test.py:25(setUp)
  513/264    0.693    0.001   31.746    0.120 mock.py:2697(create_autospec)

```
So of the 38 second run time, 32 seconds were inside calls to `create_autospec`.

### Context

Current main

samtygier-stfc:
Save time, by preventing create_autospec creating both a class and instance mock.

### Issue

Closes #2433

### Description

Without setting `instance`, calls `create_autospec()` create a mock that behaves as an instance or a class. This doubles the already long time it takes to inspect the class to make the auto_spec.

Using the profiler on
`python -m cProfile -o profile-after -m pytest mantidimaging/gui/windows/main/test/presenter_test.py`
the line for `create_autospec` is reduced from
`513/264    0.693    0.001   31.746    0.120 mock.py:2697(create_autospec)`
to
`264    0.434    0.002   16.114    0.061 mock.py:2697(create_autospec)`
Showing that the time is halved (31.746 -> 16.114), and the internal second call is prevented (513 calls, reduced to 264).

Overall the runtime of the full unit test suite, was reduced from around 174s to 143s.
For running `python -m pytest -p no:xdist -p no:randomly`, and using the pytest reported time. There is more run to run variation without disabling randomly, but still a notable improvement.

Before: `173.64s 172.07s 174.11s 175.10s 173.90s`
After: `144.57s 141.80s 141.47s 144.13s 144.25s`

All tests were done with CPU set to 2GHz to avoid variations
`sudo cpupower frequency-set -g performance ; sudo cpupower frequency-set --max 2G`

Check that auto mocks, are still catching problems, by adding call to a non-existent method in `mantidimaging/gui/windows/main/test/presenter_test.py`
e.g.
`self.view.does_not_exist()`

### Documentation

Release notes

### Notes

Change was performed with Comby
`comby "create_autospec(:[arg1])"  "create_autospec(:[arg1], instance=True)" .py -stats -i`

### DIFF
--- a/docs/release_notes/next/dev-2433-autospec-speedup
+++ b/docs/release_notes/next/dev-2433-autospec-speedup
@@ -1,0 +1,1 @@
+#2433: Speedup autospec in tests
--- a/mantidimaging/core/data/test/image_stack_test.py
+++ b/mantidimaging/core/data/test/image_stack_test.py
@@ -73,7 +73,7 @@ class ImageStackTest(unittest.TestCase):
     def test_loading_metadata_preserves_existing_log(self):
         json_file = io.StringIO('{"pixel_size": 30.0, "log_file": "/old/logfile"}')
         mock_log_path = Path("/aaa/bbb")
-        mock_log_file = mock.create_autospec(InstrumentLog, source_file=mock_log_path)
+        mock_log_file = mock.create_autospec(InstrumentLog, source_file=mock_log_path, instance=True)
 
         imgs = ImageStack(np.asarray([1]))
         self.assertEqual({}, imgs.metadata)
--- a/mantidimaging/core/io/loader/test/loader_test.py
+++ b/mantidimaging/core/io/loader/test/loader_test.py
@@ -77,11 +77,11 @@ class LoaderTest(FakeFSTestCase):
         filenames = [Path(f"foo_{n}.tif") for n in range(20)]
         angles = np.array([(n * 137.507764) % 360 for n in range(20)])
 
-        mock_filename_group = mock.create_autospec(FilenameGroup, metadata_path=None)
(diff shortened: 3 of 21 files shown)


ashleymeigh2 (approved):
This change is well-done and improves performance significantly without breaking functionality.

PR #2423, opened 2024-12-02
Refactor ROI Handling: Remove get_list_of_roi_names() and get_roi from Model

ashleymeigh2:
### Issue

Closes #2297

### Description

Removed get_list_of_roi_names() and get_roi methods from the model.
Refactored get_spectrum to only accept SensibleROI objects.
Updated all dependent methods (do_add_roi, do_remove_roi, redraw_spectrum, etc.) to retrieve and pass SensibleROI instead of strings.

### Testing 

Updated unit tests for methods like do_add_roi, do_remove_roi, redraw_spectrum, and redraw_all_rois to confirm correct behavior with the updated get_spectrum.
Manually tested adding, removing, and updating ROIs in the UI to ensure proper functionality.

### Acceptance Criteria 

All unit tests pass for updated methods.
UI operations related to ROI management work as expected without errors.

PR #2435, opened 2024-12-13
Refactor ROI Handling: removed rename_roi and remove_roi from Model

ashleymeigh2:
### Issue

Closes #2423 

### Description

Refactored rename_roi and remove_roi to handle ROI renaming directly in the presenter.
Improved error handling with raise ... from err for better traceability.

### Testing 

Updated unit tests to verify correct renaming behavior and proper exceptions for invalid operations.
Manually tested renaming ROIs in the UI to ensure changes are reflected correctly.

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -375,14 +375,7 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         @param old_name: Name of the ROI to rename
         @param new_name: New name of the ROI
         """
-        try:
-            self.view.spectrum_widget.rename_roi(old_name, new_name)
-        except RuntimeError as err:
-            raise RuntimeError(f"Cannot remove ROI: {old_name}") from err
-        except KeyError as err:
-            raise KeyError(
-                f"Cannot rename {old_name} to {new_name}. Available: {list(self.view.spectrum_widget.rois.keys())}"
-            ) from err
+        self.view.spectrum_widget.rename_roi(old_name, new_name)
 
     def do_remove_roi(self, roi_name: str | None = None) -> None:
         """
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -375,8 +375,14 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         @param old_name: Name of the ROI to rename
         @param new_name: New name of the ROI
         """
-        self.view.spectrum_widget.rename_roi(old_name, new_name)
-        self.model.rename_roi(old_name, new_name)
+        try:
+            self.view.spectrum_widget.rename_roi(old_name, new_name)
+        except RuntimeError as err:
+            raise RuntimeError(f"Cannot remove ROI: {old_name}") from err
+        except KeyError as err:
+            raise KeyError(
+                f"Cannot rename {old_name} to {new_name}. Available: {list(self.view.spectrum_widget.rois.keys())}"
+            ) from err
 
     def do_remove_roi(self, roi_name: str | None = None) -> None:
         """
@@ -391,7 +397,6 @@ class SpectrumViewerWindowPresenter(BasePresenter):
             self.model.remove_all_roi()
         else:
             self.view.spectrum_widget.remove_roi(roi_name)
-            self.model.remove_roi(roi_name)
 
     def handle_export_tab_change(self, index: int) -> None:
         self.export_mode = ExportMode(index)
--- a/mantidimaging/gui/windows/spectrum_viewer/test/model_test.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/test/model_test.py
@@ -392,26 +392,6 @@ class SpectrumViewerWindowModelTest(unittest.TestCase):
         self.model.set_stack(image_stack)
         self.assertEqual(self.model.has_stack(), expected)
 
-    def test_WHEN_roi_removed_THEN_roi_name_removed_from_list_of_roi_names(self):
-        self.model.set_stack(generate_images())
-        rois = ["roi", "new_roi"]
-        for roi in rois:
-            self.model.set_new_roi(roi)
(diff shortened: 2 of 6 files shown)


mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -375,8 +375,14 @@ def rename_roi(self, old_name: str, new_name: str) -> None:
...
-        self.view.spectrum_widget.rename_roi(old_name, new_name)
-        self.model.rename_roi(old_name, new_name)
+        try:
+            self.view.spectrum_widget.rename_roi(old_name, new_name)
+        except RuntimeError as err:
+            raise RuntimeError(f"Cannot remove ROI: {old_name}") from err
+        except KeyError as err:
+            raise KeyError(
+                f"Cannot rename {old_name} to {new_name}. Available: {list(self.view.spectrum_widget.rois.keys())}"
+            ) from err
```
samtygier-stfc, line 385: These exceptions don't get caught because, `spectrum_widget.rename_roi()` does not raise them. Probably best to do the check and raise the exceptions in `spectrum_widget.rename_roi()`. Then you don't need a try/except here.

samtygier-stfc: So in `spectrum_widget.rename_roi()` you have something like:
```
if old_name not in self.roi_dict.keys():
    raise RuntimeError(f"Try to remove non-existent ROI: {old_name}")
if new_name in self.roi_dict.keys():
    raise RuntimeError(f"New ROI name already exists: {new_name}")
```

mantidimaging/gui/windows/spectrum_viewer/test/model_test.py
```diff
@@ -392,26 +392,6 @@ def test_WHEN_stack_value_set_THEN_can_export_returns_(self, _, image_stack, exp
...
-        self.model.set_stack(generate_images())
-        with self.assertRaises(RuntimeError):
-            self.model.remove_roi("all")
-        self.assertListEqual(list(self.model._roi_ranges.keys()), ["all"])
-
-    def test_WHEN_invalid_roi_removed_THEN_keyerror_raised(self):
-        self.model.set_stack(generate_images())
-        with self.assertRaises(KeyError):
-            self.model.remove_roi("non_existent_roi")
-
```
samtygier-stfc, line 414: Tests like this could be added to spectrum widget tests.

mantidimaging/gui/windows/spectrum_viewer/test/presenter_test.py
```diff
@@ -259,13 +259,17 @@ def test_WHEN_do_add_roi_to_table_called_THEN_roi_added_to_table(self):
...
         self.presenter.do_remove_roi("roi_1")
-        self.assertEqual(["all", "roi"], list(self.presenter.model._roi_ranges.keys()))
+
+        self.presenter.view.spectrum_widget.remove_roi.assert_called_once_with("roi_1")
+        expected_roi_dict = {"all": mock.Mock(), "roi": mock.Mock()}
+        actual_roi_dict = {
+            key: value
+            for key, value in self.presenter.view.spectrum_widget.roi_dict.items() if key != "roi_1"
+        }
+        self.assertEqual(expected_roi_dict.keys(), actual_roi_dict.keys())
```
samtygier-stfc, line 272: This part is only testing if python dictionaries work, so could be removed.

All the test can usefully do is check that when `self.presenter.do_remove_roi("roi_1")` is called,  that `self.presenter.view.spectrum_widget.remove_roi()` gets called

mantidimaging/gui/windows/spectrum_viewer/test/presenter_test.py
```diff
@@ -285,18 +289,34 @@ def test_WHEN_rits_roi_clicked_THEN_rois_not_updated(self):
...
+        self.view.spectrum_widget.rename_roi = mock.Mock(
+            side_effect=lambda old_name, new_name: self.view.spectrum_widget.rois.update(
+                {new_name: self.view.spectrum_widget.rois.pop(old_name)}))
+
         self.presenter.rename_roi("roi_1", "new_name")
         self.view.spectrum_widget.rename_roi.assert_called_once_with("roi_1", "new_name")
-        self.assertIn("new_name", self.presenter.model._roi_ranges)
-        self.assertNotIn("roi_1", self.presenter.model._roi_ranges)
+        self.assertIn("new_name", self.view.spectrum_widget.rois)
+        self.assertNotIn("roi_1", self.view.spectrum_widget.rois)
```
samtygier-stfc, line 300: Using the lambda means this is only really testing the test. As this is a test for the presenter, then only bit to test is that it calls the function in spectrum widget.

mantidimaging/gui/windows/spectrum_viewer/test/presenter_test.py
```diff
@@ -285,18 +289,34 @@ def test_WHEN_rits_roi_clicked_THEN_rois_not_updated(self):
...
+    def test_WHEN_ROI_renamed_to_existing_name_THEN_keyerror(self, name):
+        rois = ["all", "roi", "roi_1"]
+        self.view.spectrum_widget.roi_dict = {roi: mock.Mock() for roi in rois}
+        self.view.spectrum_widget.rois = {roi: mock.Mock() for roi in rois}
+
+        self.view.spectrum_widget.rename_roi = mock.Mock(side_effect=lambda old_name, new_name: (
+            (_ for _ in ()).throw(KeyError(f"Cannot rename to existing ROI: {new_name}"))
+            if new_name in self.view.spectrum_widget.rois else None))
         with self.assertRaises(KeyError):
             self.presenter.rename_roi("roi", name)
```
samtygier-stfc, line 321: These are also testing the test. Right now `spectrum_widget.rename_roi` never raises an exception, so this is testing a response that doesn't happen.

Better would be to implement the checks in `spectrum_widget.rename_roi` and test them in the test for that. e.g. based on the tests removed from the model.

samtygier-stfc (changes requested):

samtygier-stfc (approved):
Thanks for the fixes. Looks good. I can add remove and rename ROIs.

PR #2440, opened 2025-01-03
Optimise convergence plotting

### LINKED ISSUE #2412: Convergence plot optimisation
Some optimsations for the the convergence plot in the reconstruction progress window

### Desired Behaviour

- Only create the `PlotWidget` the first time `set_progress_plot()` is called
- Only call `update_progress_plot()` when there is something new to plot (the loss is only calculated every 10th iteration). (maybe if the last element in iterations is the same as the current step, or something like that).

### Current Behaviour

PlotWidget is always created, even when it is not used

Plot gets updates on every iteration, even if nothing has changed

https://github.com/mantidproject/mantidimaging/pull/2399#pullrequestreview-2444990508

samtygier-stfc:
### Issue

Closes #2412

### Description

Reduce some unnecessary work, to make the convergence/residual plotting a bit smoother
* Only add the plot widget when it is needed
* Only update the plot for new data
* Don't store all the plot data and residuals in the progress history
* Don't plot the inital convergence value (it is from before the first iteration)
* Remove unneeded string formatting logic

Run a PDHG recon and check that plots and residual update


JackEAllen (approved):
Happy with the changes made here, the optimisations are sensible and nice to see the difference in performance. Thanks for the optimisation 👍

PR #2445, opened 2025-01-06
Add the ruff INP001 rule

### LINKED ISSUE #2444: Enable ruff implicit-namespace-package rule for source directory
### Desired Behaviour

The INP001 rule https://docs.astral.sh/ruff/rules/implicit-namespace-package/ would catch missing `__init__.py` files. This would avoid issues like #2442 

It needs to only apply to the `mantidimaging` directory to avoid false positives on things like the documentation `conf.py` and the stand alone scripts.

This will require an update to ruff, to allow negated `per-file-ignores` rules.

samtygier-stfc:
### Issue

Closes #2444  

### Description

Enable ruff INP001 rule to catch missing `__init__.py` files.
Use `per-file-ignores` to prevent false positives on stand alone scripts
Update ruff to allow negative `per-file-ignores` rules
Add missing `__init__.py` in `palette_changer/test`

Try removing a `__init__.py` file and run `make ruff`

### Documentation

Release notes

### DIFF
--- a/docs/release_notes/next/dev-2444-ruff-implicit-namespace-package
+++ b/docs/release_notes/next/dev-2444-ruff-implicit-namespace-package
@@ -1,0 +1,1 @@
+#2444: Ruff rule to check for missing __init__.py files
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -29,7 +29,7 @@ dependencies:
   - pyinstaller==6.9.*
   - pyright==1.1.*
   - make==4.3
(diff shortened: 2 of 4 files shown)


JackEAllen (approved):
Thank you turning on the [per-file-ignores](https://docs.astral.sh/ruff/settings/#lint_per-file-ignores) rule in Ruff. I can confirm that running `make ruff` will now find missing `__init__.py` files. Happy to approve! 👍

PR #2451, opened 2025-01-08
Refactored set_new_roi into the presenter.py

### LINKED ISSUE #2297: Remove _roi_ranges dict from SpectrumViewerWindowModel
Originally `SpectrumViewerWindowModel` had the dictionary mapping names to ROIs for the spectrum window. This allows things like `get_spectrum()` to be passed a name. `SpectrumWidget` also has a dict of ROIs, which is probably more essential.

This requires code to keep them in sync, and has been a source of issues.

 `get_spectrum()` needed to be updated to be able to receive a `SensibleROI` for the RITS export, so its less important that the model knows about the names. Maybe it is possible to remove `_roi_ranges` completely.

Step 1:
In `SpectrumViewerWindowModel.get_spectrum()` remove the option to call with a str as the name. Update all callers.

Find all other cases where model uses its own list of ROIs. Pass them as arguments instead.

Step N:
Remove `set_new_roi`, `set_roi`, `get_roi`, `rename_roi`, `remove_roi`, `get_list_of_roi_names` from the model.

ashleymeigh2:
### Issue

Closes #2297  

### Description
Set_roi removed from model 
refactored presenter to no longer use set_new_roi from model 

### Testing 

Updated unit tests to verify correct renaming behavior and proper exceptions for invalid operations.
Manually tested renaming ROIs in the UI to ensure changes are reflected correctly.

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -328,10 +328,9 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         roi_name = self.model.roi_name_generator()
         if roi_name in self.view.spectrum_widget.roi_dict:
             raise ValueError(f"ROI name already exists: {roi_name}")
-        self.model.set_new_roi(roi_name)
-        roi = self.model._roi_ranges.get(roi_name)
-        if roi is None:
-            raise ValueError(f"ROI for {roi_name} is not valid.")
+        height, width = self.model.get_image_shape()
+        roi = SensibleROI.from_list([0, 0, width, height])
+        self.model._roi_ranges[roi_name] = roi
         self.view.spectrum_widget.add_roi(roi, roi_name)
         spectrum = self.model.get_spectrum(roi, self.spectrum_mode, self.view.shuttercount_norm_enabled())
         self.view.set_spectrum(roi_name, spectrum)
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -328,10 +328,9 @@ def do_add_roi(self) -> None:
         roi_name = self.model.roi_name_generator()
         if roi_name in self.view.spectrum_widget.roi_dict:
             raise ValueError(f"ROI name already exists: {roi_name}")
-        self.model.set_new_roi(roi_name)
-        roi = self.model._roi_ranges.get(roi_name)
-        if roi is None:
-            raise ValueError(f"ROI for {roi_name} is not valid.")
+        height, width = self.model.get_image_shape()
+        roi = SensibleROI.from_list([0, 0, width, height])
+        self.model._roi_ranges[roi_name] = roi
```
samtygier-stfc, line 333: Does `self.model._roi_ranges` still need to be set?

mantidimaging/gui/windows/spectrum_viewer/test/model_test.py
```diff
@@ -380,7 +380,7 @@ def test_WHEN_rois_deleted_THEN_name_generator_is_reset(self):
 
     def test_when_new_roi_set_THEN_roi_name_added_to_list_of_roi_names(self):
         self.model.set_stack(generate_images())
-        self.model.set_new_roi("new_roi")
+        self.model._roi_ranges["new_roi"] = SensibleROI.from_list([0, 0, 10, 10])
         self.assertIn("new_roi", self.model._roi_ranges)
         self.assertListEqual(list(self.model._roi_ranges.keys()), ["all", "new_roi"])
```
samtygier-stfc, line 385: This test is no longer needed.

samtygier-stfc (changes requested):
There is also a yapf issue that needs resolving

samtygier-stfc (approved):
Looks good.

PR #2457, opened 2025-01-14
Live Viewer calculates intensity of live images via QThreads

### LINKED ISSUE #2448: Calculate Live Viewer Spectrum asynchronously
### Desired Behaviour

When the Spectrum is calculated in the Live Viewer, we want to do this asynchronously to the GUI such that the GUI remains responsive to the user.

### Current Behaviour

Currently, MI calculates the Spectrum in the same thread as the GUI which can lead to unresponsiveness for large datasets.

MikeSullivan7:
### Issue

Closes #2448

### Description

The calculation of the Intensity Profile in the Live Viewer is now done asynchronously via a QThread and the GUI is updated via the main thread upon the QThread termination. The intensity is calculated in chunks of 100 points, each done via a different thread. It is done in this way so that the GUI will remain responsive and smooth while the ROI is moved on the image, all while the intensity updates to the new position. The calculation speed is also increased by using the `ImageCache` class.  
Unit and System tests have been added for the Live Viewer.

~~This PR will require PR #2425 after which this PR will need to be rebased.~~

### Testing 

`make check`
`make test-system`

### Acceptance Criteria 

Check that all tests pass.

1. Open Live Viewer
2. Right click -> Operations -> Calculate Intensity
3. Ensure that the Intensity plot opens, the ROI is visible, and that the intensity is plotted correctly (check against known data)
4. Move and resize the ROI and check that the Intensity plot updates smoothly with no GUI interruption.

### Documentation

Release Note


mantidimaging/gui/windows/live_viewer/live_view_widget.py
```diff
@@ -18,6 +23,10 @@ class LiveViewWidget(GraphicsLayoutWidget):
     @param parent: The parent widget
     """
     image: MIMiniImageView
+    image_shape: tuple = (-1, -1)
```
samtygier-stfc, line 26: Safer for the default value to be `None`, so the type checker can make sure we check when needed and to avoid weird behaviour if we ever used (-1,-1).

`image_shape: tuple[int, int] | None = None`

mantidimaging/gui/windows/live_viewer/live_view_widget.py
```diff
@@ -46,5 +56,36 @@ def handle_deleted(self) -> None:
...
+        roi = SensibleROI.from_list([0, 0, width, height])
+        self.roi_object = SpectrumROI('roi', roi, rotatable=False, scaleSnap=True, translateSnap=True)
+        self.roi_object.colour = (255, 194, 10, 255)
+        self.roi_object.hoverPen = mkPen(self.roi_object.colour, width=3)
+        self.roi_object.roi.sigRegionChangeFinished.connect(self.roi_changed.emit)
+        self.roi_object.roi.sigRegionChanged.connect(self.roi_changing.emit)
+        self.image.vb.addItem(self.roi_object.roi)
+
+    def set_image_shape(self, shape: tuple) -> None:
+        self.image_shape = shape
```
samtygier-stfc, line 75: I don't think this is used

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -3,18 +3,79 @@
...
+def load_image_from_path(image_path: Path) -> np.ndarray:
+    """
+    Load a .Tif, .Tiff or .Fits file only if it exists
+    and returns as an ndarray
+    """
+    if image_path.suffix.lower() in [".tif", ".tiff"]:
+        with tifffile.TiffFile(image_path) as tif:
+            image_data = tif.asarray()
+    elif image_path.suffix.lower() == ".fits":
+        with fits.open(image_path.__str__()) as fit:
```
samtygier-stfc, line 35: There might be a conflict here with #2462

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -125,9 +188,11 @@ def _handle_image_changed_in_list(self, image_files: list[Image_Data]) -> None:
         :param image_files: list of image files
         """
         self.images = image_files
+        # if dask_image_stack.image_list:
+        #     self.image_stack = dask_image_stack
```
samtygier-stfc, line 192: Comments can be removed

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -137,6 +202,48 @@ def close(self) -> None:
...
+        self.mean_paths.add(image_data_obj.image_path)
+        self.mean = np.append(self.mean, mean_to_add)
+
+    def clear_mean_partial(self) -> None:
+        self.mean_paths.clear()
+        self.mean = np.full(len(self.images), np.nan)
+
+    def clear_mean(self) -> None:
+        self.mean_paths.clear()
+        self.mean = np.delete(self.mean, np.arange(self.mean.size))
```
samtygier-stfc, line 222: I think this is not used

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -137,6 +202,48 @@ def close(self) -> None:
...
+
+    def calc_mean_fully(self) -> None:
+        if self.images is not None:
+            for image in self.images:
+                self.add_mean(image, self.image_cache.load_image(image))
```
samtygier-stfc, line 227: I think this is not used.

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -137,6 +202,48 @@ def close(self) -> None:
...
+            if nanInds.size > 0:
+                for ind in nanInds[-1:-chunk_size:-1]:
+                    buffer_data = self.image_cache.load_image(self.images[ind[0]])
+                    if buffer_data is not None:
+                        buffer_mean = np.mean(buffer_data[top:bottom, left:right])
+                        np.put(self.mean, ind, buffer_mean)
+
+    def calc_mean_all_chunks(self, chunk_size: int) -> None:
+        while np.isnan(self.mean).any():
+            self.calc_mean_chunk(chunk_size)
```
samtygier-stfc, line 245: I think this is not used.

mantidimaging/gui/windows/live_viewer/presenter.py
```diff
@@ -160,3 +176,49 @@ def load_as_dataset(self) -> None:
...
+    def set_roi_enabled(self, enable: bool) -> None:
+        if self.view.live_viewer.roi_object is not None:
+            self.view.live_viewer.roi_object.roi.translatable = enable
+            self.view.live_viewer.roi_object.roi.resizable = enable
+            self.view.live_viewer.roi_object.roi.blockSignals(not enable)
+
+    def try_next_mean_chunk(self) -> None:
+        if np.isnan(self.model.mean).any():
+            if not self.handle_roi_change_timer.isActive():
+                self.handle_roi_change_timer.start(100)
```
samtygier-stfc, line 224: I think this could also be `10`

mantidimaging/gui/windows/live_viewer/test/model_test.py
```diff
@@ -159,3 +163,56 @@ def test_WHEN_sub_directory_change_THEN_images_emitted(self, _mock_time):
...
+
+class ImageCacheTest(unittest.TestCase):
+
+    def setUp(self) -> None:
+        super().setUp()
+        self.image_data_list = []
+        self.image_array_mock_list = [np.random.default_rng().random(5)] * 5
+        for i in range(5):
+            self.image_data_list.append(mock.create_autospec(Image_Data))
+            self.image_data_list[i] = mock.create_autospec(Image_Data)
```
samtygier-stfc, line 176: I think you are creating this twice here.

samtygier-stfc (changes requested):
Looking good from a first pass, here are some requests for the code.

mantidimaging/gui/windows/live_viewer/live_view_widget.py
MikeSullivan7: If I remember correctly, my reason for putting `(-1, -1)` was due to when trying to get `roi.height` and `roi.width`, if the ROI has not been fully initialised, the class gives a default of `(-1, -1)` so I was trying to catch that situation. This might not be needed anymore though as the code has been rewritten a few times, I'll change it and check

mantidimaging/eyes_tests/live_viewer_window_test.py
```diff
@@ -56,7 +56,8 @@ def test_live_view_opens_without_data(self, _mock_time, _mock_image_watcher):
         self.imaging.show_live_viewer(self.live_directory)
         self.check_target(widget=self.imaging.live_viewer_list[-1])
 
-    @mock.patch('mantidimaging.gui.windows.live_viewer.presenter.LiveViewerWindowPresenter.load_image')
+    # @mock.patch('mantidimaging.gui.windows.live_viewer.model.ImageCache.loading_func')
```
samtygier-stfc, line 59: If the commented code is not needed it can be removed

mantidimaging/gui/windows/live_viewer/model.py
```diff
@@ -3,18 +3,67 @@
...
+            self.cache_dict[image] = image_array
+
+    def _get_oldest_image(self) -> Image_Data:
+        time_ordered_cache = min(self.cache_dict.keys(), key=attrgetter('image_modified_time'))
+        return time_ordered_cache
+
+    def _remove_oldest_image(self) -> None:
+        del self.cache_dict[self._get_oldest_image()]
+
+    def use_loading_function(self, func):
```
samtygier-stfc, line 50: This needs a type annotation

mantidimaging/gui/windows/live_viewer/live_view_widget.py
```diff
@@ -18,6 +23,10 @@ class LiveViewWidget(GraphicsLayoutWidget):
     @param parent: The parent widget
     """
     image: MIMiniImageView
+    image_shape: tuple[int, ...] | None = None
```
samtygier-stfc, line 26: `tuple[int, int]` is a bit more specific as it fixes the length to 2

samtygier-stfc (changes requested):
Looking good. I couple of code nitpicks.

I hit a couple of issues when running the simulate script in slow mode `--mode slow`, but I'd be happy to merge this and fix those after.

mantidimaging/gui/windows/live_viewer/live_view_widget.py
samtygier-stfc: Ok, not a big issue.

One thing to try would be to put `assert len(image.shape) == 2` in `show_image()`.

samtygier-stfc (approved):
Looks good. I have tested locally.

There are a few issues to be done as separate PRs, #2469 #2474

PR #2465, opened 2025-01-17
Added stack properties dialog in the Main Window

### LINKED ISSUE #2368: Stack Properties Menu
### Desired Behaviour

Add a right click menu option to stacks to allow users to view a properties of a stack through a dialog window.

Shape, datatype, memory size, original file paths, ...

### Current Behaviour

Users a not able to view the properties of files within a given stack in Mantid Imaging.

MikeSullivan7:
### Issue

Closes #2368

### Description

Added a custom dialog window which can be accessed by right clicking a stack in the main window and clicking on "Stack properties". This gives information on the stack e.g, Dataset name, File Path, Data type, Memory size, Shape, Log file. More information can be added if needed. 

### Testing 

Make check

### Acceptance Criteria 

Load a dataset into MI and open the Stack Properties dialog window via the right click window. Check that all the information displayed is correct and that no errors occur

### DIFF
--- a/mantidimaging/gui/windows/main/view.py
+++ b/mantidimaging/gui/windows/main/view.py
@@ -40,6 +40,7 @@ from mantidimaging.gui.windows.settings.view import SettingsWindowView
 from mantidimaging.gui.windows.spectrum_viewer.view import SpectrumViewerWindowView
 from mantidimaging.gui.windows.live_viewer.view import LiveViewerWindowView
 from mantidimaging.gui.windows.stack_choice.compare_presenter import StackComparePresenter
+from mantidimaging.gui.windows.stack_properties_dialog.view import StackPropertiesDialog
 from mantidimaging.gui.windows.stack_visualiser import StackVisualiserView
 from mantidimaging.gui.windows.welcome_screen.presenter import WelcomeScreenPresenter
 from mantidimaging.gui.windows.wizard.presenter import WizardPresenter
@@ -660,6 +661,8 @@ class MainWindowView(BaseMainWindowView):
                 delete_action = self.menuTreeView.addAction("Delete")
                 delete_action.triggered.connect(self._delete_container)
             if self.dataset_tree_widget.itemAt(position).id in self.presenter.all_stack_ids:
+                properties_action = self.menuTreeView.addAction("Stack Properties")
+                properties_action.triggered.connect(self._stack_properties)
                 move_action = self.menuTreeView.addAction("Move Stack")
                 move_action.triggered.connect(self._move_stack)
 
@@ -682,6 +685,10 @@ class MainWindowView(BaseMainWindowView):
     def _move_stack(self) -> None:
         stack_id = self.dataset_tree_widget.selectedItems()[0].id
         self.presenter.notify(PresNotification.SHOW_MOVE_STACK_DIALOG, stack_id=stack_id)
+
+    def _stack_properties(self) -> None:
+        stack_id = self.dataset_tree_widget.selectedItems()[0].id
+        self.presenter.notify(PresNotification.SHOW_PROPERTIES_DIALOG, stack_id=stack_id)
 
     def _bring_stack_tab_to_front(self, item: QTreeDatasetWidgetItem) -> None:
         """
@@ -757,3 +764,9 @@ class MainWindowView(BaseMainWindowView):
         self.move_stack_dialog = MoveStackDialog(self, origin_dataset_id, stack_id, origin_dataset_name,
                                                  stack_data_type)
         self.move_stack_dialog.show()
+
+    def show_stack_properties_dialog(self, stack_id: uuid.UUID, origin_dataset: Dataset, stack_data_type: str) -> None:
+        stack = self.presenter.model.get_images_by_uuid(stack_id)
+        assert stack is not None
+        stack_properties_dialog = StackPropertiesDialog(self, stack, origin_dataset, stack_data_type)
+        stack_properties_dialog.show()
--- a/mantidimaging/gui/windows/stack_properties_dialog/presenter.py
+++ b/mantidimaging/gui/windows/stack_properties_dialog/presenter.py
@@ -1,0 +1,36 @@
+# Copyright (C) 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from mantidimaging.core.utility.size_calculator import full_size_MB
+from mantidimaging.gui.mvp_base import BasePresenter
+
+if TYPE_CHECKING:
+    from mantidimaging.gui.windows.stack_properties_dialog.view import StackPropertiesDialog
+
+
+class StackPropertiesPresenter(BasePresenter):
+
+    def __init__(self, view: StackPropertiesDialog):
+        super().__init__(view)
+
+    def set_stack_data(self) -> None:
+        if self.view.stack is not None:
+            self.view.stack_shape = self.view.stack.data.shape
+        self.view.stack_size_MB = self.get_stack_size_MB()
+
+    def set_stack_directory(self) -> None:
+        if self.view.stack.filenames is not None:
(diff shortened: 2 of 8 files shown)


mantidimaging/gui/windows/stack_properties_dialog/presenter.py
```diff
@@ -0,0 +1,42 @@
...
+            if self.view.stack.filenames is not None:
+                self.view.directory = self.view.stack.filenames[0].replace(self.view.stack.filenames[0].split("\\")[-1],
+                                                                           '')
+            self.view.stack_shape = self.view.stack.data.shape
+        self.view.stack_size_MB = self.get_stack_size_MB()
+
+    def get_stack_size_MB(self):
+        return sum(
+            os.path.getsize(self.view.directory + f)
+            for f in os.listdir(self.view.directory) if os.path.isfile(self.view.directory + f)) / 1024 / 1024
```
samtygier-stfc, line 42: I get
```
  File "/home/sam/git/mantidimaging/mantidimaging/gui/windows/stack_properties_dialog/presenter.py", line 42, in get_stack_size_MB
    for f in os.listdir(self.view.directory) if os.path.isfile(self.view.directory + f)) / 1024 / 1024
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: ''
```

In general it would be better to do all the path handing with pathlib, e.g. `path.parent()` instead using string functions.

But here it would be better to use the memory size of the array rather than the file sizes. There are functions in `size_calculator.py` to do it.

samtygier-stfc (changes requested):

mantidimaging/gui/windows/main/view.py
```diff
@@ -757,3 +765,7 @@ def show_move_stack_dialog(self, origin_dataset_id: uuid.UUID, stack_id: uuid.UU
         self.move_stack_dialog = MoveStackDialog(self, origin_dataset_id, stack_id, origin_dataset_name,
                                                  stack_data_type)
         self.move_stack_dialog.show()
+
+    def show_stack_properties_dialog(self, stack_id: uuid.UUID, origin_dataset: Dataset, stack_data_type: str) -> None:
+        self.stack_properties_dialog = StackPropertiesDialog(self, stack_id, origin_dataset, stack_data_type)
+        self.stack_properties_dialog.show()
```
samtygier-stfc, line 771: Can probably just be
```
stack_properties_dialog = StackPropertiesDialog(self, stack_id, origin_dataset, stack_data_type)
stack_properties_dialog.show()
```
Because we done need to access it again from the main window. Also as it currently is, the value gets over written if multiple dialogs are opened.

mantidimaging/gui/windows/stack_properties_dialog/presenter.py
```diff
@@ -0,0 +1,40 @@
...
+        if origin_data_type == "Sample":
+            self.view.stack = self.view.origin_dataset.sample
+        elif origin_data_type == "Flat Before":
+            self.view.stack = self.view.origin_dataset.flat_before
+        elif origin_data_type == "Flat After":
+            self.view.stack = self.view.origin_dataset.flat_after
+        elif origin_data_type == "Dark Before":
+            self.view.stack = self.view.origin_dataset.dark_before
+        elif origin_data_type == "Dark After":
+            self.view.stack = self.view.origin_dataset.dark_after
```
samtygier-stfc, line 29: This seems quite round about. We already have the stack id, so shouldn't need find the stack from the dataset. Also this breaks for other stack types.

mantidimaging/gui/windows/stack_properties_dialog/presenter.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def set_stack_data(self):
+        if self.view.stack is not None:
+            if self.view.stack.filenames is not None:
+                self.view.directory = self.view.stack.filenames[0].replace(self.view.stack.filenames[0].split("\\")[-1],
```
samtygier-stfc, line 34: Safer to use `pathlib` for path manipulation. E.g. `self.view.stack.filenames[0].parent` or if its not already a path `Path(self.view.stack.filenames[0]).parent`

mantidimaging/gui/windows/stack_properties_dialog/view.py
```diff
@@ -0,0 +1,61 @@
...
+        self.presenter.notify(Notification.ACCEPTED)
+        self.close()
+
+    @property
+    def origin_dataset_id(self) -> uuid.UUID:
+        return self._origin_dataset_id
+
+    @property
+    def stack_id(self) -> uuid.UUID:
+        return self._stack_id
```
samtygier-stfc, line 61: I don't think these need to be properties

samtygier-stfc (changes requested):

mantidimaging/gui/windows/stack_properties_dialog/presenter.py
```diff
@@ -0,0 +1,36 @@
...
+    from mantidimaging.gui.windows.stack_properties_dialog.view import StackPropertiesDialog
+
+
+class StackPropertiesPresenter(BasePresenter):
+
+    def __init__(self, view: StackPropertiesDialog):
+        super().__init__(view)
+
+    def set_stack_data(self) -> None:
+        if self.view.stack is not None:
```
samtygier-stfc, line 21: Not sure if this is from a mypy suggestion or older version of the code.
If `self.view.stack` is `None` then something has gone wrong (e.g. side bar has got out of sync), so I don't think it should be silently ignored here.

It might be better handled with an `assert self.stack is not None` in the view, so that the rest of the dialog can assume that it is set. Or this and the call to `get_images_by_uuid()` could move into the main window.

mantidimaging/gui/windows/stack_properties_dialog/presenter.py
```diff
@@ -0,0 +1,36 @@
...
+            self.view.stack_shape = self.view.stack.data.shape
+        self.view.stack_size_MB = self.get_stack_size_MB()
+
+    def set_stack_directory(self):
```
samtygier-stfc, line 25: Needs a `-> None`

mantidimaging/gui/windows/main/view.py
```diff
@@ -683,6 +687,10 @@ def _move_stack(self) -> None:
         stack_id = self.dataset_tree_widget.selectedItems()[0].id
         self.presenter.notify(PresNotification.SHOW_MOVE_STACK_DIALOG, stack_id=stack_id)
 
+    def _stack_properties(self):
```
samtygier-stfc, line 690: Needs a `-> None`

mantidimaging/gui/windows/main/view.py
```diff
@@ -109,6 +110,7 @@ class MainWindowView(BaseMainWindowView):
     nexus_save_dialog: NexusSaveDialog | None = None
     add_to_dataset_dialog: AddImagesToDatasetDialog | None = None
     move_stack_dialog: MoveStackDialog | None = None
+    stack_properties_dialog: StackPropertiesDialog | None = None
```
samtygier-stfc, line 113: This is no longer needed, I think its left over from when there was a `self.stack_properties_dialog`

mantidimaging/gui/windows/stack_properties_dialog/view.py
```diff
@@ -0,0 +1,54 @@
+# Copyright (C) 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+from __future__ import annotations
+
+import uuid
+
+from PyQt5.QtWidgets import QLabel, QGridLayout
+
+from mantidimaging.core.data import ImageStack
+from mantidimaging.core.data.dataset import Dataset
```
samtygier-stfc, line 10: These could be in a `if TYPE_CHECKING:`

samtygier-stfc (changes requested):
I get an error if I try to view properties of a 180 projection. I think a case is missing in `_get_stack_data_type()`

samtygier-stfc (approved):
Looks good, tested locally. I did a squash, because there were quite a few commits.

PR #2491, opened 2025-02-03
Live Viewer Intensity Plot recalculates when images modified

### LINKED ISSUE #2490: Handle images arriving simultaneously and images being deleted in Live Viewer (Intensity Plot)
### Summary

When the Live Viewer image display can handle cases where images arrive together quickly or when images are deleted/renamed etc, this behaviour needs to be reflected when the intensity plot is updated or when the ROI is moved.

### Steps To Reproduce

Open the Live Viewer
With images showing and the Intensity plot active. delete images in the directory. The image should update in the LV but the mean is not recalculated correctly

### Current Behaviour

When the overall state of the images in the directory changes, the Intensity plot does not update and instead attempts to add further data (the latest image) which is not correct

### Expected Behaviour

When the overall state of the images in the directory changes, the Intensity plot should update appropriately by clearing and replotting.

MikeSullivan7:
### Description

The Live Viewer now keeps track of the Updated Image List and compares it to the new Image List when changes are made in the directory. This allows the LV to detect if more than one image has appeared at a time, or if any number of images have been deleted and recalculate the mean if this has happened. The `update_image_list()` method in the `presenter` is now put on a QTimer as when a large number of images are deleted or changed, it can take long enough that `update_image_list()` could be triggered multiple times in short succession which lead to peoples with the QThreads.

### Developer Testing 

`make check`
`make test-system`

- [ ] Open the LV with some images in the directory
- [ ] Add and delete images of various quantities in the directory and check that the Intensity Plot updates accordingly.

### DIFF
--- a/mantidimaging/gui/windows/live_viewer/presenter.py
+++ b/mantidimaging/gui/windows/live_viewer/presenter.py
@@ -52,6 +52,7 @@ class LiveViewerWindowPresenter(BasePresenter):
     op_func: Callable
     thread: QThread
     worker: Worker
+    old_image_list_paths: list[Path] = []
 
     def __init__(self, view: LiveViewerWindowView, main_window: MainWindowView):
         super().__init__(view)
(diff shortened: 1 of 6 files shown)


mantidimaging/gui/windows/live_viewer/presenter.py
```diff
@@ -90,28 +97,43 @@ def handle_deleted(self) -> None:
...
-                self.model.images = images_list
-                try:
-                    image_data = self.model.image_cache.load_image(images_list[-1])
-                    self.model.add_mean(images_list[-1], image_data)
-                except ImageLoadFailError as error:
-                    logger.error(error.message)
-                    self.model.add_mean(images_list[-1], None)
-                self.update_intensity(self.model.mean)
+                images_list_paths = [image.image_path for image in images_list]
+                if self.old_image_list_paths + images_list_paths[-1:] == images_list_paths:
```
samtygier-stfc, line 114: Can this be
`if self.old_image_list_paths == images_list_paths[:-1]:`

samtygier-stfc (changes requested):

samtygier-stfc (approved):
Looks good, and seem robust in my testing on linux

PR #2517, opened 2025-03-04
Avoid forcing a redraw on ROI row selection change

### LINKED ISSUE #2508: Spectrum plotting when ROI is disabled
### Summary

In the Spectrum Viewer, ROIs are plotted even once they have been disabled in the ROI Table.

### Steps To Reproduce

1. Load data in MI and open Spectrum Viewer.
2. Add ROIs so that you have multiple and move them so they are distinct.
3. Disable one of the ROIs and check that the plot is not visible in the plot widget.
4. Either click on a row in the ROI Table or move an active ROI
5. Verify that all ROIs (even disabled ones) are now plotted in the plot widget.

### Expected Behaviour

When disabled, the ROI's should not plot at all, and only plot when they are re-enabled

### Context

main branch

### Screenshot(s)

Image

samtygier-stfc:
### Description

Avoid forcing a redraw on ROI row selection change

This also reduces some unneeded replots during opening the window. This reduces the time to open the spectrum viewer with the 2kx2k flower dataset from
`INFO: SpectrumViewerWindowView shown in 15.486498130019754`
to

### Developer Testing 

- ...

- [ ] Follow steps on issue

Not needed because its fixing a recent regression

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -404,8 +404,6 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         self.handle_tof_unit_change()
 
     def refresh_spectrum_plot(self) -> None:
-        self.view.spectrum_widget.spectrum.clearPlots()
-        self.view.spectrum_widget.spectrum.update()
         self.view.show_visible_spectrums()
         self.view.spectrum_widget.spectrum_plot_widget.add_range(*self.model.tof_plot_range)
         self.view.spectrum_widget.spectrum_plot_widget.set_image_index_range_label(*self.model.tof_range)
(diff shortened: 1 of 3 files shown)


MikeSullivan7 (changes requested):
When you have multiple ROIs but some are disabled, they no longer redraw when you move an activate ROI or click on the table. However, if you click on an inactive ROI on the table and then resize it via the ROI Properties box, the spectrum is drawn when it should not be. Perhaps we should disable the Properties box if the selected ROI is inactive?

image

samtygier-stfc:
Hmm. I have been playing with some more significant changes to how the redraws are triggered (motivated by reducing unneeded redraws). I think it might be best to roll that into this PR and try to fix it all at once.

samtygier-stfc:
The above bug should now be fixed.

It does not solve the 3 recalculations of the spectrum on window opening, but I think that will be be dealt with in a separate issue.

PR #2541, opened 2025-03-31
Fix COR/Tilt display not updating after manual refinement

### LINKED ISSUE #2531: Bug: Top COR/Tilt display not updating after Manual or Minimize Error refinement
### Summary

The top COR/Tilt display does not automatically update when new values are set using the Refinement or Minimize Error methods. This leads to a mismatch between what is shown to the user and what is being used internally for reconstruction. For accurate values, refer to the COR table instead.

### Steps To Reproduce

Load a dataset in Mantid Imaging.

Go to the COR and Tilt tab.

Run Minimize Error or Manual Refine on slices.

Check the top COR/Tilt display — it does not update.

Run reconstruction — it uses the correct values from the table.

### Current Behaviour

When using Minimize Error or Manual Refinement, the values in the top display do not automatically update, even though the reconstruction internally uses the correct CORs from the table.

### Expected Behaviour

The top COR/Tilt display should be updated to match the refined values from the COR table.

### Context

- The issue occurs consistently across different reconstruction algorithms (FBP, SIRT).
- Verified that reconstructions use the correct (updated) COR values even when the top display does not match.

ashleymeigh2:
## Issue  
Closes #2531

### Description  
Fixes top COR/Tilt display not updating after manual refinement. Presenter now updates the view using set_results().

### Developer Testing  
- Verified unit tests pass: `python -m pytest -vs`
- Manually tested COR refinement updates display

### Acceptance Criteria  
- [ ] COR/Tilt display updates after manual refinement
- [ ] Reconstruction uses correct COR
- [ ] No crashes or errors

### DIFF
--- a/mantidimaging/gui/windows/recon/presenter.py
+++ b/mantidimaging/gui/windows/recon/presenter.py
@@ -321,6 +321,7 @@ class ReconstructWindowPresenter(BasePresenter):
             self.model.data_model.set_cor_at_slice(slice_idx, new_cor.value)
             self.model.last_cor = new_cor
             # Update reconstruction preview with new COR
+            self.view.set_results(*self.model.get_results())
             self.set_preview_slice_idx(slice_idx)
 
     def _do_refine_iterations(self) -> None:


samtygier-stfc (approved):
That looks good now. Thanks

PR #2544, opened 2025-04-04
Flat fielding apply to all stacks histogram levels fix

### LINKED ISSUE #1952: Special case to reset bright levels in operations window after flat-fielding
### Summary

MI defaults to locking brightness levels in the operations window, so that the effect of changes is clear. We already have some special casing for flat-fielding to set the levels in the after panel. We could also adjust the levels of the before panel, after running the flatfielding.

### Steps To Reproduce

Run a flat field operation

### Expected Behaviour

After running, the object should be visible in the before panel

### Current Behaviour

Before panel is still set to the levels before the operation.

### Context
Current main, but not a new issue

Despiix:
### Description
MI defaults to locking brightness levels in the operations window, so that the effect of changes is clear. We already have some special casing for flat-fielding to set the levels in the after panel. It should also adjust the levels of the before panel, after running the flatfielding. 

Instead of running only for flat fielding,it should run for any filter

Steps To Reproduce
Run a flat field operation

Expected Behaviour
After running, the object should be visible in the before panel

### Developer Testing 

- [x] I have verified unit tests pass locally: `python -m pytest -vs`

- [ ] Works as mentioned above

### DIFF
--- a/docs/release_notes/next/fix-1952-Fix-Brightness-levels-when-applying-flat-fielding
+++ b/docs/release_notes/next/fix-1952-Fix-Brightness-levels-when-applying-flat-fielding
@@ -1,0 +1,1 @@
+#1952: Fix brightness levels when applying flat fielding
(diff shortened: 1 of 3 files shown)


samtygier-stfc (approved):
Thanks. Works well.

(I tweaked the realease note)

PR #2546, opened 2025-04-04
add image view to spectrum plot

### LINKED ISSUE #2389: Fitting Window: Add/Create an Image View widget
Fitting Window: Add/Create an Image View widget which displays the sample Image View in the top right corner of the Spectrum Line Plot.

The Image view should not be fixed in location such that a user can move the Image view around the spectrum line plot

Note that displaying the ROI will be completed as part of a separate issue.

Image

ashleymeigh2:
### Description
- Added a floating sample image view to FittingDisplayWidget.
- Updated presenter to pass the averaged image to the widget.

### Developer Testing 
- Manually tested that the floating image view remains visible and draggable.

- [ ] Unit tests pass locally
- [ ] Floating image view displays only the image and remains visible when moved

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
@@ -1,8 +1,7 @@
 # Copyright (C) 2021 ISIS Rutherford Appleton Laboratory UKRI
 # SPDX - License - Identifier: GPL-3.0-or-later
 import numpy as np
-from PyQt5.QtWidgets import QWidget, QVBoxLayout
-from PyQt5.QtWidgets import QGraphicsItem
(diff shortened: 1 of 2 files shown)


samtygier-stfc (changes requested):
I think it needs to be something like
```
        self.image_item = ImageItem()
        self.image_item.setFlag(QGraphicsItem.ItemIsMovable, True)
        self.image_item.setFlag(QGraphicsItem.ItemIsSelectable, True)
        self.image_item.setParentItem(self.spectrum_plot.spectrum)
        self.image_item.setZValue(20)
        self.image_item.setPos(self.spectrum_plot.width()-150, 10)
```
So that it does not move when you pan or zoom on the plot.

samtygier-stfc (approved):

PR #2551, opened 2025-04-15
Display Selected ROI within Image view

### LINKED ISSUE #2390: Fitting Window: Display Selected ROI within Image view
Display the selected ROI from the dropdown menu within the image view.

Displaying the ROI will allow users to know where their selected ROI is over the sample as a quick reference without needing to navigate the the full image window. 

If the selected ROI from the downtown menu changes, this should update the ROI within the Image view. 

Image

Image

ashleymeigh2:
### Description

This PR adds support for displaying the selected Region of Interest (ROI) as an overlay on the thumbnail image in the fitting window. The ROI overlay:
- Matches the position and size of the ROI shown in the main image window
- Adopts the same color as the selected ROI
- Updates live when an ROI is selected or moved

This improves the user experience by making it easier to visually identify which part of the image the fitting corresponds to, directly from the fitting window.

### Developer Testing 

- Manually tested ROI selection via the dropdown
- Verified that the thumbnail image ROI overlay matches the full image ROI
- Verified the ROI overlay updates live on ROI movement

- [ ] Select an ROI from the dropdown — a green box appears on the thumbnail image, matching the main image ROI
- [ ] Move or resize the ROI in the main image view — the thumbnail overlay updates live
- [ ] ROI color changes are reflected in the thumbnail overlay
- [ ] Thumbnail ROI remains accurate when switching datasets or tabs

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
@@ -3,7 +3,7 @@
 import numpy as np
 from PyQt5.QtWidgets import QWidget, QVBoxLayout, QGraphicsItem
 from pyqtgraph import RectROI, mkPen, ImageItem
-from mantidimaging.gui.windows.spectrum_viewer.spectrum_widget import SpectrumPlotWidget
+from mantidimaging.gui.windows.spectrum_viewer.spectrum_widget import SpectrumPlotWidget, SpectrumROI
 
 
 class FittingDisplayWidget(QWidget):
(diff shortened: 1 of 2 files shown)


samtygier-stfc (changes requested):
Looking good.
Can you stop the resize handle (diamond) showing. Might work to set resizable to false or something like that.
Also looks like the scaling is not right. I think you might not need to do the scaling in `show_roi_on_thumbnail_from_widget`

samtygier-stfc (approved):
Works well here. Thanks

PR #2552, opened 2025-04-15
Change `Apply to all stacks` to `Apply to dataset`

### LINKED ISSUE #1329: Change "Apply to ALL Stacks" to "Apply to entire dataset" or something similar
### Current Behaviour

The user can choose to apply an operation either to a single image stack, or all stacks.

It would be more useful offer apply to stack or apply to dataset.

Despiix:
### Description
Updated the functionality of the previous `Apply to all stacks` button to only apply to the current selected dataset instead of all stacks open.

### Developer Testing 

- [ ] Verify that the filters were only applied to the currently selected dataset


samtygier-stfc (approved):
Code looks good and works well for me.

PR #2553, opened 2025-04-24
Speedup of Spectrum Viewer startup

### LINKED ISSUE #2523: Avoid multiple calculation of the spectrum when opening spectrum viewer
### Summary

The spectrum view window is slow to load. This is because calls `model.get_spectrum()` 3 times when it should only be needed once.

It would also be good for it to allow the window to be drawn before calculating the spectrum.

### Steps To Reproduce

Enable to performance log, Open a large  (e.g. 2kx2kx1k dataset).
Open the spectrum viewer
You'll see something like
in the logs.

Put a print statement in `SpectrumViewerWindowModel.get_spectrum()` and see that it gets called 3 times.

### Current Behaviour

Slow opening of window

### Expected Behaviour

Fast opening of window

### Context

Current main

MikeSullivan7:
### Description

To improve the performance of the Spectrum Viewer on first opening, a small cache of recently used spectrums is stored to reduce the number of unnecessary spectrum calculations. The spectrum of the images over the whole image size is permanently stored so it does not need to be recalculated when new ROIs are created.
Previously, the GUI would wait until the dropdown menus for Stack selection are configured and all details are calculated, leading to a long wait time before the Spectrum Viewer window is displayed, this is particularly a problem for large datasets. Now the GUI is forced to process and show before the stacks are configured, giving better responsiveness.

### Developer Testing 

- I have verified unit tests pass locally: `make check`

- [ ] Unit tests pass locally: `make check`
- [ ] To compare against current behaviour, on the `main` branch, load in a dataset and open the Spectrum Viewer. Take note of the amount of time it takes to open and display the window.
- [ ] Switch to this PR and repeat the same steps, and confirm that the window displays faster.
- [ ] Check that the Spectrum Viewer behaves as normal otherwise.

### Current results on main branch for (571 x 2048 x 2048) dataset

```
2025-04-28 16:40:23,991 [perf.mantidimaging.gui.mvp_base.view:L65] INFO: SpectrumViewerWindowView shown in 6.390999999999622
```

### Current results on PR branch for (571 x 2048 x 2048) dataset

```
2025-04-28 16:43:00,177 [perf.mantidimaging.gui.mvp_base.view:L65] INFO: SpectrumViewerWindowView shown in 0.6089999999967404
```

### DIFF
--- a/docs/release_notes/next/fix-2523-Improve-spectrum-viewer-startup
+++ b/docs/release_notes/next/fix-2523-Improve-spectrum-viewer-startup
@@ -1,0 +1,1 @@
+2523: The Spectrum Viewer configures stack and calculates spectrum after window is displayed. Spectrum is now cached.
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -210,12 +210,7 @@ class SpectrumViewerWindowModel:
         return ""
 
     def get_spectrum(self, roi: SensibleROI, mode: SpecType, normalise_with_shuttercount: bool = False) -> np.ndarray:
-        #print("=-=-=-=-=-=-=-=-=-=-=-=- get_spectrum CALLED =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-")
-        # print(*roi)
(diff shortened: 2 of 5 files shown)


samtygier-stfc (approved):
Looks good. Nice speedup.

PR #2558, opened 2025-04-29
Use np.nanmin/max to handle NaN values in fitting ROI positioning

### LINKED ISSUE #2556: Handle NaNs in Data when placing the fitting region
### Current Behaviour

Currently, if there are NaN (Not a Number) values in the spectrum data, they cause the default position of the fitting ROI to have NaNs

### Desired Behaviour

Sensible position of the ROI even with NaNs

### Importance of Feature

Handling NaNs will improve the visual clarity of fitting plots, and help prevent issues when applying fitting algorithms later. It also provides a better user experience when inspecting data.

ashleymeigh2:
### Description

Replace np.min/max with np.nanmin/max in set_default_region to properly handle NaN values in the input data when positioning the fitting ROI. This makes the ROI positioning more robust when working with datasets that contain missing or invalid values.

### Developer Testing
- Tested with sample data containing NaN values to verify ROI positioning works correctly
- Verified existing behavior remains unchanged for data without NaNs
- Checked that ROI dimensions and positioning constraints are still respected

- [x] Load a dataset with NaN values and verify the fitting ROI positions correctly
- [x] Confirm the ROI can still be:
    - [ ] Moved manually after initial positioning
    - [ ] Resized using the corner handles
    - [ ] Updated via `set_selected_fit_region`

- [x] Verify ROI positioning is consistent with these cases:
    - [ ] Data containing all valid numbers
    - [ ] Data with some NaN values
    - [ ] Data with NaN values at the edges

- [x] Check that minimum size constraints (x_span ≥ 20.0, y_span ≥ 10.0) are still enforced
- [x] Verify y_start remains non-negative as per the original implementation

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
@@ -63,8 +63,8 @@ class FittingDisplayWidget(QWidget):
 
     def set_default_region(self, x_data: np.ndarray, y_data: np.ndarray) -> None:
         """Position the ROI centrally over the plotted data."""
-        x_min, x_max = float(np.min(x_data)), float(np.max(x_data))
-        y_min, y_max = float(np.min(y_data)), float(np.max(y_data))
+        x_min, x_max = float(np.nanmin(x_data)), float(np.nanmax(x_data))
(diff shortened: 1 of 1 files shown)


samtygier-stfc (approved):
Looks good and fixes the issue for me. I'll squash the commits because the 2nd undoes a change in the first

samtygier-stfc (approved):

PR #2562, opened 2025-05-06
Refactor `set_default_region` to use dynamic spans based on data range

ashleymeigh2:
### Description

Refactored the `set_default_region` method to calculate region spans (x_span and y_span) dynamically based on the range of input data. This eliminates hardcoded values, making the method adaptable to both normalized and non-normalized data. The ROI is now centrally positioned using a proportion of the data range with fallback spans determined directly from the data.

### Developer Testing
- Verified that the ROI adjusts and centers correctly for datasets with varying ranges, including normalized data.
- Reviewed functionality to ensure no regressions in behavior.

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
@@ -67,11 +67,10 @@ class FittingDisplayWidget(QWidget):
         """Position the ROI centrally over the plotted data."""
         x_min, x_max = float(np.nanmin(x_data)), float(np.nanmax(x_data))
         y_min, y_max = float(np.nanmin(y_data)), float(np.nanmax(y_data))
-        x_span = max((x_max - x_min) * 0.25, 20.0)
-        y_span = max((y_max - y_min) * 0.5, 10.0)
+        x_span = (x_max - x_min) * 0.25
+        y_span = (y_max - y_min) * 0.5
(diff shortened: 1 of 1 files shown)


mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
```diff
@@ -67,11 +67,10 @@ def set_default_region(self, x_data: np.ndarray, y_data: np.ndarray) -> None:
         """Position the ROI centrally over the plotted data."""
         x_min, x_max = float(np.nanmin(x_data)), float(np.nanmax(x_data))
         y_min, y_max = float(np.nanmin(y_data)), float(np.nanmax(y_data))
-        x_span = max((x_max - x_min) * 0.25, 20.0)
-        y_span = max((y_max - y_min) * 0.5, 10.0)
+        x_span = max((x_max - x_min) * 0.25, (x_max - x_min) * 0.05)
+        y_span = max((y_max - y_min) * 0.5, (y_max - y_min) * 0.05)
```
samtygier-stfc, line 71: I think this could just be
```python
x_span = (x_max - x_min) * 0.25
y_span = (y_max - y_min) * 0.5
```
Maybe make them both 0.25 for consistency

samtygier-stfc (changes requested):

samtygier-stfc (approved):
Thanks. Fixes the issue here.

(I squashed the commits)

PR #2574, opened 2025-05-16
Add fitting spectrum updates after spectrum modifications

### LINKED ISSUE #2523: Avoid multiple calculation of the spectrum when opening spectrum viewer

### LINKED ISSUE #2563: Fitting Display Fails to Update Properly After Adjustments
### Summary

The fitting display is not updating correctly when adjustments are made to the ROI or fitting parameters. Users expect the display to refresh and show updated fitting results in real time, but currently, the updates seem to either not trigger or have delays.

ashleymeigh2:
### Description

Ensure the fitting spectrum is updated consistently when ROI spectra are set.  
This resolves potential discrepancies in the fitting display — particularly the initial roi
Now updates fitting spectrum immediately on adding initial `"roi"` or `"rits_roi"`.

### Developer Testing 

- Confirmed fitting display updates correctly with one ROI
- Tested with both normal and RITS ROI

- [x] Fitting display updates immediately with first ROI
- [x] Spectrum and fit region display correctly
- [x] Normal and RITS ROIs behave consistently
- [x] No regression in ROI handling

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/roi_selection_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/roi_selection_widget.py
@@ -23,6 +23,7 @@ class ROISelectionWidget(QtWidgets.QGroupBox):
         layout.setContentsMargins(10, 5, 10, 5)
         layout.setSpacing(2)
 
+        self.current_roi_name = None
         self.roiDropdown = QtWidgets.QComboBox(self)
         self.roiDropdown.currentIndexChanged.connect(self._on_selection_changed)
         layout.addWidget(self.roiDropdown)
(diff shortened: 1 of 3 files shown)


samtygier-stfc (changes requested):
Can you also handle the case of updating the fitting spectrum when the ROI is moved.

samtygier-stfc (approved):
Looks good. Working locally.

PR #2590, opened 2025-05-30
Fit Bragg Edge for all ROIs

### LINKED ISSUE #2586: Method to run fit on all ROIs
### Current Behaviour

Fitting is run manually on a single ROI

### Desired Behaviour

Button on Export tab, to run with the current initial parameters against all ROIs and show fitted parameters in the table

### Importance of Feature

Key part of fitting workflow

samtygier-stfc:
### Description

Add a Fit All button to the export tab that will run the fit against all ROIs
Small refactoring, split `run_region_fit()`, type alias for clarity.

### Developer Testing 

- ...

- [ ] Manual test
  - [ ] Load a ToF dataset into the spectrum viewer
  - [ ] Select several ROIs
  - [ ] On the fitting tab, position the fitting region and get initial parameters
  - [ ] On the export tab press "Fit All". The table should be populated with a row for each ROI, with sensible parameters (assuming a good fit is possible)

### DIFF
--- a/mantidimaging/core/fitting/fitting_functions.py
+++ b/mantidimaging/core/fitting/fitting_functions.py
@@ -4,11 +4,17 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from math import sqrt
+from typing import NamedTuple
 
 import numpy as np
 from scipy.special import erf
 
-FittingRegionType = tuple[float, float, float, float]
+
+class FittingRegion(NamedTuple):
+    x_min: float
+    x_max: float
+    y_min: float
+    y_max: float
 
 
 class BaseFittingFunction(ABC):
@@ -18,7 +24,7 @@ class BaseFittingFunction(ABC):
         return list(self.parameter_names)
 
     @abstractmethod
-    def get_init_params_from_roi(self, region: FittingRegionType) -> dict[str, float]:
+    def get_init_params_from_roi(self, region: FittingRegion) -> dict[str, float]:
         ...
 
     @abstractmethod
@@ -34,7 +40,7 @@ class ErfStepFunction(BaseFittingFunction):
         y = h * 0.5 * (1 + erf((xdata - mu) / (sigma * sqrt(2)))) + a
         return y
 
-    def get_init_params_from_roi(self, region: FittingRegionType) -> dict[str, float]:
+    def get_init_params_from_roi(self, region: FittingRegion) -> dict[str, float]:
         x1, x2, y1, y2 = region
         init_params = {
             "mu": (x1 + x2) / 2,
--- a/mantidimaging/core/fitting/fitting_functions.py
+++ b/mantidimaging/core/fitting/fitting_functions.py
@@ -8,6 +8,8 @@ from math import sqrt
 import numpy as np
 from scipy.special import erf
 
+FittingRegionType = tuple[float, float, float, float]
+
 
 class BaseFittingFunction(ABC):
     parameter_names: list[str]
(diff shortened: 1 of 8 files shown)


mantidimaging/core/fitting/fitting_functions.py
```diff
@@ -8,6 +8,8 @@
 import numpy as np
 from scipy.special import erf
 
+FittingRegionType = tuple[float, float, float, float]
```
JackEAllen, line 11: Nice use of a type alias! If the elements in the tuple have meaning, would using a `NamedTuple` or a `dataclass` improve clarity, maybe something like this:
```Python
class FittingRegion(NamedTuple):
    x_min: float
    x_max: float
    y_min: float
    y_max: float
```

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -580,18 +581,34 @@ def show_initial_fit(self):
...
+        result = self.fit_single_region(self.fitting_spectrum, self.view.get_fitting_region(), self.model.tof_data,
+                                        self.view.scalable_roi_widget.get_initial_param_values())
 
-        result = self.model.fitting_engine.find_best_fit(xvals, yvals, init_params)
         self.view.scalable_roi_widget.set_fitted_parameter_values(result)
         self.show_fit(list(result.values()))
         roi_name = self.view.roiSelectionWidget.current_roi_name
         self.view.exportDataTableWidget.update_roi_data(roi_name=roi_name, params=result, status="Fitted")
 
+    def fit_single_region(self, spectrum: np.ndarray, fitting_region: FittingRegionType, tof_data: np.ndarray,
```
JackEAllen, line 592: This may not be the best place to add extra validation. It may be better to extract into a new method if the below issues are applicable in `run_region_fit`, fit_single_region` and `fit_all_regions` - Do you think it would be worthwhile to check that:
* `tof_data, `spectrum` and `fitting_region` are valid and non-empty?
* Can the `fitting_region` be outside of `tof_data`? - if so should we add a check for this, maybe by altering how the start and end indices in `tof_data` are found to:
```Python
    start, stop = np.searchsorted(tof_data, (fitting_region[0], fitting_region[1]))
    if start == stop:
        raise ValueError("Fitting region does not overlap with tof_data.")

    xvals = tof_data[start:stop]
    yvals = spectrum[start:stop]
```
* Is there a possibility of running into zero-length slices where if the resulting slice in the `fitting_region` falls outside the range of `tof_data` or if both values map to the same index in `tof_data` where the region is too narrow or boundaries are equal? if so, should we consider a check for this? - just something like `if len(xvals) == 0 or len(yvals) == 0:`?

JackEAllen:
Actually, as this is a GUI change, it's probably worth adding release notes.

mantidimaging/gui/widgets/spectrum_widgets/export_settings_widget.py
```diff
@@ -13,6 +13,10 @@ def __init__(self, parent=None):
         super().__init__(parent)
 
         layout = QVBoxLayout(self)
+
+        self.fit_all_button = QPushButton("Fit All")
```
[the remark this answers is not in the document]
samtygier-stfc: Good to call it out. We should maybe have a full decision on case conventions.
We generally use PEP8, but with some pragmatism because pyQt uses Qt conventions and sometimes we don't have a choice. We could decide on:
1) PEP8 everywhere we can
2) PEP8 every except attributes which are QObjects

samtygier-stfc:
> Actually, as this is a GUI change, it's probably worth adding release notes.

We do already have a release note item for "Add a fitting system for Bragg Edges in Time of Flight data" which covers all of this work

JackEAllen (approved):
Retested after your changes and all seems well. Happy to approve. Thank you for adding this in. This feature is cleanly implemented.

PR #2600, opened 2025-06-05
2593 Fix ROI preview not being anchored to the spectrum viewer

### LINKED ISSUE #2593: Spectrum Viewer - Fitting Tab: Sample image Not Pinned to the Top Right Corner on Window Resize
### Summary

When resizing the spectrum viewer window from the default window size when the fitting tab is in focus, the sample image should be pinned to the top right of the window to achieve a consistent layout no matter what size the window is. 
To achieve this, it may be required to place the component in a new `QHBoxLayout` or similar.

### Steps To Reproduce

- Open the spectrum viewer within mantid imaging with a ToF dataset loaded
- Navigate to the fitting window
- Resize the window aspect ratio horizontally - either manually or by maximising the window
- The sample view that displays the location of the ROI no longer stays at the top right of the window and is located more centrally.

### Current Behaviour

The sample ROI view in the fitting tab is not pinned the the top right of the fitting tab window when the window is resized. 

### Expected Behaviour

The sample ROI view in the fitting tab should be pinned to the top right of the window so that the location of the component is consistent when resizing the window. 

### Context

Issue found on Ubuntu during manual testing.

### Screenshot(s)

Sample ROI view is located in the correct location at the top right of window when using the default size of the window on initilisation:

Image

Sample ROI view is no longer located in the correct location at the top right of window when resizing the window (maximised in this example):

Image

hussam-stfc:
### Description

Added new methods to the `MIPlotItem` class to enable anchoring a child to the top-right of the spectrum viewer plot.

This is done by overriding the super `PlotItem`'s `resizeEvent()` callback, adding a call to the new `update_anchor()` function. This function dynamically sets the child object's position to a specified offset from the top right. This offset and the child object are set through a new `set_right_anchored_child()` method.

A call to this function has been added, replacing the child `image_item`'s `setPos()` call as the position is now set dynamically.

### Developer Testing 

- I have manually checked that any form of window change (resize, minimise/maximise, resizing a separate tab) keeps the preview anchored to the top-right corner

- [x] ROI preview remains anchored after any window resize

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 from PyQt5.QtCore import pyqtSignal, Qt, QSignalBlocker, QEvent
-from PyQt5.QtGui import QColor
+from PyQt5.QtGui import QColor, QResizeEvent
 from PyQt5.QtWidgets import QColorDialog, QAction, QMenu, QSplitter, QWidget, QVBoxLayout, QActionGroup
 
 from pyqtgraph import ROI, GraphicsLayoutWidget, LinearRegionItem, PlotItem, mkPen, ViewBox, PlotDataItem, GraphicsItem
@@ -393,7 +393,7 @@ class MIPlotItem(PlotItem):
 
         self.base_menu.addMenu(self.plot_menu)
 
-        self.anchored_child = None
+        self.anchored_child: None | GraphicsItem = None
         self.anchored_offset: None | tuple[int, int] = None
 
     def plot(self, *args, **kwargs) -> PlotDataItem:
@@ -425,7 +425,7 @@ class MIPlotItem(PlotItem):
         self.anchored_offset = offset
         self.update_anchor()
 
-    def resizeEvent(self, event) -> None:
+    def resizeEvent(self, event: QResizeEvent) -> None:
         super().resizeEvent(event)
         self.update_anchor()
 


mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
```diff
@@ -393,6 +393,9 @@ def __init__(self, *args, **kwds) -> None:
 
         self.base_menu.addMenu(self.plot_menu)
 
+        self.anchored_child = None
```
samtygier-stfc, line 396: This can be `self.anchored_child: None | GraphicsItem = None`

mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
```diff
@@ -410,3 +413,24 @@ def set_join_plot(self) -> None:
...
+        Anchor a child object to the top right of this item
+
+        @param child: child object
+        @param offset: tuple (x,y) offset from the top right
+        """
+        self.anchored_child = child
+        self.anchored_offset = offset
+        self.update_anchor()
+
+    def resizeEvent(self, event) -> None:
```
samtygier-stfc, line 428: The event argument should have an annotation
`def resizeEvent(self, event: QResizeEvent) -> None:`

samtygier-stfc (changes requested):
Looks good and resolves the issue. Just a couple of type annotation tweaks.

mantidimaging/gui/windows/spectrum_viewer/spectrum_widget.py
hussam-stfc: Fixed through commit 42ab3a1

samtygier-stfc (approved):
Thanks for the fixes

PR #2620, opened 2025-06-13
2603 tof data type revision

### LINKED ISSUE #2603: Spectrum Viewer: Refactor tof_data Type to Never be None.
### Current Behaviour

`tof_data` is currently set to be `np.ndarray | None`. This means whenever `tof_data` is used, we need to first `assert tof_data is not None`. I suspect that we will use `tof_data` regularly going forward. 

I would like to propose that we refactor `tof_data` to be `np.ndarray` only, by having a default state of `np.array([])` to remove the need to assert if `tof_data` is not `None` in every method it is used going forward as was the case for [504969b](https://github.com/mantidproject/mantidimaging/pull/2601/commits/504969b34b9dc2b9f41bd955f97b197768b2e013)

### Desired Behaviour

The default type of `tof_data` is `np.ndarray` to remove the need to `assert tof_data is not None` when used in any method.

### Undesired Behaviour

`tof_data` is currently set to be `np.ndarray | None` requiring `assert tof_data is not None` checks wherever it is used. 

### Importance of Feature

Minor codebase optimisation.

### Additional Context

`tof_data` is located at: [mantidimaging/gui/windows/spectrum_viewer/model.py#L94](https://github.dev/mantidproject/mantidimaging/blob/16e5a2010ba8699f9fa11e2b00292b2ba7134e27/mantidimaging/gui/windows/spectrum_viewer/model.py#L94)

JackEAllen:
### Description

Refactor `tof_data` type from `tof_data: np.ndarray | None = np.array([])` to `tof_data: np.ndarray = np.array([])` to avoid the need to `assert self.model.tof_data is not None` going forward as `tof_data` will likely increase in use as the spectrum viewer is further developed. 
This change included changes to any references to `tof_data` in `model.py`, `presenter.py`, `view.py`, associated tests and added protection against `RuntimeWarning: divide by zero encountered in divide` by refactoring `check_data`.

### Developer Testing 

- I have inspected the spectrum viewer to ensure all referenced to `tof_data` including indirect references are updated to handle empty arrays instead of None.
- Tested basic workflows within the spectrum viewer

- [ ] Check all direct and indirect references to `tof_data` are treated as `np.array([])` and not `None`
- [ ] Test a variety of workflows within the spectrum viewer with a ToF dataset.

Documentation not needed as this change is not user facing.

### DIFF
--- a/mantidimaging/core/utility/unit_conversion.py
+++ b/mantidimaging/core/utility/unit_conversion.py
@@ -42,7 +42,10 @@ class UnitConversion:
 
     def check_data(self) -> None:
         try:
-            self.velocity = self.target_to_camera_dist / (self.tof_data_to_convert + self.data_offset)
+            tof_with_offset = self.tof_data_to_convert + self.data_offset
+            if np.any(tof_with_offset == 0):
+                raise ValueError("Time-of-flight (with offset) contains zero(s), which is likely invalid.")
+            self.velocity = self.target_to_camera_dist / tof_with_offset
         except AttributeError as exc:
             raise TypeError("No data to convert") from exc
 
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -91,7 +91,7 @@ class SpectrumViewerWindowModel:
     tof_range: tuple[int, int] = (0, 0)
     tof_plot_range: tuple[float, float] | tuple[int, int] = (0, 0)
     tof_mode: ToFUnitMode = ToFUnitMode.WAVELENGTH
-    tof_data: np.ndarray | None = np.array([])
+    tof_data: np.ndarray = np.array([])
     spectrum_cache: dict[tuple, np.ndarray] = {}
 
     def __init__(self, presenter: SpectrumViewerWindowPresenter):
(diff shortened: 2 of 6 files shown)


mantidimaging/core/utility/unit_conversion.py
```diff
@@ -42,7 +42,11 @@ def set_data_to_convert(self, data_to_convert: np.ndarray) -> None:
 
     def check_data(self) -> None:
         try:
-            self.velocity = self.target_to_camera_dist / (self.tof_data_to_convert + self.data_offset)
+            tof_with_offset = self.tof_data_to_convert + self.data_offset
+            self.velocity = np.divide(self.target_to_camera_dist,
+                                      tof_with_offset,
+                                      out=np.full_like(tof_with_offset, np.inf, dtype=np.float64),
+                                      where=tof_with_offset != 0)
```
samtygier-stfc, line 49: Is there a case where time is zero? I wonder if we should be raising an error for that case as it probably means something has gone wrong.

samtygier-stfc (changes requested):

JackEAllen:
This will fail until I have rebased the changes proposed by #2620 when merged in

samtygier-stfc (dismissed):
Looks good.

JackEAllen:
> But looks like it needs some test fixes

Yes, I have a fix ready to push, but will sadly have to wait until VPN is restored

samtygier-stfc (approved):

PR #2623, opened 2025-06-16
Fix units for set_centre_of_rotation() call

### LINKED ISSUE #2621: Reconstruction: COR issue with CIL Recon
### Summary

There appears to be an issue with the center of rotation being used in CIL reconstruction, resulting in back outputs.

samtygier-stfc:
### Description
When call `self.set_centre_of_rotation()` pass the angle units explicitly. This prevents the angle being interpreted as radians.

### Developer Testing 

- [x] Follow the steps on #2621, and confirm that the reconstruction with CIL matches FBP

Fix for recent regression, so release note not needed


hussam-stfc (approved):
Both CUDA and CIL reconstructions look very similar and all tests have passed.

PR #2633, opened 2025-06-18
Fix RuntimeError in recolor_links when layout is deleted

### LINKED ISSUE #2628: RuntimeError in recolor_links()
### Current Behaviour

Occasional RuntimeError: wrapped C/C++ object of type QGridLayout has been deleted when shutting down or refreshing UI.

### Expected Behaviour

recolor_links() should safely handle missing layouts and avoid crashing.

ashleymeigh2:
## Issue
Closes #2628

### Description
Prevent `RuntimeError` in `recolor_links()` by checking if `link_box_layout` exists and is not deleted before accessing it.

### Developer Testing 

- [ ] Verify UI can be opened and closed without error
- [ ] Confirm no crash occurs on theme refresh or shutdown

### DIFF
--- a/mantidimaging/gui/windows/welcome_screen/view.py
+++ b/mantidimaging/gui/windows/welcome_screen/view.py
@@ -54,6 +54,8 @@ class WelcomeScreenView(QWidget):
 
         self.close_button = CloseButton(self)
 
+        self.close_button.clicked.connect(self.close_welcome_screen)
+
         # Done to make sure the button appears in the top-right corner after rendering
         QTimer.singleShot(1, self.position_close_button)
 


JackEAllen (changes requested):
I think we can probably target the cause of the issue to remove the need to have this check by ensuring that when the welcome screen is closed, all references to it and it's presenter are set to `None` in the main window by adding `self.welcome_screen.closed.connect(self.close_welcome_screen) ` `mantidimaging/gui/windows/main/view.MainWindowView.create_welcome_screen()` after initialising ` self.welcome_screen` and `self.welcome_presenter = None` to the bottom of `mantidimaging/gui/windows/main/view.MainWindowView.close_welcome_screen()`
and to handle things from the `WelcomeScreenView` side of things, we can add the signal `closed = pyqtSignal()` directly under `mantidimaging/gui/windows/welcome_screen/view.WelcomeScreenView` and refactor `mantidimaging/gui/windows/welcome_screen/view.WelcomeScreenView.close_welcome_screen()` to no longer call it's parent directly which i believe breaks encapsulation  where the welcome screen should only manage it's own state and not make assumptions about it's parent's lifecycle. and instead let the welcome screen close itself and emit a signal for the main window to listen out for:
```python
def close_welcome_screen(self):
    self.close()
    self.closed.emit()
```

mantidimaging/gui/windows/welcome_screen/view.py
```diff
@@ -91,3 +93,7 @@ def add_issues(self, issues_text: str) -> None:
                                         "  margin-top: 5px; "
                                         "}")
         self.issues_label.setText(issues_text)
+
```
JackEAllen, line 96: we can replace:
```python
def close_welcome_screen(self):
    self.parent().close()
```
with
```python
def close_welcome_screen(self):
    self.close()
    self.closed.emit()
```
so that we no longer call it's parent directly which I believe breaks encapsulation.

mantidimaging/gui/windows/welcome_screen/presenter.py
```diff
@@ -56,6 +56,7 @@ def recolor_links(self):
         Removes existing link labels and re-adds them with the new theme color logic.
         """
         layout = self.view.link_box_layout
+
```
JackEAllen, line 59: Is this intentional?

docs/release_notes/next/fix-2682-RuntimeError-in-recolor_links
```diff
@@ -0,0 +1 @@
+2633: Fixed an error where closing the Welcome Screen could raise an AttributeError: 'WelcomeScreenView' object has no attribute 'closed'. T
```
JackEAllen, line 1: Possibly a typo?

PR #2636, opened 2025-06-18
Notify users of previous crash and provide access to log file

### LINKED ISSUE #2366: Warning Dialog MI shutdown was Not Clean and Point to Logs
### Desired Behaviour

We should open a warning dialog window if shutdown of MI was not clean. the warning dialog should also provide a link pointing to the log files produced during a non-clean shutdown of MI.

Providing user with easy identification of a non-clean shutdown of MI and providing users with easy access to the logs produced during the shutdown will improve user bug reports of such instances 

### Current Behaviour

Mantid Imaging does not warn of a non-clean shutdown to users on closure.

ashleymeigh2:
Closes #2366

### Description
- Show warning dialog if previous shutdown was not clean
- Display latest log file path if available, fallback message if not
- Track shutdown state via QSettings

### Developer Testing
- Verified unit tests pass: `python -m pytest -vs`
- Manually tested warning appears on forced crash
- Verified fallback when no log file exists

### Reviewer Checks
- [x] Unit tests pass
- [x] Warning shows after non-clean shutdown
- [x] No warning on clean shutdown
- [x] Graceful message when no logs available

### Notes
- [x] Release notes updated

### DIFF
--- a/mantidimaging/main.py
+++ b/mantidimaging/main.py
@@ -10,6 +10,7 @@ import os
 from pathlib import Path
 
 import darkdetect
+from PyQt5.QtCore import QSettings
 from PyQt5.QtWidgets import QApplication, QMessageBox
 from PyQt5 import QtCore
 from PyQt5.QtGui import QGuiApplication, QFont, QFontInfo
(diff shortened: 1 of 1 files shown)


mantidimaging/main.py
```diff
@@ -102,26 +102,49 @@ def main() -> None:
...
+
+        msg = QMessageBox()
+        msg.setIcon(QMessageBox.Warning)
+        msg.setWindowTitle("Previous Shutdown Warning")
+        if latest_log and latest_log.exists():
+            msg.setText(f"Mantid Imaging did not shut down cleanly last time.\n\nLog file:\n{latest_log}")
+        else:
+            msg.setText("Mantid Imaging did not shut down cleanly last time.\n\nNo log file is available.")
+        msg.setStandardButtons(QMessageBox.Ok)
+        msg.exec_()
```
samtygier-stfc, line 123: Can this all be put into a function.

Also the second message could say add "To record log files to disk, configure a directory in File -> Settings"

mantidimaging/main.py
```diff
@@ -102,26 +102,49 @@ def main() -> None:
...
     process_count = settings.value("multiprocessing/process_count", 8, type=int)
 
     from mantidimaging import gui
     try:
         pm.create_and_start_pool(process_count)
         gui.execute()
         result = q_application.exec_()
+        settings.setValue("app/last_shutdown_clean", True)
     except BaseException as e:
+        settings.setValue("app/last_shutdown_clean", False)
```
samtygier-stfc, line 147: Can this be moved to be like
```
settings.setValue("app/last_shutdown_clean", False)
pm.create_and_start_pool(process_count)
gui.execute()
result = q_application.exec_()
settings.setValue("app/last_shutdown_clean", True)
```
In case there is some level of crash that bypasses the exception handling.

Basically the value us always set to `False` while MI is running, and only gets set `True` when successfully shutting down.

samtygier-stfc (changes requested):

mantidimaging/main.py
```diff
@@ -93,42 +94,66 @@ def setup_application() -> QApplication:
...
 
     from mantidimaging import gui
     try:
         pm.create_and_start_pool(process_count)
         gui.execute()
         result = q_application.exec_()
-    except BaseException as e:
-        if sys.platform == 'linux':
-            pm.clear_memory_from_current_process_linux()
-        raise e
```
samtygier-stfc, line 127: The `except` is still needed to call the `clear_memory_from_current_process_linux()` function

samtygier-stfc (changes requested):

samtygier-stfc:
Working well. Tested with modifying the qsetting file and by using `xkill` to kill a running MI.

PR #2644, opened 2025-06-20
Spectrum Viewer: Santisteban fits Bragg Edge extremes using ydata percentiles

### LINKED ISSUE #2642: Spectrum Viewer: Changing ROI and rerunning can crash window (Santisteban)
### Summary

In the Fitting Tab, when using the Santisteban fit it is possible to crash the window by moving the ROI and running a fit if the `From ROI` button is not pressed first. This is possibly due to the left and right side fitting never converging.

### Steps To Reproduce

1. Load in dataset (I used the corrected Fe dataset) and open the spectrum viewer
2. Place the ROI over a Bragg Edge, run From ROI and Run Fit to get a good fitting.
3. Move the ROI to an area which doesnt fit as well (e.g. some noise) and then press `Run fit` without pressing `From ROI` (see example screenshot below)
4. This should lead to the window becoming unresponsive and crashing

### Current Behaviour

The window becomes unresponsive and crashes when a good fit cannot be found

### Expected Behaviour

MI should warn if a good fit cannot be found, or display a bad fit at the least.

### Screenshot(s)

Image

MikeSullivan7:
### Description

When performing the left and right side fittings, lower and upper percentiles of the ydata within the ROI are used to consistently produce the same fit for a given Bragg Edge. Extending the horizontal width of the ROI now decides how much of the ydata before and after the Edge is used in the statistics to find the fitting of the additional parameters.
This method now produces better and consistent fits, even for sparser/lower resolution data.

For poorly chosen ROIs, the use is warned to resize the ROI via a custom exception, therefore preventing crashes/infinite loops.

### Developer Testing 

- [ ] follow steps in linked error and make sure that no errors occur and that a fit is always given (good or bad!)

### DIFF
--- a/docs/release_notes/next/fix-2642-Santisteban-fitting-crash
+++ b/docs/release_notes/next/fix-2642-Santisteban-fitting-crash
@@ -1,0 +1,1 @@
+#2642: While using the Santisteban fit, the window will not crash if a bad fit is found or a poor ROI is chosen. The user receives a warning if a poor ROI is used.
--- a/mantidimaging/core/fitting/fitting_functions.py
+++ b/mantidimaging/core/fitting/fitting_functions.py
@@ -12,11 +12,22 @@ from scipy.special import erf, erfc
 
 
 class LeftSideFittingException(Exception):
-    pass
+
+    def __init__(self, source_err):
+        print("LeftSideFittingException:", source_err)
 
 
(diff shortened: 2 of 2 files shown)


samtygier-stfc (approved):
This solves the issue of getting stuck in the while loop by moving to loop free code.

PR #2662, opened 2025-06-26
Santisteban tweaks

samtygier-stfc:
### Developer Testing 

- ...

- [ ] ...

### DIFF
--- a/mantidimaging/core/fitting/fitting_functions.py
+++ b/mantidimaging/core/fitting/fitting_functions.py
@@ -11,18 +11,6 @@ from scipy.optimize import curve_fit
 from scipy.special import erf, erfc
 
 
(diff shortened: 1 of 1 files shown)


MikeSullivan7:
Ive tested this on a few datasets and I think this works well! I've not been able to break it even with the loss of exception handling. Do we think that using the 5th and 95th percentiles will be good for every dataset? If so, Im happy to merge this :)

samtygier-stfc:
The only issue I can think of is when there is a small number of elements, but it looks like the 95 percentile will always be less than the max value. So I think it is safe in that there will always be at least 1 point, but for very small ranges it will be less resilient to noise.

```
>>> numpy.percentile([1,2,3,4,5], 95)
4.8
```

PR #2695, opened 2025-06-30
2693 Reconstruction Window: Refactor projection 180 check by moving to ImageStack

### LINKED ISSUE #2693: Reconstruction Window: Refactor projection 180 check by moving to ImageStack
### Current Behaviour

The reconstruction window currently holds some code that performs checks on the ImageStack's 180 degree projection, such as checking for its existence and validity. 

### Desired Behaviour

This code should be moved to the ImageStack to improve encapsulation and reduce some layered code in `view.py` and `model.py`.

### Additional Context

This is being done in preparation for the [EPAC CIL Geometry Work](https://github.com/mantidproject/mantidimaging/issues/2482).

hussam-stfc:
### Description

- Moved the `proj_180_degree_shape_matches_images()` method from `recon/model.py` to `imagestack.py`
- Updated `recon/presenter.py` to call `ImageStack`'s check code instead
- Moved tests from `recon/model_test.py` to `image_stack_test.py`
- Added a new notification handler for when a new `ImageStack` is selected in the `stackSelector` widget
- Updated relevant tests to reflect these changes

### Developer Testing 

- I have manually verified that reconstruction works as expected

- [ ] Verify that reconstruction behaves as expected and that no errors are thrown

### DIFF
--- a/docs/release_notes/next/dev-2693-refactor-180-projection-check
+++ b/docs/release_notes/next/dev-2693-refactor-180-projection-check
@@ -1,0 +1,1 @@
+#2693: Moved 180 degree projection check code from the reconstruction window to the ImageStack


mantidimaging/core/data/imagestack.py
```diff
@@ -237,6 +237,13 @@ def projection(self, projection_idx: int) -> np.ndarray:
         else:
             return self.data[projection_idx]
 
+    def proj_180_degree_shape_matches_images(self) -> bool:
```
JackEAllen, line 240: I think we may be able to drop the extra if statement here and the need to return `False` because `has_proj180deg` already checks `if self.proj180deg is not None`, returning `False` if it is `None`
```Python
def proj_180_degree_shape_matches_images(self) -> bool:
    return (
        self.has_proj180deg() 
        and self.height == self.proj180deg.height
        and self.width == self.proj180deg.width
    )
```

mantidimaging/gui/windows/recon/test/presenter_test.py
```diff
@@ -297,6 +297,26 @@ def test_set_precalculated_cor_tilt(self):
         self.presenter.do_update_projection.assert_called_once()
         self.presenter.do_preview_reconstruct_slice.assert_called_once()
 
+    def test_check_stack_for_invalid_180_deg_proj_when_proj_180_degree_shape_matches_images_is_false(self):
```
JackEAllen, line 300: We can probably condense this with parameterized (i'm quite neutral on this so feel free to modify or leave as is):
```python
    @parameterized.expand([(False, True), (True, False)])
    def test_check_stack_for_invalid_180_deg_proj(self, shape_matches, expect_error):
        uuid = mock.Mock()
        selected_images = self.presenter.main_window.get_stack(uuid)
        selected_images.proj_180_degree_shape_matches_images = mock.Mock(return_value=shape_matches)

        self.presenter.check_stack_for_invalid_180_deg_proj(uuid)
        selected_images.proj_180_degree_shape_matches_images.assert_called_once()

        if expect_error:
            self.view.show_error_dialog.assert_called_once_with(
                "The shapes of the selected stack and it's 180 degree projections do not match! This is going to cause an "
                "error when calculating the COR. Fix the shape before continuing!")
        else:
            self.view.show_error_dialog.assert_not_called()
```

mantidimaging/gui/windows/recon/presenter.py
```diff
@@ -130,7 +134,13 @@ def do_algorithm_changed(self) -> None:
         self.do_preview_reconstruct_slice()
         self.view.change_refine_iterations()
 
-    def set_stack_uuid(self, uuid) -> None:
+    def do_stack_uuid_changed(self) -> None:
```
JackEAllen, line 137: NITPICK: Would the name of this method be more user friendly as `on_stack_uuid_changed`?

mantidimaging/gui/windows/recon/view.py
```diff
@@ -173,7 +172,7 @@ def on_row_change(item, _):
         self.on_table_row_count_change()
 
         self.stackSelector.subscribe_to_main_window(main_window)
-        self.stackSelector.stack_selected_uuid.connect(self.check_stack_for_invalid_180_deg_proj)
+        self.stackSelector.stack_selected_uuid.connect(lambda: self.presenter.notify(PresN.SET_STACK_UUID))
```
JackEAllen, line 175: Not certain on this one so it's more of a question - but if `stack_selected_uuid` emits the UUID, can we connect signals directly to the presenter instead of a lambda, which could be more idiomatic in PyQt and easier to debug where the lambda ignores the signals arguments unless we don't care about them?
```Python
self.stackSelector.stack_selected_uuid.connect(self.presenter.set_stack_uuid)
```

JackEAllen (changes requested):
Functionally seems to work from testing and tests pass locally for me. Thank you for the refactor! 👍 
Just a few small comments and a question , but otherwise good

mantidimaging/core/data/imagestack.py
hussam-stfc: That check is there because mypy keeps complaining about `proj180deg` possibly being `None`, thus may not have `height` and `width` attributes.

mantidimaging/gui/windows/recon/presenter.py
hussam-stfc: Probably so, but I've named it this way for consistency with other notification methods. I'm happy to change it if that isn't important

mantidimaging/gui/windows/recon/view.py
hussam-stfc: Pretty much all the other widgets use a lambda for their connect method, but admittedly I don't understand why and whether this case is different.

PR #2702, opened 2025-07-01
Skip moving the fitting region if there is already data points in it

### LINKED ISSUE #2699: Spectrum Viewer: When switching fitting ROI only resize region if needed
### Current Behaviour

When switching ROI on the fitting tab the fitting region always resets to a default size.

### Desired Behaviour

Often the ROIs have similar spectrums and in these cases moving the fitting region can be annoying.

A sensible heuristic would be to only move the fitting region if there is no data inside it with the new spectrum.

### Undesired Behaviour

The fitting region not being move when it is no longer a good fit.

### Importance of Feature

Solves an annoyance. Makes it much easier to fine tune a the fitting region for multiple ROIs

samtygier-stfc:
### Description

Add a check to see if there are any data points already in the box. If so keep the existing box because its probably fine and might have been carefully placed by the user.

This also changes the behaviour when there is no valid spectrum. Previously an fallback region was made 0,0,1,1. Now it hides the region and re-shows when there is data. This prevents the issue where there is data inside that fall back region, but its not a default region.

### Developer Testing 

- ...

- [ ] ...

Not needed, small behaviour tweak

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/fitting_display_widget.py
@@ -34,7 +34,7 @@ class FittingDisplayWidget(QWidget):
         self.layout.addWidget(self.spectrum_plot)
         self._current_fit_line: PlotDataItem | None = None
 
-        self.fitting_region = RectROI([0, 0], [1, 1], pen=mkPen((255, 0, 0), width=2), movable=True)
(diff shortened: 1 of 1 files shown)


samtygier-stfc:
Visual change on the screenshot tests is due to the turning on normalisation. I think there is an in between where normalisation is on but the open beam is not selected, this means the spectrum has not data so the fallback default 0,0,1,1 region is set. When the correct data arrives some of the points are within that region so it is not reset.

Issue #2687, opened 2025-06-27
Use pathlib instead of os.path


JackEAllen:
Pathlib Path objects are more robust than using `os.path` methods on strings.

Functions receiving a path as an argument should receive a `Path` instead of a `str`.

A search for `os.path`, will find all places that paths are manipulated with the older methods.

For example

```python
def find_first_file_that_is_possibly_a_sample(file_path: str) -> str | None:
    # Grab all .tif or .tiff files
    possible_files = glob.glob(os.path.join(file_path, "**/*.tif*"), recursive=True)

    for possible_file in sorted(possible_files):
        lower_filename = os.path.basename(possible_file).lower()
        if "flat" not in lower_filename and "dark" not in lower_filename and "180" not in lower_filename:
            return possible_file
    return None
```
could be updated to
```python
def find_first_file_that_is_possibly_a_sample(file_path: Path) -> Path | None:
    # Grab all .tif or .tiff files
    possible_files = file_path.glob("**/*.tif*")

    for possible_file in sorted(possible_files):
        lower_filename = possible_file.name.lower()
        if "flat" not in lower_filename and "dark" not in lower_filename and "180" not in lower_filename:
            return possible_file
    return None
```

Following where these are called from will find other cases of paths being passed around as strings.

PR #2737, opened 2025-07-10
(Eyes tests) Refactor Codebase for Pathlib Compatibility

ashleymeigh2:
This PR continues the work in #2687 to replace legacy os.path usage with pathlib.Path for robust, cross-platform path handling. 

### Description

This PR refactors the Eyes tests and related code to replace all os.path usage with pathlib.Path for clearer, cross-platform path handling. It updates base_eyes.py to search multiple locations for test data—including an environment variable, a sibling repo directory, and the user’s home directory—removing the previous hard-coded reliance on Path.home(). The changes simplify the code, ensure proper typing to avoid errors, and have been verified to pass tests and run correctly on both Windows and Linux.

### Developer Testing

-  Application runs without errors after the refactor

### Acceptance Criteria 

- [x] Application runs without errors after the refactor
- [x] File loading, saving, and path manipulations work correctly using `pathlib`
- [x] All affected code uses `pathlib` instead of `os.path` for path handling
- [x] No regressions in features related to changes
- [x] Run screenshot tests locally: `make test-system`

### DIFF
--- a/mantidimaging/eyes_tests/base_eyes.py
+++ b/mantidimaging/eyes_tests/base_eyes.py
@@ -7,6 +7,7 @@ import unittest
 import getpass
 from pathlib import Path
 from tempfile import mkdtemp
+import logging
 from uuid import uuid4
 
 from PyQt5.QtGui import QFont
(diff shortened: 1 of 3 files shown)


mantidimaging/eyes_tests/base_eyes.py
```diff
@@ -37,25 +37,24 @@
 if TEST_NAME is None:
     TEST_NAME = f"{getpass.getuser()}'s Local Test"
 
-LOAD_SAMPLE = str(Path.home()) + "/mantidimaging-data/ISIS/IMAT/IMAT00010675/Tomo/IMAT_Flower_Tomo_000000.tif"
+LOAD_SAMPLE = Path.home() / "/mantidimaging-data/ISIS/IMAT/IMAT00010675/Tomo/IMAT_Flower_Tomo_000000.tif"
```
MikeSullivan7, line 40: I'm not sure if its worth changing but I used to be able to run `make test-screenshots-win` with the `mantidimaging` and `mantidimaging-dev` in the same directory which isn't the case anymore as `base_eyes.py` now looks in `Path.home()` which for me in Windows is the root of the C drive.

Could we have `base_eyes.py` look in both `Path.home()` and in the parent directory for the test data?

MikeSullivan7 (dismissed):

mantidimaging/eyes_tests/base_eyes.py
```diff
@@ -27,35 +27,43 @@
...
+POTENTIAL_DATA_DIRS.append(Path(__file__).resolve().parent.parent / "mantidimaging-data")
+POTENTIAL_DATA_DIRS.append(Path.home() / "mantidimaging-data")
 
-APPLITOOLS_IMAGE_DIR = os.getenv("APPLITOOLS_IMAGE_DIR")
-if APPLITOOLS_IMAGE_DIR is None:
-    APPLITOOLS_IMAGE_DIR = mkdtemp(prefix="mantid_image_eyes_")
+for candidate in POTENTIAL_DATA_DIRS:
+    if candidate.is_dir():
+        DATA_ROOT = candidate
+        break
```
JackEAllen, line 45: We can be a bit more pythonic here where instead of checking `POTENTIAL_DATA_DIRS` in order, and using the first directory that exists (which is fine), we could use a generator expression with `next()`:
```Python
DATA_ROOT = next(
    (candidate for candidate in POTENTIAL_DATA_DIRS if candidate.is_dir()),
    None
)
if DATA_ROOT is None:
    raise RuntimeError(...)
```

mantidimaging/eyes_tests/base_eyes.py
```diff
@@ -27,35 +27,43 @@
 # APPLITOOLS_BATCH_ID will be set by Github actions to the commit SHA, or a random UUID for individual developer
 # execution
 
-APPLITOOLS_BATCH_ID = os.getenv("APPLITOOLS_BATCH_ID")
-if APPLITOOLS_BATCH_ID is None:
-    APPLITOOLS_BATCH_ID = str(uuid4())
-
+APPLITOOLS_BATCH_ID = os.getenv("APPLITOOLS_BATCH_ID") or str(uuid4())
 API_KEY_PRESENT = os.getenv("APPLITOOLS_API_KEY")
+TEST_NAME = os.getenv("GITHUB_BRANCH_NAME") or f"{getpass.getuser()}'s Local Test"
```
JackEAllen, line 32: As we are using `os.getenv` directly quite a few times here, can we wrap this so we have more control and can add logging while also centralise environment variable access?
```python
def get_env_var(name: str, default=None, required: bool = False) -> str | None:
    value = os.getenv(name, default)
    if value is not None:
        # logging.debug that we found the environment variable {name} and {value}
    else:
        if required:
            # logging.error that we require the variable {name} that is not set
            # raise a runtime error 
        # logging.info that the variable {name} not set and that we are using the default {default}
    return value
```
which could be called like:
```python
API_KEY_PRESENT = get_env_var("APPLITOOLS_API_KEY")
TEST_NAME = get_env_var("GITHUB_BRANCH_NAME", default=f"{getpass.getuser()}'s Local Test")
env_data_path = get_env_var("MANTIDIMAGING_DATA_DIR")
```

mantidimaging/eyes_tests/base_eyes.py
```diff
@@ -27,35 +27,43 @@
...
+                               "git clone https://github.com/mantidproject/mantidimaging-data.git\n")
+NEXUS_SAMPLE = DATA_ROOT / "Diamond/i13/AKingUVA_7050wSSwire_InSitu_95RH_2MMgCl2_p4ul_p4h/24737.nxs"
+
+applitools_image_dir_env = os.getenv("APPLITOOLS_IMAGE_DIR")
+if applitools_image_dir_env is None:
+    APPLITOOLS_IMAGE_DIR = Path(mkdtemp(prefix="mantid_image_eyes_"))
+else:
+    APPLITOOLS_IMAGE_DIR = Path(applitools_image_dir_env)
+    if not APPLITOOLS_IMAGE_DIR.is_dir():
         raise ValueError(f"Directory does not exist: APPLITOOLS_IMAGE_DIR = {APPLITOOLS_IMAGE_DIR}")
```
JackEAllen, line 61: we can simplify this if we use a wrapper around `os.getenv` like `get_env_var()` suggested above as we can pass `mkdtemp(prefix="mantid_image_eyes_"` as a default and just check:
```python
if not APPLITOOLS_IMAGE_DIR.is_dir():
    raise ValueError(f"Directory does not exist: APPLITOOLS_IMAGE_DIR = {APPLITOOLS_IMAGE_DIR}")
```

JackEAllen (changes requested):

hussam-stfc (approved):

JackEAllen (approved):

hussam-stfc (approved):

PR #2761, opened 2025-07-24
Refactor Codebase for Pathlib Compatibility

### LINKED ISSUE #2687: Use pathlib instead of os.path

ashleymeigh2:
## Issue Closes #2687

This PR continues the work in #2687 to replace legacy os.path usage with pathlib.Path for robust, cross-platform path handling. 

### Description

Updated methods to accept and return Path objects where appropriate.
Removed redundant os imports where no longer necessary.

### Developer Testing

-  Application runs without errors after the refactor

### Acceptance Criteria 

- [x] Application runs without errors after the refactor
- [x] File loading, saving, and path manipulations work correctly using `pathlib`
- [x] All affected code uses `pathlib` instead of `os.path` for path handling
- [x] No regressions in features related to changes

### DIFF
--- a/docs/release_notes/next/feature-2687-update-pathlib
+++ b/docs/release_notes/next/feature-2687-update-pathlib
@@ -1,0 +1,1 @@
+2687: Replaced all os.path usage with pathlib.Path for more robust, readable, and modern path handling.
--- a/mantidimaging/gui/test/gui_system_base.py
+++ b/mantidimaging/gui/test/gui_system_base.py
@@ -2,7 +2,6 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 from __future__ import annotations
 
-import os
 from pathlib import Path
 import unittest
 from unittest import mock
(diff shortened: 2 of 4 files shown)


JackEAllen (commented):
Thank you for the changes. I've left a few suggestions I don't feel too strongly about regarding possible improvement opportunities to `add_data_files`. Feel free to disagree as the suggestions are technically minor scope creep.
The second suggestion is to keep `add_conda_dynamic_libs` closer to the original implementation which I feel is clearer to understand compared to list comprehension in this case.

PR #2767, opened 2025-07-25
Fix welcome screen layout warnings

### LINKED ISSUE #2754: System Tests - Main Banner Layout: Banner Layout Error
### Summary

When running GUI system tests, the welcome screen banner `Layout` can cause error tracebacks due to the layout already existing (see failure logs). This is likely caused by opening many instances of the main window, not necessarily before closing others between tests. 

The banner (mantidimaging/gui/windows/welcome_screen/view.py) already checks is a layout exists befre trying to create a new layout:
```python
        if not self.Banner_container.layout():
            self.Banner_container.setLayout(QVBoxLayout())

        # Set the banner image
        banner = (finder.ROOT_PATH / "gui" / "ui" / "images" / "welcome_banner.png").as_posix()
        self.banner_label = QLabel(self)
        self.banner_label.setPixmap(QPixmap(banner))
        self.banner_label.setScaledContents(False)
        self.banner_label.setMinimumSize(self.Banner_container.size())
        self.Banner_container.layout().addWidget(self.banner_label)
```

Explore modifying tests startup and teardown to handle widget cleanup to resolve multiple instances of windows potentially existing at the same time causing this issue.

### Steps To Reproduce

Check existing GUI system tests: [PR system tests exhibiting issue](https://github.com/mantidproject/mantidimaging/actions/runs/16418181435/job/46389286995#step:17:702)
Run system tests locally with: `make test-system`

### Current Behaviour

Somewhere we are trying to assign a new layout to a QWidget that already has one which is not allowed in Qt - this is likely caused by the system tests not correctly handling teardown of widgets

### Expected Behaviour

Correctly handle widget cleanup in system tests to prevent multiple layout issues occurring which do not correctly reflect the state of the application in realtime use 

### Failure Logs

```python
QWidget::setLayout: Attempting to set QLayout "" on QWidget "Banner_container", which already has a layout
QWidget::setMinimumSize: (/QDockWidget) Negative sizes (0,-1) are not possible
QWidget::setMaximumSize: (/QDockWidget) Negative sizes (0,-1) are not possible
```

The negative (0, 01) part of the traceback probably means something like a QDockWidget is being assigned a minimum or maximum size with a negative value which is not possible with Qt, this could be related to a size calculation being slightly incorrect or is a widget is being show/hidden in an unexpected state.

samtygier-stfc:
### Description
Fixes warnings related to the welcome window

- don't setLayout on banner container. This is always created in the ui file so is never needed.
- set layout on widget passed to setTitleBarWidget(). We pass an empty QWidget to hide the title, but it needs to have a valid sizeHint which requires it to have a layout

Also extract method to hide the title for a DockWidget to make it clear what is happening

### Developer Testing 

- Run the system test and check that the warnings from the issue are not visible
- Open Mantid Imaging and check that the welcome screen is shown

- [x] Run the system test and check that the warnings from the issue are not visible
- [x] Open Mantid Imaging and check that the welcome screen is shown

Release notes not needed, just a warning seen in tests

### DIFF
--- a/mantidimaging/gui/windows/main/view.py
+++ b/mantidimaging/gui/windows/main/view.py
@@ -13,7 +13,7 @@ import numpy as np
 from PyQt5.QtCore import Qt, pyqtSignal, QUrl, QPoint
 from PyQt5.QtGui import QIcon, QDragEnterEvent, QDropEvent, QDesktopServices
 from PyQt5.QtWidgets import QAction, QDialog, QLabel, QMessageBox, QMenu, QFileDialog, QSplitter, \
-    QTreeWidgetItem, QTreeWidget, QDockWidget, QWidget
+    QTreeWidgetItem, QTreeWidget, QDockWidget, QWidget, QVBoxLayout
 
 from mantidimaging.core.data import ImageStack
 from mantidimaging.core.io.utility import find_first_file_that_is_possibly_a_sample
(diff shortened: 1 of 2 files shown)


hussam-stfc (approved):
All tests pass and welcome screen appears as expected with no warnings.

PR #2774, opened 2025-07-28
Update tomopy from 1.12.2 to 1.15.2

### LINKED ISSUE #2667: Package Management: Update TomoPy
### Action

Update `tomopy` from ` 1.12.2 ` to `1.15.2 `

### Additional Context

* Explore if this will cause any compatibility issues with any of our other packages
* Manually test areas of Mantid Imaging that make use of TomopP

ashleymeigh2:
### Description

Updated `tomopy` from `1.12.2` to `1.15.2`.  
No compatibility issues were identified with Mantid Imaging after this upgrade.  
Manual and automated tests confirmed correct behavior in core features that rely on TomoPy.

### Developer Testing 

- Manually tested operations that use TomoPy

### Testing
- [ ] Run reconstruction
- [ ] Ring removall
- [ ] rotation center 

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -25,7 +25,7 @@ requirements:
     - numpy=1.26.*
     - numexpr=2.8.*
     - algotom=1.6.*
-    - tomopy=1.12.*
+    - tomopy=1.15.*
     - cudatoolkit=11.8.*
     - cupy=12.3.*
     - astra-toolbox=2.1.0=cuda*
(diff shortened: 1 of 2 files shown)


ashleymeigh2:
We're hitting a version conflict with tomopy=1.15.*, which requires scipy<1.14.

However, we currently need scipy>=1.14 to stay compatible with newer versions of other packages like sphinx and cupy.

This means we can’t install the latest tomopy without downgrading scipy, which would then break other dependencies.
Alternatively, we could downgrade tomopy to 1.14.0, but that means missing out on its latest fixes and features.

Would love thoughts on which direction makes the most sense.

samtygier-stfc:
Lets try with tomopy 1.15.3 which should fix the issue

PR #2778, opened 2025-07-30
System Tests: Limit number of allowed alive objects during tearDown

### LINKED ISSUE #2787: System Tests: Impose leaking object limit in Leak Tracker
### Current Behaviour

The `leak_tracker` in the system tests keeps track of objects that are still alive during the `tearDown` of each test. For some system tests, a number of alive objects is acceptable but not for others. 

### Desired Behaviour

Impose a limit on how many alive objects are allowed for certain tests, where are Traceback is given and an error is raised if too many objects are kept alive.

Make sure that the default number of allowed alive objects is =0 for any newly written tests.

### Importance of Feature

This is allow us to catch instances where a large number of alive objects exist during the `tearDown` of a test, and make it easier to debug

MikeSullivan7:
### Description

A `leak_count_limit` of zero has been imposed at the `GuiSystembase` level, but this can be superseded by the system tests at a Class level or a method level. This allows specific tests to have an amount of leaked objects without failing (i.e. if we would expect the test to have leaked objects), but prevents new tests having uncounted leaked objects and tells us if tests are unexpectedly leaking objects.

### Developer Testing 

- `make test-system`

- [ ] run `make test-system` and check that no leaked objects warnings/errors appear
- [ ] locate a test which leaks objects, i.e. `mantidimaging/gui/test/gui_system_operations_test.py::TestGuiSystemOperations` and alter the `self.leak_count_limit` to a lower number. Running the tests again should raise a `Leak_Tracker` error

### DIFF
--- a/mantidimaging/gui/test/gui_system_base.py
+++ b/mantidimaging/gui/test/gui_system_base.py
@@ -33,13 +33,12 @@ SHORT_DELAY = 100
 @start_qapplication
 class GuiSystemBase(unittest.TestCase):
     app: QApplication
-    leak_count_limit: int
+    leak_count_limit: int = 0
 
     def setUp(self) -> None:
         self.main_window = MainWindowView()
         self.main_window.show()
         QTest.qWait(SHORT_DELAY)
-        self.leak_count_limit = 0
 
     def tearDown(self) -> None:
         """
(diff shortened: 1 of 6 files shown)


samtygier-stfc (approved):
Looks good. Tested locally.

PR #2793, opened 2025-08-06
Spectrum Viewer: Prevent QThread from accessing Presenter data during runtime

### LINKED ISSUE #2715: Spectrum Viewer: Refactor to prevent Qthread from accessing mutable data
### Current Behaviour

The `TaskWorkerThread.task_function` in the Spectrum viewer uses objects during calculation which could possibly be modified during a calculation, leading to bugs or miscalculations which are hard to preproduce.

### Desired Behaviour

 This should be refactored such that the needed objects are passed to the worker thread such that they are not modified, or Thread Locking should be used to protect the necessary data.

MikeSullivan7:
### Description

The PR is to prevent the QThread in the Spectrum Viewer from changing data which may be accessed by different processes, while the QThread is running as this may lead to unreliable results due to a race condition. To achieve this, the function in the Worker Thread no longer directly accesses the attributes in the presenter, but instead takes them as arguments and outputs the altered attributes as `thread.result()`. The attributes in the presenter space are then changed during the `thread_cleanup` instead of inside the thread itself.

Also to make sure that nothing could be changed in the running of the thread, a `Lock()` is used inside the thread function.

### Developer Testing 

- [x] Load data in the spectrum viewer and open multiple ROIs
- [x] Move the position and sizes of the ROIs in any combination of ways you can think of and as fast as you feasibly can.
- [x] ensure that the spectrums update as expected and that no errors are raised.

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import csv
-import threading
 from enum import Enum
 from functools import partial
 from typing import TYPE_CHECKING
@@ -71,8 +70,6 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         self.handle_roi_change_timer = QTimer()
         self.handle_roi_change_timer.setSingleShot(True)
         self.handle_roi_change_timer.timeout.connect(self.handle_roi_moved)
-
-        self.spectrum_calculation_lock = threading.Lock()
 
     def handle_stack_modified(self) -> None:
         """
--- a/mantidimaging/gui/windows/spectrum_viewer/view.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/view.py
@@ -334,9 +334,14 @@ class SpectrumViewerWindowView(BaseMainWindowView):
         if self.normalise_error_issue and self.normalisation_enabled():
             self.normaliseErrorIcon.setPixmap(self.normalise_error_icon_pixmap)
             self.normaliseErrorIcon.setToolTip(self.normalise_error_issue)
+            for roi_name in self.table_view.get_roi_names():
+                self.set_roi_visibility_flags(roi_name, visible=False)
         else:
             self.normaliseErrorIcon.setPixmap(QPixmap())
             self.normaliseErrorIcon.setToolTip("")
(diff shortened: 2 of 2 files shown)


mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -244,28 +247,29 @@ def handle_notify_roi_moved(self, roi: SpectrumROI) -> None:
...
     def run_spectrum_calculation(self):
-        roi_name = list(self.roi_to_process_queue.keys())[0]
-        roi = self.roi_to_process_queue[roi_name]
-        chunk_size = 100
-        if chunk_size > 0:
-            nanInds = np.argwhere(np.isnan(self.image_nan_mask_dict[roi_name]))
-            chunk_start = int(nanInds[0, 0])
-            if len(nanInds) > chunk_size:
-                chunk_end = int(nanInds[chunk_size, 0])
+        with self.spectrum_calculation_lock:
```
samtygier-stfc, line 250: I am not sure if this solves the problem. We still access lots of presenters variables from inside the worker thread, which might simultaneously be access/modified in the main thread. I think this approach would also need the lock to be acquired everywhere that things like `self.roi_to_process` are accessed.

Would it work to put input data in the `TaskWorkerThread` kwargs and then return the spectrum chunk and put in to `spectrum_data_dict` from in `thread_cleanup`?

Or would it be possible to put the tasks that need doing in a [queue](https://docs.python.org/3/library/queue.html)?

samtygier-stfc (changes requested):

ashleymeigh2 (dismissed):
Looks good

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -71,6 +72,8 @@ def __init__(self, view: SpectrumViewerWindowView, main_window: MainWindowView):
         self.handle_roi_change_timer.setSingleShot(True)
         self.handle_roi_change_timer.timeout.connect(self.handle_roi_moved)
 
+        self.spectrum_calculation_lock = threading.Lock()
```
samtygier-stfc, line 75: I think this is not used any more, can be removed.

samtygier-stfc (changes requested):

PR #2804, opened 2025-08-12
use QT_QPA_PLATFORM=minimal and message handler to surpress QFont war…

MikeSullivan7:
…nings

### Developer Testing 

- ...

- [ ] ...

### DIFF
--- a/mantidimaging/test_helpers/qt_test_message_handler.py
+++ b/mantidimaging/test_helpers/qt_test_message_handler.py
@@ -1,0 +1,20 @@
+# Copyright (C) 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+from PyQt5 import QtCore
+
+
+def qt_message_handler(mode, context, message):
+    if mode == QtCore.QtInfoMsg:
+        mode = 'INFO'
+    elif mode == QtCore.QtWarningMsg:
+        mode = 'WARNING'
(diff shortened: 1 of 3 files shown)


mantidimaging/test_helpers/qt_test_message_handler.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+from PyQt5 import QtCore
+
+
+def qt_message_handler(mode, context, message):
```
RasmiaKulan, line 6: I wanted to suggest replacing the `elif` chain with a mapping dictionary and keep the logic for the special `propogateSizeHints` case in on place.


```suggestion
def qt_message_handler(mode, context, message):
      mode_map = {
              QtCore.QtInfoMsg: 'INFO',
              QtCore.QtWarningMsg: 'WARNING',
              QtCore.QtCriticalMsg: 'CRITICAL',
              QtCore.QtFatalMsg: 'FATAL'
          }
    
      # Get the string label for mode, otherwise default to 'DEBUG'
      mode_str = mode_map.get(mode, 'DEBUG')

      if mode == QtCore.QtWarningMsg and 'propagateSizeHints' in message:
             return
      print(f"qt_message_handler: line: {context.line}, func: {context.function}, file: {context.file}")
      print(f"{mode_str}: {message}\n")
```

Issue #2769, opened 2025-07-28
GitHub Action to Convert PR to Draft on PR Change Request


JackEAllen:
GitHub action to convert a PR to a draft PR if change request has been made by a reviewer.
If associated Issue or PR is no issue is in Mantid Imaging Work project board in the "Review" column, move to "In Progress" column

JackEAllen:
We can maybe stop after converting to a draft PR if updating the columns proves to be complicated. Testing may not be straight forward.

PR #2818, opened 2025-08-19
FLASH PR: Update Token to Use Custom Token

JackEAllen:
Resolve Repo permissions by creating a new fine-grained token called `ACTION_CONVERT_TO_DRAFT` to allow for #2769 to work as expected which is currently failing with: `Error: A GitHub token is required with the pull_requests:write permission`


samtygier-stfc (approved):
Looks good

PR #2829, opened 2025-08-27
Pyproject.toml: Pyright Optional Call Flag removed

MikeSullivan7:
### Description

This PR removes the Pyright optional call flag from pyproject.toml. The flag is no longer needed with the current version of Pyright, and removing it simplifies configuration while keeping type checking behavior consistent.

### Developer Testing 

- [ ] run `pyright mantidimaging` and check that no errors or warning occur


samtygier-stfc:
Looks like this only failed in the queue due to network issues, so you can try to merge it again.

Wondering though if `assert setup is not None` ... would be more appropriate here, as these only get used in the case where the the function is not None.

PR #2830, opened 2025-08-27
Operations Bug: Display of 'difference' overlayed on image

### LINKED ISSUE #2803: Operations: Display of 'difference' overlayed on image
### Current Behaviour

I am not sure if I have understood the behaviour of the difference overlay.
An example on the 'ring removal' operation:

Image

The difference overlay is only shown for some, not all of the places in the image where there is a difference in values - e.g. in the 3rd image we see a lot more differences highlighted.

Is this a bug or intentional?

Note: The amount of green does not change when I zoom in on the different areas

### Desired Behaviour
All of the differences highlighted in the overlay

ashleymeigh2:
### Description

Fixes a bug in the Overlay Difference view for filters like Ring Removal.  
Previously, the overlay only showed differences above a hardcoded threshold (`1e-3`), which caused many real changes to be hidden. So updated hardcoded threshold (`1e-6`), so changes are now accurately reflected in the overlay.

### Developer Testing 

- [x] Manually verified overlay now matches the full difference image visually
- [x] Overlay correctly highlights faint ring artifacts that were previously missed
- [ ] I have verified unit tests pass locally: `python -m pytest -vs`

- [ ] Load any reconstructed volume and apply the Ring Removal filter
- [ ] Enable "Overlay difference" checkbox
- [ ] Confirm that the overlay now reflects all differences (even subtle ones)
- [ ] Visually verify that overlay matches the “Image difference” panel more accurately than before

- [x] Release Notes have been updated (if applicable)


ashleymeigh2:
Closed this pull request because it got out of sync with the origin branch. Opened a new PR with the same fix so changes are easier to review and merge.

PR #2835, opened 2025-08-28
2834 Geometry: Add new type property

### LINKED ISSUE #2834: Geometry: Add new 'type' property
### Current Behaviour

The current Geometry type needs to be obtained from the CIL `AcquistionGeometry` via `self.config.system.geometry` + `self.config.system.dimension`. These properties have also been changed before in CIL's code.

### Desired Behaviour

The Geometry class should have a `type` property which returns an enum with a display-friendly string. This saves us having to perform any complex retrievals and can be limited to Parallel3D and Conebeam3D (only types currently supported by Mantid Imaging). 

### Additional Context

The only place currently using Geometry type in Mantid Imaging is the Stack Properties dialogue window, which will need to be modified to use this new property.

The [new Geometry window](https://github.com/mantidproject/mantidimaging/pull/2817) will make use of this property.

hussam-stfc:
### Description

- Added a new `type` property to the `Geometry` class which return the current Geometry type, wrapping CIL's internal `self.config.system.geometry` and `self.config.system.dimension` in a simpler format.
- The new property returns an enum suitable for logic comparisons, with a display-friendly `str` value.
- A new test has been added to verify this property's functionality.
- The Stack Properties dialogue window has been updated to use this new property.

### Developer Testing 

- I have verified that the Stack Properties window displays the string "Parallel 3D" under "type".

- [x] Verify the Stack Properties dialogue displays the correct value

The property's docstring could be improved, any feedback is welcome.

### DIFF
--- a/docs/release_notes/next/dev-2834-add-geometry-type-property
+++ b/docs/release_notes/next/dev-2834-add-geometry-type-property
@@ -1,0 +1,1 @@
+#2834: Added a new `type` property to the Geometry class


samtygier-stfc (approved):
Looks good. I've confirmed that it displays in the stack properties dialog.

PR #2837, opened 2025-08-28
Add function to auto find a bounding box for given ImageStack data

### LINKED ISSUE #2784: Core: method to find the sample bounds
To enable auto cropping in automated workflows and to give good initial coordinates for Crop and ROI normalisation operations, it would be useful to have an automated way to identify a bounding box for the sample within the image.

### Desired Behaviour

This should be implemented in a new file in core/fitting. It should provide a way to pass in an imagestack and get an ROI back.

Initial version should be quick and simple, using numpy/scipy. It may work to use `mean` to flatten to the x dimension and find the edge when the sample starts, then do the same for y.

It is better to have the estimated bound be to big than to over crop.

Good to avoid adding new dependencies.

### Importance of Feature

Needed for automation

MikeSullivan7:
…n tolerance

### Developer Testing 

- ...

- [ ] ...

### DIFF
--- a/mantidimaging/core/fitting/bounding_box.py
+++ b/mantidimaging/core/fitting/bounding_box.py
@@ -1,0 +1,41 @@
+# Copyright (C) 2021 ISIS Rutherford Appleton Laboratory UKRI
+# SPDX - License - Identifier: GPL-3.0-or-later
+import matplotlib.pyplot as plt
+import numpy as np
+
+from mantidimaging.core.data import ImageStack
+from mantidimaging.core.utility.sensible_roi import SensibleROI
+
+
+def get_bounding_box(stack: ImageStack) -> SensibleROI:
+    y_tol = 0.02
+    x_tol = 0.02
+    data = stack.data[:, 5:-5, 5:-5]  # clip dark edges
+    all_col_mean = np.mean(data, axis=(1, 0))
+    all_row_mean = np.mean(data, axis=(2, 0))
+    all_col_mean_flipped = np.flip(all_col_mean)
+    all_row_mean_flipped = np.flip(all_row_mean)
+
+    plt.plot(np.diff(all_row_mean) / all_row_mean[1:])
+    plt.show()
+
+    def find_bound(data_temp, tol, dim=0):
+        data_temp = data_temp[0:round(len(data_temp) / 2)]
+        while True:
+            if not np.where((np.diff(data_temp) / data_temp[1:]) < -tol)[0].size == 0:
+                if dim == 0:
+                    return np.where((np.diff(data_temp) / data_temp[1:]) < -tol)[0][0]
+                else:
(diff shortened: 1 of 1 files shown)


mantidimaging/core/fitting/bounding_box.py
```diff
@@ -0,0 +1,41 @@
...
+    all_row_mean = np.mean(data, axis=(2, 0))
+    all_col_mean_flipped = np.flip(all_col_mean)
+    all_row_mean_flipped = np.flip(all_row_mean)
+
+    plt.plot(np.diff(all_row_mean) / all_row_mean[1:])
+    plt.show()
+
+    def find_bound(data_temp, tol, dim=0):
+        data_temp = data_temp[0:round(len(data_temp) / 2)]
+        while True:
```
samtygier-stfc, line 24: Can this be done without a potentially infinite loop?

Could `tol` be set based on the range of `np.diff(data_temp)`?

mantidimaging/core/fitting/bounding_box.py
```diff
@@ -0,0 +1,41 @@
...
+            if not np.where((np.diff(data_temp) / data_temp[1:]) < -tol)[0].size == 0:
+                if dim == 0:
+                    return np.where((np.diff(data_temp) / data_temp[1:]) < -tol)[0][0]
+                else:
+                    return data.shape[dim] - np.where((np.diff(data_temp) / data_temp[1:]) < -tol)[0][0]
```
samtygier-stfc, line 29: Is `data.shape[dim]` the same as `data_temp.size`? Can you avoid needing to pass in `dim`.

mantidimaging/core/fitting/bounding_box.py
```diff
@@ -0,0 +1,41 @@
...
+            else:
+                tol -= 0.001
+
+    # add a buffer around the found bounds to make sure all the sample is captured in the bounding box
+    buffer_pixels = round(data.shape[1] / 100)
+
+    top_bound = find_bound(all_row_mean, x_tol, 0) - buffer_pixels
+    bottom_bound = find_bound(all_row_mean_flipped, x_tol, 2) + buffer_pixels
+    left_bound = find_bound(all_col_mean, y_tol, 0) - buffer_pixels
+    right_bound = find_bound(all_col_mean_flipped, y_tol, 1) + buffer_pixels
```
samtygier-stfc, line 39: Do the 5 pixels from the edge clipping need to be added/subtracted to these?

samtygier-stfc (changes requested):

MikeSullivan7:
Increasing the length of data which the noise characteristics are calculated from seem to yield better results:
```
noise_sigma = data_diff[:20].std()
noise_mean = data_diff[:20].mean()
```

image

image

PR #2856, opened 2025-09-18
2485 recon algos get imagestack geometry

### LINKED ISSUE #2485: Recon algorithms get geometry from imagestack
### Current Behaviour

Recon algorithms are passed the COR and tilt from the recon window

### Desired Behaviour

Use the AcquisitionGeometry from the image stack
Do required conversion for calling none CIL recon libraries
Check that the AcquisitionGeometry is compatible with a given recon algorithm. (e.g. all current algorithms require parallel beam, and only handle tilt in one axis)

hussam-stfc:
### Developer Testing 

- ...

- [ ] ...


samtygier-stfc (approved):
Looking good now. I have tested locally.

PR #2884, opened 2025-10-10
2614 geometry refactor projection angles handling

### LINKED ISSUE #2614: Geometry: refactor handling of stack projection angles
### Current Behaviour

There are a few issues:

- Angle is held in the 3 places in the imagestack (`self._projection_angles`, `self._log_file.projection_angles()` and `self.geometry`)
- `ImageStack.projection_angles()` has an awkward way to make up angles if they are missing, this is done on the fly in the reconstruction process if needed
- The reconstruction GUI has a `Maximum projection angle` input, that is only used in the above case

### Desired Behaviour

- Make the geometry the main store of the angles
- If there is no angle information from logs, make a way to add a user defined range
- Remove the `Maximum projection angle` input. Put a warning if the stack has no angle data, the user can now add them using above method.

Note: The solution might need a bit more though or discussion.

### Importance of Feature

Simplify the code.
Remove a confusing GUI element

hussam-stfc:
### Description

- Refactored `ProjectionAngles` logic in `Geometry` and `ImageStack`. `Geometry` is now the sole source of truth for angle data.
- Added guards against `None` for `ProjectionAngles` at point of use
- Updated some reconstruction tests to use more accurate 3D-array test data.
- Improved method docstrings in reconstruction window and I/O testing code.

### Developer Testing 

- I have manually verified that reconstruction continues to work as expected

 - [ ] Unit tests pass locally: python -m pytest -vs
 - [ ]  Verify that reconstruction performs as expected (recon results match previous commits)

### DIFF
--- a/mantidimaging/gui/test/gui_system_loading_test.py
+++ b/mantidimaging/gui/test/gui_system_loading_test.py
@@ -220,7 +220,7 @@ class TestGuiSystemLoading(GuiSystemBase):
         image_count, *image_shape = sample.shape
         self.assertEqual(image_shape, [128, 128])
         self.assertEqual(image_count, expected_count)
-        self.assertEqual(len(sample.real_projection_angles().value), expected_count)
+        self.assertEqual(len(sample.projection_angles().value), expected_count)
 
     @mock.patch("mantidimaging.gui.windows.main.MainWindowView._get_file_name")
     def test_replace_image_stack(self, mocked_select_file):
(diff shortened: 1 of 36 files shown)


mantidimaging/gui/test/gui_system_loading_test.py
```diff
@@ -75,29 +75,43 @@ def _get_log_angle(self, log_path):
...
+        self.assertNotIn("log_file", sample.metadata)
+        self.assertIsNone(sample.projection_angles())
 
         # Load sample log
         QTimer.singleShot(SHORT_DELAY, lambda: self._click_stack_selector())
         QTimer.singleShot(SHORT_DELAY * 2, lambda: self._click_messageBox("OK"))
         self.main_window.actionSampleLoadLog.trigger()
-
         self.assertIn("log_file", sample.metadata)
         self.assertEqual(sample.metadata['log_file'], str(log_path))
```
samtygier-stfc: Could also check the angles here, as they will now have been loaded from the log.

samtygier-stfc (changes requested):
Looking good. A few changes, but they are mostly small things.

samtygier-stfc (approved):
I have fixed the last few comments and small issue for stacks without angles.

Tested with several datasets.

PR #2889, opened 2025-10-17
Spectrum Viewer: the current ROIs persist when switching stacks

### LINKED ISSUE #2828: Spectrum Viewer: Don't clear ROIs when changing stacks
### Current Behaviour

When the stack is switched in the spectrum viewer the ROIs are reset.

### Desired Behaviour

If a user has set of similar datasets they often want to be using the same ROIs for each of them. A simple solution would be to not reset the ROIs when changing stacks.

If the new stack is a smaller size than the previous one, then the old ROIs may not be within bounds. In this case they should be reset.

Need to make sure that the spectrums are correctly updated when switching stacks

### Importance of Feature

Saves time for workflows with multiple datasets

MikeSullivan7:
### Description

The logic in the Spectrum Viewer has been changed such that the current ROIs are no longer deleted when the Sample Stack is changed. Instead the relevant spectrums are cleared and are recalculated upon sample change in the Spectrum model. Due to the order of the logic changing, some changes were made to the Spectrum system-tests and eyes-tests to ensure that threads from previous tests have finished running before starting another test.

### Developer Testing 

- ...

- [ ] Load data and open the Spectrum Viewer
- [ ] Add a few rois and move them around so they are distinct
- [ ] Change to a different stack (that has multiple images) and check that the rois stay where they were and the spectrum are recalculated.
- [ ] Delete all rois and then switch stacks, the Spectrum Viewer should add a new roi after the new stack is loaded.

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -118,6 +118,13 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         Called when the stack has been changed in the stack selector.
         """
         self.view.roi_form.exportTabs.setDisabled(uuid is None)
+        if self.current_stack_uuid and uuid:
+            if ((current_stack := self.main_window.get_stack(self.current_stack_uuid))
+                    and (new_stack := self.main_window.get_stack(uuid))):
+                if current_stack.shape[1:] != new_stack.shape[1:]:
+                    self.do_remove_roi()
+                    self.view.table_view.clear_table()
+                    LOG.info("The new image stack has a different size to previous image stack: deleting all ROIs")
 
         if uuid == self.current_stack_uuid:
             return


mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -156,8 +157,9 @@ def handle_sample_change(self, uuid: UUID | None) -> None:
             except RuntimeError:
                 norm_stack = None
             self.model.set_normalise_stack(norm_stack)
-        self.do_add_roi()
-        self.add_rits_roi()
+        if not self.view.spectrum_widget.roi_dict:
+            self.do_add_roi()
+            self.add_rits_roi()
```
samtygier-stfc: Is this still needed if the ROIs are made on lines 147, 148?

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -297,7 +294,8 @@ def handle_roi_moved(self) -> None:
             "spec_mode": self.spectrum_mode,
             "shutter_norm": self.view.shuttercount_norm_enabled(),
             "chunk_start": chunk_start,
-            "chunk_end": chunk_end
+            "chunk_end": chunk_end,
+            "spectrum_data_dict": self.view.spectrum_widget.spectrum_data_dict
```
samtygier-stfc: This goes a bit against my intuition. `thread_finished_processing` should be running back in the main thread (because its triggered via a signal). And it needs to be updating the real cache. So using `self.view` there should be fine.

samtygier-stfc (approved):
Thanks. Working well now,

Issue #2685, opened 2025-06-27
Clean up how 180 files are handled


JackEAllen:
### Current Behaviour

The current behaviour comes from a time when MI was only used for tomography. It has a number of issues.

- MI will automatically create a "180 Projection" for any dataset that does not have one, which makes no sense for non-tomography data
- Even for tomography the auto 180 proj is a bit clunky, it adds a redundant image stack
- When finding the CoR and Tilt using correlation, only option is to use the first image in the sample and the 180 stack. It should also work to use any 2 projections separated by 180 degrees.

### Desired Behaviour
Sub Issues:
- [ ] in recon correlate button for finding cor and tilt, provide options for which angles are used (should be 180 apart i.e. 0-180 or 90-270 if 360 dataset - don't require 180
- [ ] Don't generate 180 automatically
- [ ] explore cleanup opportunities in operation window (if any)

JackEAllen:
Lets mockup some ideas and designs for if this is places in the operations windows, a new window entirely or leave in the recon window

PR #2908, opened 2025-10-23
feat(recon): add support for selecting projection pairs in correlation

ashleymeigh2:
## Issue
subissue of #2685

### Description
Implements the first sub-issue of #2685 by allowing correlation between any
two projections separated by ~180°, without automatically generating or
requiring a 180° projection stack.

- Updated `_auto_find_correlation` to use arbitrary projection indices
- Modified `find_center()` to accept configurable projection pairs
- Removed dependency on `has_proj180deg()` and `.proj180deg`
- Adjusted related unit tests to reflect new behavior

### Developer Testing
- Verified all unit tests pass locally: `python -m pytest -vs`

- [ ] Confirm correlation works for arbitrary projection pairs
- [ ] Verify no “180 Projection” stack is auto-generated
- [ ] Unit tests pass locally

- [ ] Release notes updated to reflect new correlation options

### DIFF
--- a/mantidimaging/core/rotation/polyfit_correlation.py
+++ b/mantidimaging/core/rotation/polyfit_correlation.py
@@ -24,9 +24,26 @@ def do_calculate_correlation_err(store, search_index: int, p0_and_180, image_wid
     store[:] = np.square(np.roll(p0_and_180[0], search_index, axis=1) - p0_and_180[1]).sum(axis=1) / image_width
 
 
-def find_center(images: ImageStack, progress: Progress) -> tuple[ScalarCoR, Degrees]:
-    if images is None or images.proj180deg is None:
-        raise ValueError("images and images.proj180deg cannot be None")
+def find_center(images: ImageStack,
+                progress: Progress,
+                use_projections: tuple[int, int] | None = None) -> tuple[ScalarCoR, Degrees]:
+
+    if images is None:
+        raise ValueError("images cannot be None")
+
+    proj180 = getattr(images, "proj180deg", None)
+    if use_projections is not None:
+        start_idx, end_idx = use_projections
+        proj_a = images.projection(start_idx)
+        proj_b = np.fliplr(images.projection(end_idx))
+        LOG.info(f"Using projections {start_idx} and {end_idx} for correlation")
+    elif proj180 is not None:
+        proj_a = images.projection(0)
+        proj_b = np.fliplr(proj180.data[0])
+        LOG.info("Using flipped images.proj180deg for correlation (legacy behaviour)")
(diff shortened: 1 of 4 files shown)


samtygier-stfc (changes requested):

mantidimaging/core/rotation/polyfit_correlation.py
```diff
@@ -24,9 +24,29 @@ def do_calculate_correlation_err(store, search_index: int, p0_and_180, image_wid
...
+    elif proj180 is not None:
+        proj_a = images.projection(0)
+        proj_b = np.fliplr(proj180.data[0])
+        LOG.info("Using flipped images.proj180deg for correlation (legacy behaviour)")
+    else:
+        start_idx = 0
+        end_idx = (images.num_projections // 2) % images.num_projections
+        proj_a = images.projection(start_idx)
+        proj_b = np.fliplr(images.projection(end_idx))
+        LOG.info(f"No proj180deg found; using projections {start_idx} and {end_idx}")
```
samtygier-stfc: Let raise an exemption in this third case. The caller has not given a proj180 or `use_projections`.

mantidimaging/core/rotation/test/polyfit_correlation_test.py
```diff
@@ -52,8 +52,23 @@ def test_find_center_offset(self):
         self.crop_images(images.proj180deg, (2, 10, 0, 10))
         mock_progress = mock.create_autospec(Progress, instance=True)
         res_cor, res_tilt = find_center(images, mock_progress)
-        assert res_cor.value == 4.0, f"Found {res_cor.value}"
-        assert abs(res_tilt.value) < 1e-6, f"Found {res_tilt.value}"
+        assert abs(res_cor.value - 4.0) <= 1.1, f"Found {res_cor.value}"
```
samtygier-stfc: I think we can put the original test values back in now
```
        assert res_cor.value == 4.0, f"Found {res_cor.value}"
        assert abs(res_tilt.value) < 1e-6, f"Found {res_tilt.value}"
```

samtygier-stfc (changes requested):

samtygier-stfc (changes requested):
Also, the changes in gui should be moved to separate PR with some user control over whether to use the proj 180, or choose which angles to use

samtygier-stfc (changes requested):
There is still the change in `gui_system_reconstruction_test.py` that I think can be removed.

Also might be worth squashing as there are some back and forth changes.

samtygier-stfc (approved):
Looking good now

PR #2921, opened 2025-11-11
AsyncTaskDialogPresenterTest: fix happy case, add error case

### LINKED ISSUE #2919: Async task: Error in test
### Summary

`AsyncTaskDialogPresenterTest.test_basic_happy_case` has a bug which sometimes shows an error in the test, even though the test passes. It is possible that it is contributing to unreliability of other tests.

7a5f8bdbd5520aa668d76e450776f0e801ba4c92 removed the unused `*args` from `call_with_known_parameters`. The change in `AsyncTaskDialogPresenterTest` left a bug because there is no value passed for `a`. This does not cause the test to fail because it never check the return value or for an error.

In some cases the backtrace can show:
```
ERROR    mantidimaging.gui.dialogs.async_task.task:task.py:58 Failed to execute task: AsyncTaskDialogPresenterTest.test_basic_happy_case.<locals>.f() missing 1 required positional argument: 'a'
  File "/mnt/ceph/home/t1110869/pixi_test/mantidimaging/mantidimaging/gui/dialogs/async_task/task.py", line 49, in run
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/ceph/home/t1110869/pixi_test/mantidimaging/mantidimaging/core/utility/func_call.py", line 12, in call_with_known_parameters
    return func(**ka)
           ^^^^^^^^^^
TypeError: AsyncTaskDialogPresenterTest.test_basic_happy_case.<locals>.f() missing 1 required positional argument: 'a'
PASSED
```

### Steps To Reproduce

```
pytest -vs -o log_cli=true mantidimaging/gui/dialogs/async_task/test/presenter_test.py
```
but the trace does not always show.

You can some output to the end of the test
```python
        print(p.model.task.error)
        print(p.model.task.result)
```

### Current Behaviour

Error happens but is not always shown

### Expected Behaviour

Test is for happy case, so should call the function with correct args

### Context

Current main.

Spotted on pixi PR, which is not the cause maybe makes it more visible?

Self point as 1

samtygier-stfc:
And add checks for result and error to each case

### Description

Happy case:
Fix arguments passed to worker function so that there is not an exception
Check the result and that there is no error

Error case:
Check that an error is passed back in `task.error`

### Developer Testing 

- [ ] Check that tests run on github actions

Not needed

### DIFF
--- a/mantidimaging/gui/dialogs/async_task/test/presenter_test.py
+++ b/mantidimaging/gui/dialogs/async_task/test/presenter_test.py
@@ -23,7 +23,7 @@ class AsyncTaskDialogPresenterTest(unittest.TestCase):
 
         p = AsyncTaskDialogPresenter(v)
         p.set_task(f)
-        p.set_parameters(b=4)
+        p.set_parameters(a=3, b=4)
         self.assertFalse(p.task_is_running)
 
         p.notify(Notification.START)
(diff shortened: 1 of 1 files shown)


MikeSullivan7 (approved):
All tests run fine, tests behave as expected when made to fail.

PR #2940, opened 2025-11-21
Testing: leak info is only shown if test doesnt pass

### LINKED ISSUE #2918: Leak tracker: misleading leak info when a test fails
### Summary

In the system tests we use the leak tracker to check for references to ImageStacks no being deleted. This is useful for detecting cases where there are left over references to objects that could waste large amounts of memory.

We check for leaked objects in `GuiSystemBase.tearDown`.

In the case where a test has failed, e.g. by hitting an assertion. We still run the leak tracker. This is a bad idea because

- The assertion error is the real cause, leak messages can be distracting
- Hitting an error means that the test exits early, so its its expected that some cleanup code may have been missed

(I am not sure if the objects are being kept alive as part of the exception object, or if they have just not been cleaned up because the test was aborted)

### Steps To Reproduce

Modify `test_load_log()` in `mantidimaging/gui/test/gui_system_loading_test.py` by adding
```python
self.assertTrue(False)
```
to the end of the test (around line 107).

Run that test
```
pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true --run-system-tests mantidimaging/gui/test/ -k test_load_log
```

### Current Behaviour

Note that there is lots of information printed about leaked images

```
mantidimaging/gui/test/gui_system_loading_test.py::TestGuiSystemLoading::test_load_log 
Items still alive: 4
 Object: <class 'numpy.ndarray'> pyid=139901382450544  ndarray: (100, 128, 128) float32
  item_info.msg='ImageStack IMAT_Flower_Tomo_000000'

 Object: <class 'mantidimaging.core.parallel.utility.SharedArray'> pyid=139902177295440 
  item_info.msg='ImageStack IMAT_Flower_Tomo_000000'

 Object: <class 'numpy.ndarray'> pyid=139901382931056  ndarray: (1, 128, 128) float32
  item_info.msg='ImageStack IMAT_Flower_180deg_000000'

 Object: <class 'mantidimaging.core.parallel.utility.SharedArray'> pyid=139901382591744 
  item_info.msg='ImageStack IMAT_Flower_180deg_000000'

```

### Expected Behaviour

The leak errors should not be show. Only the actual back trace for the failed assertion.

```
====================== FAILURES ================
____ TestGuiSystemLoading.test_load_log ________
...
>       self.assertTrue(False)
E       AssertionError: False is not true

mantidimaging/gui/test/gui_system_loading_test.py:107: AssertionError

```

### Context
Current main

MikeSullivan7:
### Description

In `GuiSystemBase.tearDown` we now check if an AssertionError is raise during the test and only print the leaked object info if the test fails because of too many leaked objects.

We now have the following situations:

1. Test passes and leaked objects within limits:
No Errors printed
2. Test passes and leaked objects exceed limit:
print `leak_tracker` info and `raise RuntimeError(f"Too many leaked objects: {leak_count}")`
3. Test fails and leaked objects within limits:
Test AssertionError details printed only
4. Test fails and leaked objects exceed limits:
Test AssertionError details printed only

Restricting when the `leak_tracker` info is printed will aid when debugging failed tests so the dev will know whether the test is failing due to a broken test or due to too many leaked objects.

### Developer Testing 

- ...

- [ ] As in the linked issue, add an AssertionError manually to `mantidimaging/gui/test/gui_system_loading_test.py::TestGuiSystemLoading::test_load_log` at around line 107 and run `pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true --run-system-tests mantidimaging/gui/test/ -k test_load_log`
- [ ] Check that the expected AssertionError prints but no info about leaked objects prints
- [ ] With the AssertionError, force the leak_tracker to exceed the limit by setting `leak_count_limit = -1` at the top of `TestGuiSystemLoading` and check that only the AssertionError info prints and no info about leaked objects.
- [ ] Now remove the AssertionError and check that the test will now fail due to a raised `RuntimeError(f"Too many leaked objects: {leak_count}")`

### DIFF
--- a/mantidimaging/core/utility/leak_tracker.py
+++ b/mantidimaging/core/utility/leak_tracker.py
@@ -87,7 +87,6 @@ class LeakTracker:
 
     def __init__(self):
         self.tracked_objects = []
-        self.leak_limit_exceeded = False
 
     def add(self, item, msg=""):
         created = traceback.format_stack()[:-1]
--- a/mantidimaging/gui/test/gui_system_base.py
+++ b/mantidimaging/gui/test/gui_system_base.py
@@ -58,16 +58,18 @@ class GuiSystemBase(unittest.TestCase):
 
         # if self._outcome.result._excinfo is None then there were no AssertionErrors during the test
         test_error = self._outcome.result._excinfo  # type: ignore
-        if test_error is None and (leak_count := leak_tracker.count()) > self.leak_count_limit:
-            print("\nItems still alive:", leak_count)
-            leak_tracker.pretty_print(debug_init=False, debug_owners=False, trace_depth=5)
-            try:
+        try:
+            # if the test passed but there were some leaked objects, print basic info
+            if test_error is None and (leak_count := leak_tracker.count()):
+                print("\nItems still alive:", leak_count)
+                leak_tracker.pretty_print(debug_init=False, debug_owners=False, trace_depth=5)
+                # if too many objects leaked, print debug info
                 if leak_count > self.leak_count_limit:
                     print("details:")
                     leak_tracker.pretty_print(debug_init=True, debug_owners=True, trace_depth=5)
                     raise RuntimeError(f"Too many leaked objects: {leak_count}")
-            finally:
-                leak_tracker.clear()
+        finally:
+            leak_tracker.clear()
 
         for widget in self.app.topLevelWidgets():
             if widget.isVisible():


mantidimaging/gui/test/gui_system_base.py
```diff
@@ -56,7 +56,10 @@ def tearDown(self) -> None:
         QTest.qWait(SHORT_DELAY)
         self.assertDictEqual(self.main_window.presenter.model.datasets, {})
 
-        if leak_count := leak_tracker.count():
+        # if self._outcome.result._excinfo is None then there were no AssertionErrors during the test
+        test_error = self._outcome.result._excinfo  # type: ignore
+        if leak_tracker.check_leak_limit_exceeded(self.leak_count_limit) and test_error is None:
+            leak_count = leak_tracker.count()
```
samtygier-stfc: The condition could be something like
`if test_error is None and (leak_count := leak_tracker.count()) > self.leak_count_limit:`

That would save needing the function call, and also would save calling `count()` when not needed.

mantidimaging/core/utility/leak_tracker.py
```diff
@@ -87,6 +87,7 @@ class LeakTracker:
 
     def __init__(self):
         self.tracked_objects = []
+        self.leak_limit_exceeded = False
```
samtygier-stfc: I think this is no longer needed

PR #2968, opened 2025-12-02
Operations: Append Stack GUI only

### LINKED ISSUE #2966: Operations: Append Stacks GUI Frontend
### Current Behaviour

No GUI option for selecting Append Stack operation

### Desired Behaviour

Create a GUI frontend for selecting an Append Stack, see Parent issue for GUI mockup.

MikeSullivan7:
### Description

The GUI for the Operation `Append Stacks` has been addeed to the Operations Window.
The GUI allows the user to choose:

- The Stack which is to be appended to
- The type of data (Tomography / ToF)
- The Stack to append

### Developer Testing 

- ...

- [x] Load some data and check that the Operation `Append Stacks` can be selected in the Operations Window without issue.
- [x] Check that all relevant options can be selected and that changing any of the options does not introduce any errors.
- [x] Currently applying the operation will just return the original data as there is no backend.


RasmiaKulan:
I can validate that these changes work as expected. Thank you! 😊

PR #2982, opened 2025-12-09
Main Window: Allow users to rename the Dataset and individual ImageStacks

### LINKED ISSUE #2978: Main Window: Option to rename Datasets
### Current Behaviour

No option to rename datasets.

### Desired Behaviour

There is currently no option to rename a dataset or datastack in Mantid Imaging 

### Additional Context

Email from Sylvia:

```
Just had a suggestion: If we are working on the same dataset but would like to apply different operations or reconstruction algorithms, one way to do this would be to load the same dataset twice and try the different algorithms. It would be good then to be able to rename the loaded datasets in the main window (I don't see a rename button now).
```

MikeSullivan7:
### Description

The user can now right-click on a Dataset name or on individual stacks (projections, flats, darks, etc) in the TreeView in the Main Window to rename the stack to something custom. 

This is useful when wanting to perform different operations on the same stack, so the user can load in the data, rename it, and then reload the data and have two distinguished stacks with the same data.
When the name of a stack is updated in the Main Window, the name change is reflected in the docked ImageView and in the QComboBoxes of other windows (Spectrum Viewer, Operations, etc.)

### Developer Testing 

- ...

- [x] Load in a Dataset with multiple stacks (sample, flat, darks, etc)
- [x] Right click a stack/dataset in the TreeView and check the "Rename" option appears in the menu
- [x] Give a new name to a dataset and check that the stack name is updated in other opened windows (Spectrum Viewer/Operations).
- [x] Reload the same Dataset again and rename a stack to a different name
- [x] Check that both datasets and stacks remain distinct and correct.

A screenshot example can be seen here, the datasets have been renamed to "Flower_1" and "Flower_2" with the projection stacks renamed to "Operation_gauss" and "Operation_crop":

image

- Eyes tests are needed due to new dialog window, these will be added in a separate PR

### DIFF
--- a/mantidimaging/gui/windows/stack_rename_dialog/presenter.py
+++ b/mantidimaging/gui/windows/stack_rename_dialog/presenter.py
@@ -27,5 +27,6 @@ class StackRenamePresenter(BasePresenter):
         """
         Calls the execute rename dataset method in the main view.
         """
-        assert self.view.new_name_field.text(), "Stack name can not be empty."
+        if self.view.new_name_field.text() == "":
+            raise ValueError("Stack name can not be empty.")
         self.view.parent_view.execute_rename_dataset(self.view.stack, self.view.new_name_field.text())
--- a/mantidimaging/gui/windows/main/view.py
+++ b/mantidimaging/gui/windows/main/view.py
@@ -746,6 +746,7 @@ class MainWindowView(BaseMainWindowView):
                 add_action.triggered.connect(self._add_images_to_existing_dataset)
                 delete_action = self.menuTreeView.addAction("Delete")
                 delete_action.triggered.connect(self._delete_container)
+            if self.dataset_tree_widget.itemAt(position).id in self.presenter.all_dataset_ids:
                 rename_action = self.menuTreeView.addAction("Rename Dataset")
                 rename_action.triggered.connect(self._rename_dataset)
             if self.dataset_tree_widget.itemAt(position).id in self.presenter.all_stack_ids:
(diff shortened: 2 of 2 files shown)


mantidimaging/gui/windows/stack_rename_dialog/presenter.py
```diff
@@ -0,0 +1,31 @@
...
+from __future__ import annotations
+from enum import Enum, auto
+
+from mantidimaging.gui.mvp_base import BasePresenter
+from mantidimaging.gui.windows.stack_properties_dialog.view import StackPropertiesDialog
+
+
+class Notification(Enum):
+    ACCEPTED = auto()
+    DATASET_CHANGED = auto()
```
samtygier-stfc: This is not used. (Think it got inherited from move_stack_dialog)

samtygier-stfc (changes requested):
Working well, but would be good to have a way to cancel the change.

mantidimaging/gui/windows/stack_rename_dialog/presenter.py
```diff
@@ -0,0 +1,31 @@
...
+            if n == Notification.ACCEPTED:
+                self._on_accepted()
+        except RuntimeError as err:
+            self.view.show_error_dialog(str(err))
+
+    def _on_accepted(self) -> None:
+        """
+        Calls the execute rename dataset method in the main view.
+        """
+        assert self.view.new_name_field.text(), "Stack name can not be empty."
```
samtygier-stfc: Better for this to be a `if ...:`, `raise ValueError(...)`, as it is an error condition that the user can cause. Asserts are best for flagging assumption that should be true unless a programming bug has been introduced.

ashleymeigh2 (approved):
I’ve reviewed the changes and tested the rename functionality as described.
The right-click rename option for Datasets and individual ImageStacks works as expected.

Issue #2625, opened 2025-06-16
Core: investigate ImageStack data setter


samtygier-stfc:
### Current Behaviour

`ImageStack` has as setter

```
    @data.setter
    def data(self, other: np.ndarray) -> None:
        self._shared_array.array = other
```

This modifies the `array` attribute within a `SharedArray`. This looks like it might not be safe and it might break the link between a shared array and shared buffer that holds the data.

### Desired Behaviour

Check how this is used, and if there is any unsafe case.

samtygier-stfc:
I think we should keep the `data` setter, but make it safe. Maybe like how its done in the constructor.

```
        if isinstance(data, pu.SharedArray):
            self._shared_array = data
        else:
            self._shared_array = pu.SharedArray(data, None)
```
Then the constructor can just do `self.data = data` and the shared array will be created if needed.

It would also be good to add a `shape` property to `ImageStack`, so that all the places that do `images.data.shape` can be just `images.shape`.

PR #2986, opened 2025-12-11
Investigate SharedArray shared-memory instability and document critical issues

ashleymeigh2:
…al issues

### Description

This PR adds diagnostic comments to the SharedArray implementation to
clarify several problems observed during investigation of Windows
0xC0000005 access-violation crashes. 

Notes were added to highlight:
-nondeterministic cleanup via __del__ leading to premature unlink()
-.array being overwritten instead of bound to the SharedMemory buffer
-unsafe assumptions about buffer lifetime and shape/contiguity

### DIFF
--- a/mantidimaging/core/parallel/utility.py
+++ b/mantidimaging/core/parallel/utility.py
@@ -43,6 +43,7 @@ def create_array(shape: tuple[int, ...], dtype: npt.DTypeLike = np.float32) -> S
 
 
 def _create_shared_array(shape: tuple[int, ...], dtype: npt.DTypeLike = np.float32) -> SharedArray:
+    # No alignment checks. Some dtypes require specific alignment, raw SharedMemory does not ensure it.
     size = full_size_bytes(shape, dtype)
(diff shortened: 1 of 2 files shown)


mantidimaging/core/parallel/utility.py
```diff
@@ -174,11 +186,16 @@ def __init__(self, mem_name: str | None, shape: tuple[int, ...], dtype: npt.DTyp
         self._mem_name = mem_name
         self._shape = shape
         self._dtype = dtype
+        # Cached SharedArray may become invalid if parent frees memory early.
         self._shared_array: SharedArray | None = None
 
     @property
     def array(self) -> np.ndarray:
         if self._shared_array is None:
```
ashleymeigh2:
```
    @property
    def array(self) -> np.ndarray:
        mem = shared_memory.SharedMemory(name=self._mem_name)
        return np.ndarray(self._shape, dtype=self._dtype, buffer=mem.buf)
```

samtygier-stfc:
I think this is all resolved now with the sub issues listed on #2625

PR #2998, opened 2026-01-15
Avoid copy of environment when caching for docs workflow for docs fix up

### LINKED ISSUE #2975: CI: github actions failing with "No space left on device"
### Summary

We have recently had a few failures of the Conda workflow with
```
System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/_diag/Worker_20251202-115741-utc.log' Unhandled exception. System.IO.IOException: No space left on device
```
e.g. https://github.com/mantidproject/mantidimaging/actions/runs/19857775019

Runner vms have 14GB of disk space
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for--private-repositories

samtygier-stfc:
PR #2983 missed some required changes, these are now included

Also make sure the cache key is the same to save cache space

### Developer Testing 

- Check that the docs workflow completes successfully

- [ ] Check that the docs workflow completes successfully

Not needed

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -52,9 +52,9 @@ jobs:
         env:
           CACHE_NUMBER: 0
         with:
-          path: ${CONDA}/envs/mantidimaging-dev
+          path: ~/miniconda3/envs/mantidimaging-dev
           key:
-            ${{runner.os}}-condaenv-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml')}}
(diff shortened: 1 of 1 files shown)


JackEAllen:
If we still run into disk space issues going forward, we may have to consider adding clean up tasks to remove any tar or pip caches. @samtygier-stfc does makes a good point that given we already make a ache of packages, this would cost us in download time so lets try to avoid doing this unless we really have too. 
```
- name: Clean caches
  if: always()
  shell: bash -l {0}
  run: |
    conda clean --tarballs --yes 2>/dev/null || true
    conda clean --index-cache --yes 2>/dev/null || true
    pip cache purge 2>/dev/null || true
```

PR #3000, opened 2026-01-16
Fix PR build fails on test-windows

### LINKED ISSUE #2999: PR build fails on test-windows due to distutils removal in Python 3.12
### Summary

Currently `test-windows` build is failing when creating the development environment using python `./setup.py create_dev_env`. This step runs with the system Python, as the virtual environment has not yet been created.

With the recent update to the `windows-2025` GitHub Actions runners, the default system Python version is now 3.12, where `distutils` has been removed from the standard library. Since Mantid’s setup process still relies on `distutils`, this causes the build to fail at this stage.

The eventual solution is to use pixi in CI and for packaging so we can get rid of the setup.py.

### Solutions to explore

- Use distutils from setuptools in setup.py (`from setuptools import distutils`)
- In windows.yml, instead of using `conda deactivate` , it might work to do `conda activate base` and use python from miniconda instead of system
   - Might need a `conda install distutils`
- in  windows.yml call `python3.11` instead of python (not sure if the system `python` is installed like that)

RasmiaKulan:
### Description
Fixed PR build failures on `test-windows` caused by the removal of `distutils` in Python 3.12 by:

- Replacing `from distutils.core import Command` with `from setuptools import Command` to ensure compatibility with Python 3.12.
- Adding a step in the GitHub Actions workflow to install `setuptools` before running the **Mantid Imaging developer dependencies** job, which was previously failing because the system Python didn’t include `setuptools`.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -3,12 +3,12 @@ name: Testing on Windows
 on:
   push:
     branches:
-      - "main"
-      - "release-*"
+    - 'main'
+    - 'release-*'
   pull_request:
     paths:
-      - "**"
-      - "!docs/**"
+      - '**'
+      - '!docs/**'
   release:
   merge_group:
 concurrency:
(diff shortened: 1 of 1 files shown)


samtygier-stfc:
I've updated the variation version of `test_operations_remove_all_stripes_parameters` the old variation still had the safe apply ticked.

PR #3012, opened 2026-02-04
Fix SPDHG progress bar to show correct number of steps

### LINKED ISSUE #2911: Reconstruction: 2 Extra iterations shown in progress status for SPDHG
### Summary
When running SPDHG the progress window shows two different numbers of iterations
E.g.

Image

Here the expected number was 88 (number of subsets multiplied by number of iterations)

We are not sure why it shows /90 further to the right.

Tried with some different numbers of iterations / subsets and it consistently showed 2 more than expected

ashleymeigh2:
## Issue
Closes #2911

### Description

Fix SPDHG progress bar to show the correct number of steps. The progress bar now matches the actual number of iterations for SPDHG (no extra setup step), while non-stochastic algorithms still include a setup step. This resolves the issue where the progress bar showed two extra steps for SPDHG.

### Developer Testing 

- Manual test: Ran SPDHG and confirmed progress bar matches expected iteration count.

- [ ] SPDHG progress bar matches expected number of iterations (no extra steps)
- [ ] Other algorithms’ progress bars remain unchanged

### DIFF
--- a/mantidimaging/core/reconstruct/cil_recon.py
+++ b/mantidimaging/core/reconstruct/cil_recon.py
@@ -279,10 +279,15 @@ class CILRecon(BaseRecon):
         if recon_params.stochastic:
             # The UI will pass the number of epochs in this case
             num_iter *= num_subsets
-
-        if progress:
-            progress.add_estimated_steps(num_iter + 1)
-            progress.update(steps=1, msg='CIL: Setting up reconstruction', force_continue=False)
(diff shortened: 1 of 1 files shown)


samtygier-stfc:
The trouble with "fixing" this to make the numbers match is that there are more steps than just the PDHG iterations. If we ignore the other steps, then the progress bar is less informative.

Issue #2949, opened 2025-11-24
Recon Window 180° Handling Cleanup


ashleymeigh2:
### Current Behaviour

Review and simplify the logic in the recon Window related to handling 180° datasets. Remove redundant code paths and update UI messages.

### Desired Behaviour

No redundant 180° operations appear in the Operations tab.
Code paths using 180° stacks are unified under the new correlation model.

### Additional Context

Completes the broader #2685 refactor to modernize and simplify 180° file handling.
Follows on from #2942 (projection pair selection) and #2854 (consistent error handling).

PR #3028, opened 2026-02-09
Part 1 - Remove all legacy 180° stack support

ashleymeigh2:
## Issue
realted #2685 #2949

### Description

Remove presenter-side checks that enforced the presence and shape of a
dedicated 180° projection stack, along with their associated unit tests.

These checks relied on legacy assumptions about reconstruction workflows
and are no longer valid now that CoR finding is angle-based.

### Developer Testing

No documentation updates are required.

### DIFF
--- a/mantidimaging/core/io/loader/loader.py
+++ b/mantidimaging/core/io/loader/loader.py
@@ -189,7 +189,7 @@ def create_loading_parameters_for_file_path(file_path: Path) -> LoadingParameter
     loading_parameters.image_stacks[FILE_TYPES.SAMPLE] = ImageParameters(sample_fg, sample_fg.log_path,
                                                                          sample_fg.shutter_count_path)
 
-    for file_type in [ft for ft in FILE_TYPES if ft.mode in ["images", "180"]]:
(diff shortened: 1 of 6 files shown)


mantidimaging/core/io/loader/loader.py
```diff
@@ -189,7 +189,7 @@ def create_loading_parameters_for_file_path(file_path: Path) -> LoadingParameter
     loading_parameters.image_stacks[FILE_TYPES.SAMPLE] = ImageParameters(sample_fg, sample_fg.log_path,
                                                                          sample_fg.shutter_count_path)
 
-    for file_type in [ft for ft in FILE_TYPES if ft.mode in ["images", "180"]]:
+    for file_type in [ft for ft in FILE_TYPES if ft.mode == "images"]:
```
samtygier-stfc: We still want to look for an load a 180 file if there is one. Might also mean the change in `loader_test.py` is not needed.

samtygier-stfc (changes requested):
Not a full review. But I don't think this part should be changing.

samtygier-stfc (approved):
That looks good. It now lets use the correlate Tilt/CoR tool without a 180 projection loaded.

PR #3031, opened 2026-02-12
Spectrum Viewer: Export data table and image view are only updated within Export tab

### LINKED ISSUE #3022: Spectrum viewer: decouple fitting and export tab
### Current Behaviour

Currently running fits on the fitting tab causes the corresponding row in the export tab to be updated. This adds complexity for not much benefit.

### Desired Behaviour

Only update table on export tab, when running fit all from the export tab

### Additional Context

Similar to work in #2992

MikeSullivan7:
### Description

This PR decouples the logic between the Fitting and Image tabs in the Spectrum Viewer to the Export tab. Previously, when the stack was changed or when a fit was made to the data, the Export tab would update behind the scenes which is not neccessary and leads to increased complexity in the logic with not much benefit.

Now the fitting parameter table and the mean image tabs in the Export tab only update when it is needed and when they are viewed. 

### Developer Testing 

- ...

- [ ] Load in data to MI and go to the Spectrum Viewer
- [ ] Go to the `Fitting` tab and place the ROI
- [ ] Go to the `Export` tab and click `Fit All`, the Table of paramters should update
- [ ] Go back to the `Fitting` tab and move the ROI, click on the `From ROI` and `Run fit` buttons.
- [ ] On the `Export` tab, the Table should not have updated. Now click on the `Fit All` button, the Table should now update accordingly to match the parameters shown on the `Fitting` tab.
- [ ] Go to the `Image` tab on the `Export` tab and make note of the image
- [ ] Go back to the `Image` tab and change the active stack
- [ ] Return to the `Export` tab and check that the image has been updated.

Some system tests have been added to check the above behaviours.

### DIFF
--- a/mantidimaging/gui/test/gui_system_spectrum_test.py
+++ b/mantidimaging/gui/test/gui_system_spectrum_test.py
@@ -215,22 +215,16 @@ class TestGuiSpectrumViewer(GuiSystemBase):
     def test_fit_model_does_not_update_export_table(self):
         self.spectrum_window.formTabs.setCurrentIndex(1)
         QTest.qWait(SHORT_DELAY)
-        self.spectrum_window.formTabs.setCurrentIndex(2)
-        QTest.mouseClick(self.spectrum_window.exportSettingsWidget.fitAllButton, Qt.MouseButton.LeftButton)
-        QTest.qWait(SHORT_DELAY)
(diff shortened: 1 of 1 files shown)


samtygier-stfc (approved):
Thanks, works well here.

PR #3049, opened 2026-03-11
Limit recon algorithm selection based on geometry

### LINKED ISSUE #2888: Reconstruction: Limit selection of recon algo based on geometry
### Summary

Currently CIL PDHG TV is the only recon algorithm that supports cone beam.

If the selected stack has cone geometry the other algorithms should be disabled.

### Steps To Reproduce

Load a cone data set
(or load any dataset and use the geometry window to set a cone geometry)

### Current Behaviour

Any algorithm can be selected in the recon window, but gridrec, FBP and SIRT will only use the geometry params they are aware of and will do a parallel recon. 

### Expected Behaviour

Only compatable recons are selectable

### Context

Need since the cone geometry was introduced as part of #2482

ashleymeigh2:
### Description

Limit reconstruction algorithm selection based on geometry type:

Only CIL PDHG TV is enabled when the selected stack has cone beam geometry (GeometryType.CONE3D); all other algorithms are disabled.
For parallel geometry, all algorithms remain enabled.
Presenter updates algorithm options after stack selection or geometry change.

### Developer Testing 

- Manually tested with both cone and parallel geometry stacks to confirm correct algorithm enabling/disabling.

- [x]  When a stack with cone geometry is selected, only CIL PDHG TV is selectable in the recon window.
- [x]  When a stack with parallel geometry is selected, all algorithms are selectable.
 - [x]  Changing geometry updates the available algorithms as expected.

### DIFF
--- a/docs/release_notes/next/feature-2888-Limit-recon-algorithm-selection-based-on-geometry
+++ b/docs/release_notes/next/feature-2888-Limit-recon-algorithm-selection-based-on-geometry
@@ -1,0 +1,1 @@
+#2888: Limit recon algorithm selection based on geometry


samtygier-stfc (approved):
Looks good now, tested locally. Thanks

PR #3051, opened 2026-03-12
Run fit on sub ROIs to update export table

RasmiaKulan:
### Description
Added the ability to run fit on sub ROIs to update the export table by:
- Updating `fit_all_regions` to loop over binned subROIs in Image Mode, allowing fits to be run for both Image Mode and ROI Mode.

Refactored code to apply DRY principle by:
-  Creating `compute_spectrum_fit` that handles one spectrum and returns (result, rss, rss_per_dof, status) while cleanly handling exceptions.
- Creating `fit_roi_and_update_table` to replace the inlined logic in `fit_all_regions` where it fits a single ROI/subROI and updates the export table consistently.

### Developer Testing 

- I have tested these changes with an extreme manual testing scenario where I kept the region as 64x64 with bin size 10 and step size 1. It did run and update the export table eventually but while it was running in the background, the Export tab header had the `Not Responding` message appear. This is expected until the PR to use async progress task is merged.

- [x] Upload an appropriate dataset in MantidImaging I used (`python -m mantidimaging --path C:\Programming\Mantid_Data\small_fe\small_fe\sample`)
- [x]  Open the Spectrum Viewer and select the Image Mode
- [x] Select a region (I used a 40x40 region)
- [x] Select a bin size and bin step (I applied 10 for both)
- [x] Select a Fitting Function (try it with both)
- [x] Switch to the Export tab and run `Fit All`. This should update the Export Table with the fit run on all possible sub ROI's in the region

image

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -61,6 +61,7 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         self.view = view
         self.main_window = main_window
         self.model = SpectrumViewerWindowModel(self)
+        #self.roiSelectionWidget = ROISelectionWidget(self)
         self.export_mode = ExportMode.ROI_MODE
         self.main_window.stack_modified.connect(self.handle_stack_modified)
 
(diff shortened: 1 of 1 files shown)


JackEAllen (changes requested):
From manual testing, I spotted a small bug with the `row` and `col` being on the wrong order when iterating through bins in the `fit_all_regions()` method.
Other than that, just a few small suggestions but otherwise is coming along nicely!

JackEAllen (approved):
Thanks for the changes. Very cleanly implemented. I've validated tests pass locally tested functionally. Fitting all bins performs as expected so lets merge this in 🥳

PR #3072, opened 2026-03-27
Spectrum Viewer: ROIs can be loaded from CSV file

### LINKED ISSUE #3043: Spectrum Viewer - Load ROIs from CSV Button and Dialog Window
### Current Behaviour

IMAT have voiced that it would be useful to load a list of ROI's into the Spectrum Viewer to make it easier to use identical ROI's across multiple samples where the number of ROI's and their positions should remain the same.

Being able to load in a list of ROI's would save IMAT time rather than manually re-drawing each ROI which can prove cumbersome when working with 10 ROI's and multiple similar samples 

### Desired Behaviour

A dedicated "Load ROIs" button in the Image tab of the Spectrum Viewer provides quick, in-context access. Clicking it opens a dialog with two steps:

* Clear existing ROIs? / Append to existing ROIs - A checkbox (checked by default) to discard previously set ROI's before loading new ROI's. If unchecked, new ROI's a appended to table of ROIs without clearing previous ROIs.
* File selection - a "Browse..." button opens the system file explorer, filtered to CSV files.
This flow is optimised for ad-hoc, per-session use.
* The loaded ROIs populate the table of ROIs and the image view in their expected locations

* If a user loads ROIs then closes the spectrum viewer, should the ROIs persist when the spectrum viewer is opened again? - I think for simplicity maybe not
* Loading of ROIs should be logged in history
* Should there be a practical limit to the number of ROIs that can be loaded? If a CSV had thousands, this would likely crash or at the very least degrade the viewers performance - A warning be displayed.
* Sub-ROIs / binned are out of scope for this issue.

### Importance of Feature

Reduces overall workflow within spectrum viewer when working with a number of similar samples
Improves reproducibility of spectrum viewer workflow

MikeSullivan7:
### Description

Requested by the IMAT scientists, this PR allows the user to load in ROIS from a CSV file and have the ROIs be displayed in the Spectrum Viewer. The expected format of the loaded CSV file matches that which is created when exporting the ROIs from the Spectrum VIewer. This will allow the user to set the ROIs they want in the SV, export them to a CSV file, and then load them in another instance of MI without any intervention or additional work. An example of the expected CSV file is seen below:

image

It should be noted that any instance of RITS_ROI will be ignored in the loading of the ROIs.

### Developer Testing 

- ...

- [ ] Load data and open Spectrum Viewer
- [ ] Add a few ROIS and move them so that they are distinct and note their placements
- [ ] Click "Export Spectrum" which will save a CSV with the ROI coordinates.
- [ ] Delete all ROIs in the Spectrum Viewer and click "Load" button
- [ ] Load in the CSV file of ROI coordinates which you have just created
- [ ] Check the the ROIs are in the places you would expect with the same names.
- [ ] alternatively, alter the CSV file before loading to check that the CSV file is being read in correctly.

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -732,3 +732,11 @@ class SpectrumViewerWindowModel:
         @param map_array: Parameter Map
         """
         saver.write_img(map_array, str(path))
+
+    def load_rois_from_csv(self, path: Path):
+        loaded_rois = []
+        with open(path, encoding='utf-8', mode='r') as f:
+            csv_reader = csv.DictReader(f)
+            header = csv_reader.fieldnames
+            for line in csv_reader:
+                loaded_rois.append(line)
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -733,16 +733,17 @@ class SpectrumViewerWindowModel:
         """
         saver.write_img(map_array, str(path))
 
-    def load_rois_from_csv(self, path: Path) -> list[str]:
+    def load_rois_from_csv(self, path: Path) -> list[dict[str, str]]:
         loaded_rois_list = []
         with open(path, encoding='utf-8') as f:
             csv_reader = csv.DictReader(f)
             header = csv_reader.fieldnames
             for line in csv_reader:
-                loaded_rois_list.append(str(line))
+                loaded_rois_list.append(line)
 
-        assert header is not None, "No header has been found in the CSV file, please check the file."
-        assert all(field in header for field in ["X Min", "Y Min", "X Max", "Y Max"]), \
-            "ROI fields not found in CSV file. Please use exported ROIs instead."
+        if header is None:
+            raise RuntimeError("No header has been found in the CSV file, please check the file.")
+        if not all(field in header for field in ["X Min", "Y Min", "X Max", "Y Max"]):
+            raise RuntimeError("ROI fields not found in CSV file. Please use exported ROIs instead.")
 
         return loaded_rois_list
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -16,14 +16,11 @@ from mantidimaging.core.io.csv_output import CSVOutput
 from mantidimaging.core.io import saver
 from mantidimaging.core.io.instrument_log import LogColumn, ShutterCountColumn
 from mantidimaging.core.utility.progress_reporting import Progress
-from mantidimaging.core.utility.sensible_roi import SensibleROI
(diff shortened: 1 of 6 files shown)


mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -461,6 +461,14 @@ def handle_export_csv(self) -> None:
         )
         LOG.info("CSV export successful: file saved to '%s'", path)
 
+    def handle_load_csv(self) -> None:
+        path = self.view.get_load_csv_filename()
+        if not path:
+            return
+        path = path.with_suffix(".csv") if path.suffix != ".csv" else path
+        print(f"handle_load_csv: {path=}")
```
ashleymeigh2: Remove debug print

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -461,6 +461,14 @@ def handle_export_csv(self) -> None:
...
+        self.model.load_rois_from_csv(path)
```
ashleymeigh2: it looks like load_rois_from_csv in the model is responsible not only for parsing the CSV, but also for triggering ROI creation in the UI. A cleaner approach would be for the model to return parsed ROI data, and let the presenter handle creation.

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -591,3 +591,29 @@ def set_fit_results(self, results: list[tuple[str, SpectrumFitResult]] | None) -
 
     def get_fit_results(self) -> list[tuple[str, SpectrumFitResult]] | None:
         return self.fit_results
+
+    def load_rois_from_csv(self, path: Path):
+        loaded_rois_list = []
```
ashleymeigh2, line 737: It would be good to validate the CSV structure in load_rois_from_csv right after creating the DictReader.

mantidimaging/gui/windows/spectrum_viewer/presenter.py
ashleymeigh2: If an error occurs during loading there is no feedback to the user

mantidimaging/gui/windows/spectrum_viewer/model.py
MikeSullivan7: is this not what this piece of the code is doing?
```
 assert header is not None, "No header has been found in the CSV file, please check the file."
assert all(field in header for field in ["X Min", "Y Min", "X Max", "Y Max"]), \
            "ROI fields not found in CSV file. Please use exported ROIs instead."
```

samtygier-stfc: Should probably be a `if ...: raise ...` for an error that can occur due to something being wrong at run time.

ashleymeigh2 (dismissed):
Looks good

samtygier-stfc (approved):
Thanks for fixes

Issue #3065, opened 2026-03-23
Windows GUI Tests: Re-evaluate Longer Running Tests (20/30 Second Range) TIMEBOXED


JackEAllen:
### Current Behaviour

There seems to have been an increase in the number of longer running Windows GUI tests, taking between 20 and 30 seconds to complete. Some of them, such as:
```python
mantidimaging/gui/test/gui_system_reconstruction_test.TestGuiSystemReconstruction.test_refine_stress()
``` 
We expect `test_refine_stress` to take longer as the test tests multiple things which have helped us previously catch a bug, however other tests should be faster. 

Explore Windows GUI tests to evaluate and implement optimisations to reduce test times. 

### Desired Behaviour

Windows GUI tests where possible should be optimised to reduce execution times

### Potential Pitfalls or Risks to Avoid

Could break tests silently meaning we don't catch previously caught errors

### Importance of Feature

Reduces the likelihood of a timeout causing windows tests to fail which can inhibit the creation of package creation.

### Additional Context

```Python
============================ slowest 10 durations =============================
47.97s call     mantidimaging/gui/test/gui_system_spectrum_test.py::TestGuiSpectrumViewer::test_no_rename_for_bad_roi_name_2_all
39.41s call     mantidimaging/gui/test/gui_system_spectrum_test.py::TestGuiSpectrumViewer::test_spectrum_window_opens_with_data_in_default_state
39.35s call     mantidimaging/gui/test/gui_system_reconstruction_test.py::TestGuiSystemReconstruction::test_refine_stress
35.81s call     mantidimaging/gui/test/gui_system_reconstruction_test.py::TestGuiSystemReconstruction::test_refine
30.77s call     mantidimaging/gui/test/gui_system_spectrum_test.py::TestGuiSpectrumViewer::test_switch_tabs_updates_property_box
30.36s call     mantidimaging/gui/test/gui_system_spectrum_test.py::TestGuiSpectrumViewer::test_switch_from_joined_to_scatter_plot
29.91s call     mantidimaging/gui/test/gui_system_spectrum_test.py::TestGuiSpectrumViewer::test_remove_roi
28.99s call     mantidimaging/gui/test/gui_system_windows_test.py::TestGuiSystemWindows::test_open_operations
25.58s call     mantidimaging/gui/test/gui_system_spectrum_test.py::TestGuiSpectrumViewer::test_normalisation_toggle
[...]

PR #3073, opened 2026-03-30
SystemTest: clear leak tracker in setUp so tests it always starts clean

samtygier-stfc:
## No issue
Related to #3065

### Description

We reset the leak track between tests so that it is easier to debug where the leaks are happening. Do this in `setUp()` rather than `tearDown()` so that we can be sure that tests start with a clean state.

For example this prevents ImageStacks that created during test collection being counted by the first test that runs.

Also limit test collection to the system test directory for system tests in github action workflows. This saves the import and search of a large number of unit test files.

### Developer Testing 

Without the change I get a leak error running
`python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true --run-system-tests -k test_no_rename_for_bad_roi_name_2_all`

- [ ] ...

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -114,7 +114,7 @@ jobs:
       - name: GUI Tests System
         shell: bash -l {0}
         run: |
-          xvfb-run --auto-servernum /bin/time -v python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true --run-system-tests --durations=10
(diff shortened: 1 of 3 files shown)


JackEAllen (approved):
Changes look good. Clearing at the start of `setUp()` before anything is created is a cleaner implementation here and should mean that if `tearDown()` raises a `RuntimeError` for too many leaks, the `leak_tracker.clear()` will still happen which likely wasn't the case prior.

PR #3084, opened 2026-04-07
2955 spectrum viewer map fitted params

### LINKED ISSUE #2955: Spectrum Viewer: Show the fitted params as an image
### Current Behaviour

In "image mode" the parameters are fitted at each sub ROI on a sample image and saved to a table. The parameters exported can be understood more [HERE](https://mantidproject.github.io/mantidimaging/user_docs/tutorials/quick_start_guides/time_of_flight_workflow_guide.html)

Currently we do not overlay / map the fitted parameters to the image view in the Export tab.

### Desired Behaviour

Add Bragg edge fitting parameter mapping over sample image in export tab.

This will involve turning the fitted parameters into a spatial "parameter map" across the sample to show something such as strain - f(x,y)

For the time being, the implementation should only be running on bins, not single pixels due the the performance hit this would cause, so treating an N x N block as one "effective pixel", by averaging or summing. The map can be sampled with a step size smaller than N, producing overlapping bins. Something like a 100x100 ROI can yield 99x99 overlapping 2x2 binned positions like a rolling average. 

**Desired workflow:**
* Choose a fitted parameter to visualise (e.g. edge position, edge height, width etc
* Generate a 2D map where each sub-ROI contributes a value placed at the correct location on the sample image. 
* Display the map as an overlay aligned with the sample image.

- Scaling and outliers should be consider where a few extreme values could ruin the colour scale. Providing a means of limiting min/max values and percentile-based autoscaling may be needed.
- May want to explore either as part of this issue or a later issue the exclusion of poor fits which could affect mapping quality.
- Ensure units in legend and tooltip (if used) clearly state units
- A user may want to export the mapped image, we should consider how to implement this

JackEAllen:
### Description

- Create 2D mapping overlay of fit per parameter over sample image in Export Image view
- Allow for user control of mapping opacity to control if fit parameter map or sample is more visible
- Add 3 outlier colour range selections (1.5xIQR, Median+-3XMAD and lower and upper 2 percent clipping to improve parameter overlay visibility when more extreme outliers exist
- Allow user selected chi2 threshold to filter fit quality in parameter map overlay where if values are less than threshold, they are displayed as NaN/transparent
- Added export of mapped area only and mapped area over sample image. Mapped area only can be loaded in MI while mapped area over sample area include default viridis colouring, so not able to be loaded into MI, but useful for presentations

### Developer Testing 

- I have verified unit tests pass locally: `python -m pytest -vs`, `pixi run test-cpu` or `pixi run test-gpu`
- Parameter overlay mapped image displayed by default on completion of "Fit All"
- Parameter overlay shows viridis colour map
- ROI shape and position correctly mapped over sample image in Export Tab Image view
- ROI shape and position correctly mapped over sample image in "Full Sample image" exported Image
- "Full sample image" exported matches opacity, colour range and chi2 threshold in Spectrum Viewer Export tab Image View
- "Full sample image" exported image shows full sample shape and mapped parameters using OS image viewer
- "Mapped Region only" matches current parameter selection  in Spectrum Viewer Export tab Image View
- "Mapped Region only" possible to open in Mantid Imaging
- IQR, MAD and percentile Colour ranges correctly update colour range upper lower and upper bounds of fitted parameter map overlay (IQR and MAD should be the most accurate)

- [ ] Parameter overlay mapped image displayed by default on completion of "Fit All"
- [ ] Parameter overlay shows viridis colour map
- [ ] ROI shape and position correctly mapped over sample image in Export Tab Image view
- [ ] ROI shape and position correctly mapped over sample image in "Full Sample image" exported Image
- [ ] "Full sample image" exported matches opacity, colour range and chi2 threshold in Spectrum Viewer Export tab Image View
- [ ] "Full sample image" exported image shows full sample shape and mapped parameters using OS image viewer
- [ ] "Mapped Region only" matches current parameter selection  in Spectrum Viewer Export tab Image View
- [ ] "Mapped Region only" possible to open in Mantid Imaging
- [ ] IQR, MAD and percentile Colour ranges correctly update colour range upper lower and upper bounds of fitted parameter map overlay (IQR and MAD should be the most accurate)

**Fitted Parameter Map over Export Image view:**
image

**Exported Full Sample Map Image:**
image

**Exported Mapped Parameters only loaded into Mantid Imaging:**
image

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
@@ -2,11 +2,15 @@
 # SPDX - License - Identifier: GPL-3.0-or-later
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, TYPE_CHECKING
 
 import numpy as np
+from PyQt5.QtGui import QTransform
 from PyQt5.QtWidgets import QWidget, QVBoxLayout
 import pyqtgraph as pg
+
+if TYPE_CHECKING:
+    from mantidimaging.core.utility.sensible_roi import ROIBinner
 
 _graveyard: list[Any] = []
 
@@ -31,6 +35,11 @@ class ExportImageViewWidget(QWidget):
         self.image_view.ui.histogram.hide()
 
         layout.addWidget(self.image_view)
+
+        self.param_overlay = pg.ImageItem()
+        self.param_overlay.hide()
+        self.image_view.getView().addItem(self.param_overlay)
+
         self.clear()
 
     def update_image(self, image: np.ndarray | None, autoLevels: bool = True) -> None:
@@ -46,6 +55,33 @@ class ExportImageViewWidget(QWidget):
     def clear(self) -> None:
         """Show a blank canvas."""
         self.image_view.setImage(np.zeros((1, 1), dtype=np.float32), autoLevels=True)
+        self.image_view.ui.histogram.setImageItem(self.image_view.imageItem)
+        self.image_view.ui.histogram.hide()
+        self.param_overlay.hide()
+
+    def populate_parameter_selector(self, param_names: list[str]) -> None:
+        """Populate the parameter selector combobox with fitted parameter names"""
+        self.parent().exportSettingsWidget.populate_parameter_selector(param_names)
+
+    def show_parameter_map(self, map_array: np.ndarray, binner: ROIBinner) -> None:
+        """Display the parameter map with 50% transaparency and viridis over sample"""
+
+        self.param_overlay.setColorMap('viridis')
+        self.param_overlay.setOpacity(0.5)
+
+        # Align mapping with sample and scale to match binned resolution
+        transform = QTransform()
+        transform.translate(binner.left_indexes[0], binner.top_indexes[0])
+        transform.scale(binner.step_size, binner.step_size)
+        self.param_overlay.setImage(map_array.T, autoLevels=False)
+        self.param_overlay.setTransform(transform)
+        self.param_overlay.show()
+
+        # connect histogram to parameter values and match colourmap
+        hist = self.image_view.ui.histogram
+        hist.setImageItem(self.param_overlay)
+        hist.gradient.setColorMap(pg.colormap.get('viridis'))
+        hist.show()
 
     @property
     def image_data(self) -> np.ndarray | None:
--- a/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
@@ -63,9 +63,13 @@ class ExportImageViewWidget(QWidget):
         """Populate the parameter selector combobox with fitted parameter names"""
         self.parent().exportSettingsWidget.populate_parameter_selector(param_names)
 
-    def show_parameter_map(self, map_array: np.ndarray, binner: ROIBinner, opacity: float = 0.5) -> None:
+    def show_parameter_map(self,
+                           map_array: np.ndarray,
+                           binner: ROIBinner,
+                           opacity: float = 0.5,
+                           levels: tuple[float, float] = (0.0, 1.0)) -> None:
         """Display the parameter map with 50% transaparency and viridis over sample"""
-
+        lower_level, upper_level = levels
         self.param_overlay.setColorMap('viridis')
         self.param_overlay.setOpacity(opacity)
 
@@ -74,6 +78,7 @@ class ExportImageViewWidget(QWidget):
         transform.translate(binner.left_indexes[0], binner.top_indexes[0])
         transform.scale(binner.step_size, binner.step_size)
         self.param_overlay.setImage(map_array, autoLevels=True)
+        self.param_overlay.setLevels(lower_level, upper_level)
         self.param_overlay.setTransform(transform)
         self.param_overlay.show()
 
@@ -81,6 +86,7 @@ class ExportImageViewWidget(QWidget):
         hist = self.image_view.ui.histogram
         hist.setImageItem(self.param_overlay)
         hist.gradient.setColorMap(pg.colormap.get('viridis'))
+        hist.setLevels(lower_level, upper_level)
         hist.show()
 
     @property
--- a/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
@@ -63,17 +63,17 @@ class ExportImageViewWidget(QWidget):
         """Populate the parameter selector combobox with fitted parameter names"""
         self.parent().exportSettingsWidget.populate_parameter_selector(param_names)
 
-    def show_parameter_map(self, map_array: np.ndarray, binner: ROIBinner) -> None:
+    def show_parameter_map(self, map_array: np.ndarray, binner: ROIBinner, opacity: float = 0.5) -> None:
         """Display the parameter map with 50% transaparency and viridis over sample"""
 
         self.param_overlay.setColorMap('viridis')
-        self.param_overlay.setOpacity(0.5)
+        self.param_overlay.setOpacity(opacity)
 
         # Align mapping with sample and scale to match binned resolution
         transform = QTransform()
         transform.translate(binner.left_indexes[0], binner.top_indexes[0])
         transform.scale(binner.step_size, binner.step_size)
-        self.param_overlay.setImage(map_array.T, autoLevels=False)
+        self.param_overlay.setImage(map_array, autoLevels=True)
         self.param_overlay.setTransform(transform)
         self.param_overlay.show()
 
--- a/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
@@ -42,7 +42,7 @@ class ExportImageViewWidget(QWidget):
 
         self.clear()
 
-    def update_image(self, image: np.ndarray | None, autoLevels: bool = True) -> None:
+    def update_image(self, image: np.ndarray | None) -> None:
         if image is None:
             self.clear()
             return
@@ -50,7 +50,9 @@ class ExportImageViewWidget(QWidget):
         arr = np.asarray(image)
         if arr.ndim == 3 and arr.shape[0] > 0:
             arr = arr.mean(axis=0)
-        self.image_view.setImage(arr, autoLevels=autoLevels)
+        finite = arr[np.isfinite(arr)]
+        levels = (float(np.percentile(finite, 0.5)), float(np.percentile(finite, 99.5))) if finite.size else (0.0, 1.0)
+        self.image_view.setImage(arr, autoLevels=False, levels=levels)
 
     def clear(self) -> None:
         """Show a blank canvas."""
@@ -68,7 +70,7 @@ class ExportImageViewWidget(QWidget):
                            binner: ROIBinner,
                            opacity: float = 0.5,
                            levels: tuple[float, float] = (0.0, 1.0)) -> None:
-        """Display the parameter map with 50% transaparency and viridis over sample"""
+        """Display the parameter map overlay using viridis with the given opacity and colour levels."""
         lower_level, upper_level = levels
         self.param_overlay.setColorMap('viridis')
         self.param_overlay.setOpacity(opacity)
@@ -77,8 +79,8 @@ class ExportImageViewWidget(QWidget):
         transform = QTransform()
         transform.translate(binner.left_indexes[0], binner.top_indexes[0])
         transform.scale(binner.step_size, binner.step_size)
-        self.param_overlay.setImage(map_array, autoLevels=True)
-        self.param_overlay.setLevels(lower_level, upper_level)
+
+        self.param_overlay.setImage(map_array, autoLevels=False, levels=(lower_level, upper_level))
         self.param_overlay.setTransform(transform)
         self.param_overlay.show()
 
--- a/mantidimaging/gui/windows/spectrum_viewer/model.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/model.py
@@ -632,20 +632,20 @@ class SpectrumViewerWindowModel:
         @param param_name: The name of the fitted parameter to map
         @param binner: The ROIBinner used when generating for fit
         @param  chi2_threshold: Optional upper bound on rss_per_dof. Exclude fits if reduced chi2 higher than threshold
-        @return np.ndarray: 2D array of parameters (n_cols, n_rows) matching binner dimensions.
+        @return np.ndarray: 2D array of parameters (n_rows, n_cols) matching binner dimensions.
         """
         if self.fit_results is None:
             raise ValueError("No fit results available. Run 'Fit All' before building a parameter map.")
 
         n_cols, n_rows = binner.lengths()
-        param_map = np.full((n_cols, n_rows), np.nan, dtype=np.float64)
+        param_map = np.full((n_rows, n_cols), np.nan, dtype=np.float64)
 
         for col, row in np.ndindex(n_cols, n_rows):
             if (fit_result := self._good_fits_by_roi_name.get(binner.get_roi_name(col, row))) is None:
                 continue
             if chi2_threshold is not None and fit_result.rss_per_dof > chi2_threshold:
                 continue
-            param_map[col, row] = fit_result.params[param_name]
+            param_map[row, col] = fit_result.params[param_name]
 
         return param_map
 
@@ -666,7 +666,7 @@ class SpectrumViewerWindowModel:
         strategies = {
             ColourRangeMode.IQR: self._colour_levels_iqr,
             ColourRangeMode.MAD: self._colour_levels_mad,
-            ColourRangeMode.PERCENTILE: self._colour_levels_percentile,
+            ColourRangeMode.PERCENTILE: self.colour_levels_percentile,
         }
         lower, upper = strategies[mode](finite_values)
 
@@ -675,8 +675,8 @@ class SpectrumViewerWindowModel:
         return lower, upper
 
     @staticmethod
-    def _colour_levels_percentile(finite_values: np.ndarray) -> tuple[float, float]:
-        """Show only 2nd-98th percentile range"""
+    def colour_levels_percentile(finite_values: np.ndarray) -> tuple[float, float]:
+        """Return the 2nd-98th percentile range of the given finite values."""
         return float(np.percentile(finite_values, 2)), float(np.percentile(finite_values, 98))
 
     @staticmethod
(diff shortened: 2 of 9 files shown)


JackEAllen:
Screenshot tests excluded from this PR - will be created as a separate PR. Unit tests should cover cover logic for the time being

JackEAllen:
```python
-------------------------------- live log call ---------------------------------
ERROR    mantidimaging.core.utility.cuda_check:cuda_check.py:25 CuPy installed, but unable to load CUDA
self.live_viewer_window=<mantidimaging.gui.windows.live_viewer.view.LiveViewerWindowView object at 0x7ff0f63c2f90>
qt_message_handler: line: 0, func: None, file: None
WARNING: QFileSystemWatcher::removePaths: list is empty
```
I think the Windows and Conda GUI System Tests failures are completely unrelated to this PR and an issue with something like the timing in `test_open_close_intensity_profile`  and possibly the mask condition in the same test - possible checking `mean_nan_mask) > 0` to prevent passing empty array from making a missing render? 🤷‍♂️ - either way we should explore and resolve separately @samtygier-stfc can you make an issue to look into this while i'm away as this could affect release if it become less flakey/infrequent

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+            if result.is_good_fit and np.isfinite(result.rss_per_dof) and result.rss_per_dof > 0
+        ]
+        return float(np.percentile(valid_chi2_values, percentile)) if valid_chi2_values else None
+
+    def build_parameter_map(self,
+                            param_name: str,
+                            binner: ROIBinner,
+                            chi2_threshold: float | None = None) -> np.ndarray:
+        """
+        Build a 2D parameter map from stored fit results for the given parameter name.
```
ashleymeigh2, line 627: if a fit result is missing for a bin, it’s skipped (left as NaN). Adding logging a warning if many bins are missing,

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+        @param binner: The ROIBinner used when generating for fit
+        @param  chi2_threshold: Optional upper bound on rss_per_dof. Exclude fits if reduced chi2 higher than threshold
+        @return np.ndarray: 2D array of parameters (n_rows, n_cols) matching binner dimensions.
+        """
+        if self.fit_results is None:
+            raise ValueError("No fit results available. Run 'Fit All' before building a parameter map.")
+
+        n_cols, n_rows = binner.lengths()
+        param_map = np.full((n_rows, n_cols), np.nan, dtype=np.float64)
+
```
ashleymeigh2, line 642: For large datasets, iterating with np.ndindex(n_cols, n_rows) is fine, but if the number of bins grows, it might be good to add vectorized approach or caching the ROI-to-index mapping.

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+        for col, row in np.ndindex(n_cols, n_rows):
+            if (fit_result := self._good_fits_by_roi_name.get(binner.get_roi_name(col, row))) is None:
+                continue
+            if chi2_threshold is not None and fit_result.rss_per_dof > chi2_threshold:
+                continue
+            param_map[row, col] = fit_result.params[param_name]
```
ashleymeigh2, line 648: Use param_map[row, col] = fit_result.params.get(param_name, np.nan) to safely handle missing parameters. or log a warning if a parameter is missing.

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+        @param map_array: parameter map
+        @param binner: The binner used when generating the map
+        @return: Array of shape (height, width) matching the sample stack dimensions
+        """
+        img_height, img_width = self.get_image_shape()
+        full_map = np.full((img_height, img_width), np.nan, dtype=np.float32)
+        step = binner.step_size
+
+        n_cols, n_rows = binner.lengths()
+        for col, row in np.ndindex(n_cols, n_rows):
```
ashleymeigh2, line 716: for row, col in np.ndindex(n_rows, n_cols):
    param_value = map_array[row, col]

map_array is indexed as [row, col], it would be clearer  with NumPy to iterate in (row, col) order

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+            param_value = map_array[row, col]
```
ashleymeigh2, line 717: no check that map_array matches the expected (n_rows, n_cols)

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+        @param percentile: The percentile of rss_per_dof values to return as the chi2 threshold
+        @return: The chi2 threshold value or None if no valid fit results are available
+        """
+        if self.fit_results is None:
+            return None
+        valid_chi2_values = [
+            result.rss_per_dof for _, result in self.fit_results
+            if result.is_good_fit and np.isfinite(result.rss_per_dof) and result.rss_per_dof > 0
+        ]
+        return float(np.percentile(valid_chi2_values, percentile)) if valid_chi2_values else None
```
ashleymeigh2, line 620: if not valid_chi2_values:
    LOG.warning("No valid chi2 values found for threshold computation")

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+        """
+        if self.fit_results is None:
+            raise ValueError("No fit results available. Run 'Fit All' before building a parameter map.")
+
+        n_cols, n_rows = binner.lengths()
+        param_map = np.full((n_rows, n_cols), np.nan, dtype=np.float64)
+
+        for col, row in np.ndindex(n_cols, n_rows):
+            if (fit_result := self._good_fits_by_roi_name.get(binner.get_roi_name(col, row))) is None:
+                continue
```
ashleymeigh2, line 645: if fit_result is None:
    LOG.debug(f"Missing fit for ROI {binner.get_roi_name(col, row)}")

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+
+        n_cols, n_rows = binner.lengths()
+        for col, row in np.ndindex(n_cols, n_rows):
+            param_value = map_array[row, col]
+            if np.isfinite(param_value):
+                x_start = binner.left_indexes[col]
+                x_end = min(x_start + step, img_width)
+                y_start = binner.top_indexes[row]
+                y_end = min(y_start + step, img_height)
+                full_map[y_start:y_end, x_start:x_end] = np.float32(param_value)
```
ashleymeigh2, line 723: If bins overlap, later values might overwrite earlier ones

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+
+        strategies = {
+            ColourRangeMode.IQR: self._colour_levels_iqr,
+            ColourRangeMode.MAD: self._colour_levels_mad,
+            ColourRangeMode.PERCENTILE: self.colour_levels_percentile,
+        }
+        lower, upper = strategies[mode](finite_values)
+
+        if lower >= upper:
+            return float(finite_values.min()), float(finite_values.max())
```
ashleymeigh2, line 674: When lower >= upper (identical values / zero IQR), this falls back to full range.
This is safe, but may still produce a flat colormap.

mantidimaging/gui/windows/spectrum_viewer/model.py
```diff
@@ -586,8 +590,145 @@ def fit_single_region(
...
+        """Return the 2nd-98th percentile range of the given finite values."""
+        return float(np.percentile(finite_values, 2)), float(np.percentile(finite_values, 98))
+
+    @staticmethod
+    def _colour_levels_mad(finite_values: np.ndarray) -> tuple[float, float]:
+        """Clips to median +- 3xMAD and falls back to full range when MAD = 0 to exclude outliers"""
+        median = np.median(finite_values)
+        mad = np.median(np.abs(finite_values - median))
+        if mad == 0:
+            return float(finite_values.min()), float(finite_values.max())
```
ashleymeigh2, line 688: This returns a wide range which may hurt visualisation.
Consider using a small range around the median instead to avoid a flat colormap.

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -799,6 +818,110 @@ def _clear_export_table(self) -> None:
...
+        """
+        Build colour parameter map (viridis) over the sample image for export
+
+        @param map_array: 2D array of parameter values for each binned ROI
+        @param binner: ROIBinner object containing the definitions of the binned ROIs
+        @return: 3D Colour array for export matching sample shape
+        """
+        colour_mode = ColourRangeMode(self.view.colour_range_mode)
+        levels = self.model.calculate_colour_levels(map_array, colour_mode)
+        full_map = self.model.build_full_sample_parameter_map(map_array, binner)
```
ashleymeigh2, line 875: This assumes map_array shape matches the binner dimensions. Might be worth validating before building the full map to avoid subtle misalignment issues.

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -799,6 +818,110 @@ def _clear_export_table(self) -> None:
...
+        @param sample_image: Background image to composite over. Pass None for map-only export.
+        @return: colour array
+        """
+        map_lower, map_upper = levels
+        fitted_pixel_mask = np.isfinite(full_map)
+        blank_background = np.zeros((*full_map.shape, 3))
+        background_rgb = self._normalise_background(sample_image) if sample_image is not None else blank_background
+
+        # only apply map to overlay over fitted pixels within levels exlcuidg outliers
+        if fitted_pixel_mask.any() and map_upper > map_lower:
```
ashleymeigh2, line 915: If map_upper <= map_lower the overlay is skipped entirely. This may result in no visible map without explanation consider handling or warning.

mantidimaging/gui/windows/spectrum_viewer/presenter.py
```diff
@@ -799,6 +818,110 @@ def _clear_export_table(self) -> None:
...
+    def _normalise_background(self, sample_image: np.ndarray) -> np.ndarray:
+        """
+        Normalise sample image using percentile levels
+
+        @param sample_image: the image to be normalised
+        @return: normalised colour image
+        """
+        sample = sample_image.astype(float)
+        finite_pixels = sample[np.isfinite(sample)]
+        lower, upper = self.model.colour_levels_percentile(finite_pixels)
```
ashleymeigh2, line 889: If finite_pixels is empty this will fail

mantidimaging/gui/windows/spectrum_viewer/view.py
```diff
@@ -191,19 +200,57 @@ def initial_setup(self) -> None:
         self.set_roi_properties()
         self.presenter.initial_roi_calc()
 
+    def _sync_export_background(self) -> None:
+        """Copy current image main spectrum view to the export widget"""
+        self._export_image_widget.update_image(self.spectrum_widget.image.image_item.image)
```
ashleymeigh2, line 205: Might be worth guarding against None or empty images to avoid downstream errors.

mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
```diff
@@ -24,21 +29,61 @@ def __init__(self, parent: QWidget | None) -> None:
...
+                           opacity: float = 0.5,
+                           levels: tuple[float, float] = (0.0, 1.0)) -> None:
+        """Display the parameter map overlay using viridis with the given opacity and colour levels."""
+        lower_level, upper_level = levels
+        self.param_overlay.setColorMap('viridis')
+        self.param_overlay.setOpacity(opacity)
+
+        # Align mapping with sample and scale to match binned resolution
+        transform = QTransform()
+        transform.translate(binner.left_indexes[0], binner.top_indexes[0])
```
ashleymeigh2, line 80: Assumes left_indexes and top_indexes are non-empty. If the binner is empty/invalid, this will raise an IndexError

mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
```diff
@@ -24,21 +29,61 @@ def __init__(self, parent: QWidget | None) -> None:
...
+        transform.scale(binner.step_size, binner.step_size)
+
+        self.param_overlay.setImage(map_array, autoLevels=False, levels=(lower_level, upper_level))
```
ashleymeigh2, line 83: Assumes map_array shape matches the binner grid. If they diverge, the overlay could be misaligned

mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
```diff
@@ -24,21 +29,61 @@ def __init__(self, parent: QWidget | None) -> None:
...
+        self.param_overlay.setTransform(transform)
+        self.param_overlay.show()
+
+        # connect histogram to parameter values and match colourmap
```
ashleymeigh2, line 87: Histogram is switched to the parameter overlay, not the background image.
Might be worth documenting, as this changes how users interpret intensity scaling.

samtygier-stfc:
I have marked some of the comments with a number:
1) These issues will be resolved by making `fit_all_regions()` return something like a dictionary with a numpy array for each parameter.
2) These are explicit checks for errors that I am not sure could be hit in practice. I don't think they should hold up merging. If we can find a way to hit them in MI, then they can be fixed as separate PRs. For the rest, we'll get a python exception if they ever happen and can decide if we need to handle or report a friendlier message to the user

mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
samtygier-stfc: 2

mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
samtygier-stfc: 1

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: 2

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: 1 or 2

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: 1

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: 2

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: 2

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: 1

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: 2

mantidimaging/gui/windows/spectrum_viewer/presenter.py
samtygier-stfc: 2

mantidimaging/gui/windows/spectrum_viewer/presenter.py
samtygier-stfc: 2

mantidimaging/gui/windows/spectrum_viewer/view.py
samtygier-stfc: 2

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: I think that would only be a problem if all the values in `finite_values` are the same, in which case a flat colour map is expected. Unless its actually causing an error somewhere.

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: I think `mad` would only be zero if more than halve of the values in the sample are identical.

I don't think there is a simple way to implement a small range around the median that's generally better than using the full range (min to max) and is not inventing a new ColourRangeMode.

samtygier-stfc:
For the colour range comments. I think it is ok if some of the modes give poor results in some cases. The user has multiple modes to choose from, so they should be able to find a good one.

Not blockers, but future ideas
I'd suggest adding a "Full range" mode, that just does min to max, in case there are any odd cases where the fancy methods are not great.

In
```
        if lower >= upper:
            return float(finite_values.min()), float(finite_values.max())
```
Is there any case where `lower == upper`, but `finite_values.min() != finite_values.max()`. In the obvious cases (all values in `finite_values` are equal), then all 4 are equal so calling `min()` and `max()` will give the same value.
So maybe better to treat `lower > upper` as bug in a colour mode and raise an error, and otherwise just `return lower, upper`.

mantidimaging/gui/windows/spectrum_viewer/model.py
samtygier-stfc: I think this ok for now. I think it would effectively something like nearest neighbour, but offset by step/2 and some oddness around the edges.

It might be possible to use `skimage.resize()` or `skimage.rescale()` to this more accurately. And give the user the option of nearest neighbour (order=0) or interpolated.

mantidimaging/gui/windows/spectrum_viewer/presenter.py
samtygier-stfc: 2

mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
samtygier-stfc: It would be nice to make the histogram and colorbar work well here. Though I am not sure what a good solution would look like. Having 2 of them would be cluttered. Not critical though so can be a future issue.

JackEAllen: I think instead of two histograms, we sync whatever colours and levels are used in the main image tab on opening the spectrum viewer across the rest of the view/tabs in the spectrum viewer

JackEAllen: So the image histogram is used for the image and the export one for export params

JackEAllen: I plan to write up an issue for this

samtygier-stfc (dismissed):
This all looks good an works well. I recommend merging as is, and picking up the following as future issues.

- Change `fit_all_regions()` return something like a dictionary with a numpy array for each parameter. This would simplify some of the code that uses the result, e.g. avoid need for `build_parameter_map()`
- Improve performance of updating the export table.
- Improve the histogram / colorbar. Make it easier for the user to adjust the colours of the sample and parameter overlay.
- Full range colour mode.
- Check if any of the noted error cases can be hit in practice. (what happens if you modify `fit_all_regions()` to set all the pixels to zero or one?)
- Reduce the logging
- Interpolation options
- Transparency and chi2 spinboxs could move in bigger steps when scrolling. transparency in 5% steps (QSpinBox.setSingleStep()), and adaptive for chi2, QDoubleSpinBox.setStepType()
- Tiff from "Mapped region only" wont open in imageJ

mantidimaging/gui/widgets/spectrum_widgets/export_image_widget.py
samtygier-stfc: That make sense. We can assume that the user adjusted the brightness to their taste on the first tab.

JackEAllen:
> Is there any case where `lower == upper`, but `finite_values.min() != finite_values.max()`. In the obvious cases (all values in `finite_values` are equal), then all 4 are equal so calling `min()` and `max()` will give the same value.

Hmm, I think the only edge case where `lower == upper` but `min != max` could occur is when the percentile range is too narrow relative to the distribution, so when say `np.percentile(finite_values, 0.5)` and `np.percentile(finite_values, 99.5)` happen to land on the same value which i guess could happen in dense/heavily quantised data where ~99% of the values are in the same integer, but a few outliers exist beyond percentile cutoffs or in the case of very sparse values where only 2-3 finite pixels exist, the -.5th and 99.5th percentile would "could" resolve to the same value even if values aren't identical. 
I guess these edge cases would be quite unlikely to hit and very unlikely in real data

samtygier-stfc (dismissed):

samtygier-stfc (dismissed):

samtygier-stfc (dismissed):
Approving again to appease github

ashleymeigh2 (dismissed):
looks good

samtygier-stfc (dismissed):

ashleymeigh2 (dismissed):
Looks good

ashleymeigh2 (dismissed):

samtygier-stfc (approved):

ashleymeigh2 (approved):

Issue #3086, opened 2026-04-09
System tests: Segmentation fault in TestGuiSpectrumViewer


samtygier-stfc:
### Summary

`mantidimaging/gui/test/gui_system_spectrum_test.py::TestGuiSpectrumViewer::test_change_roi_color Fatal Python error: Segmentation fault`

Backtrace is not very clear. Appears to be happening in the `setUp()` and maybe related to `pyqtgraph`'s `ViewBox`
```
Current thread 0x00007e5a2b44a740 (most recent call first):
  File "/home/sam/pixi/envs/mantidimaging-14710386930166075533/envs/dev/lib/python3.12/site-packages/pyqtgraph/graphicsItems/ViewBox/ViewBox.py", line ??? in <lambda>
  File "/home/sam/git/mantidimaging/mantidimaging/test_helpers/qt_test_helpers.py", line 20 in wait_until
  File "/home/sam/git/mantidimaging/mantidimaging/gui/test/gui_system_base.py", line 150 in _load_data_set
  File "/home/sam/pixi/envs/mantidimaging-14710386930166075533/envs/dev/lib/python3.12/unittest/mock.py", line 1396 in patched
  File "/home/sam/git/mantidimaging/mantidimaging/gui/test/gui_system_spectrum_test.py", line 30 in setUp
  File "/home/sam/pixi/envs/mantidimaging-14710386930166075533/envs/dev/lib/python3.12/unittest/case.py", line 586 in _callSetUp
  File "/home/sam/pixi/envs/mantidimaging-14710386930166075533/envs/dev/lib/python3.12/unittest/case.py", line 630 in run
...
```

GDB shows that it is in the destructors, so probably triggered during garbage collection.
```
#0  _PyEval_EvalFrameDefault (tstate=<optimised out>, frame=<optimised out>, throwflag=<optimised out>) at Python/bytecodes.c:1485
#1  0x00007ffff5e08193 in PyQtSlot::call(_object*, _object*) const () from /home/sam/pixi/envs/mantidimaging-14710386930166075533/envs/dev/lib/python3.12/site-packages/PyQt5/QtCore.abi3.so
#2  0x00007ffff5e08608 in PyQtSlot::invoke(void**, _object*, void*, bool) const () from /home/sam/pixi/envs/mantidimaging-14710386930166075533/envs/dev/lib/python3.12/site-packages/PyQt5/QtCore.abi3.so
[...]

PR #3087, opened 2026-04-09
system tests: delete main_window in tearDown()

samtygier-stfc:
## Issue related to #3086 

This does not fix the issue, but will make debugging simpler in some cases

### Description

This means the destructor happens in the `tearDown()` of the current test, rather than during the `setUp()` of the following test. There may be cases were this makes issues easier to find.

The `self.assertFalse(self.main_window.isVisible())` are not needed any more as there is a `wait_until()` in the top `tearDone()` 

### Developer Testing 

- ...

- [ ] ...

Not needed


ashleymeigh2 (approved):
Approved, explicitly checking that the reconstruction window closes is sensible. The changes look good.

PR #3093, opened 2026-04-13
Update Validation on Binnng ROI Changed

### LINKED ISSUE #3092: Spectrum Viewer: ROI Size and Step Warning Do Not Update on ROI Shape Change
### Summary

Bin Size and Bin Step warnings do not update on ROI change. If a warning exists and the ROI is reshaped into a state where the warning no longer applies, the warning does not correctly clear. 

### Steps To Reproduce

* Set bin size to 10, step size to 5
* Reshape ROI to a state that causes invalid step of bin size
* Reshape ROI to something that is acceptable such as 100x100
* Warning does not disappear unless bine or step size is changed

### Current Behaviour

Bin Size and Bin Step warnings do not update on ROI change. If a warning exists and the ROI is reshaped into a state where the warning no longer applies, the warning does not correctly clear. 

### Expected Behaviour

Bin Size and Bin Step warnings update on ROI change. If a warning exists and the ROI is reshaped into a state where the warning no longer applies, the warning is cleared. If no warning exists and the ROI is changed to no longer satisfy bin size or step values, warning should appear.  

### Context

Discovered during manual testing of a separate PR

JackEAllen:
Hookup  instead of  on  which already calls  so we can clear the warning of invalid bin size or step when the ROI is re-shaped to satisfy current values set. Call  in  to set value warning if ROI shape changed from valid state to invalid state based on bin size and step values after updating spinbox values since  uses a  meaning the  signal is suppressed dugin changes.

### Developer Testing 

- In spectrum viewer under Image sub-tab, set bin size to 10 and step size to 5
- A warning should appear
- Reshape ROI using drag and drop to 100x100, warning should disappear
- Reshape ROI to be slightly larger and no longer satisfy step size (say 113x113) and the warning should reapear on deselection
- Repeat the above steps using Bin size and step size Spinboxes        

- [ ] Warning for step bin and step size updates on ROI change in image view and via SpinBox selection

No need for release notes as minor fix that does not change underlying logic or user workflow

### DIFF
--- a/mantidimaging/gui/widgets/spectrum_widgets/roi_form_widget.py
+++ b/mantidimaging/gui/widgets/spectrum_widgets/roi_form_widget.py
@@ -63,7 +63,7 @@ class ROIFormWidget(BaseWidget):
 
         self.bin_step_spinBox.valueChanged.connect(self._binning_changed)
         self.bin_size_spinBox.valueChanged.connect(self._binning_changed)
-        self.roi_properties_widget.roi_changed.connect(self._update_bin_size_limit)
+        self.roi_properties_widget.roi_changed.connect(self._binning_changed)
 
     def update_bin_size_limit(self) -> None:
         self._update_bin_size_limit()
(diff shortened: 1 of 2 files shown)


RasmiaKulan (approved):

RasmiaKulan:
The changes work as expected where when the ROI is larger and does not satisfy the step size the warning appears otherwise it doesnt. Thank you for making these changes Jack! 😊

PR #3112, opened 2026-04-17
pre-commit: use same ruff version as pixi.toml

### LINKED ISSUE #3111: Ruff: difference in behaviour in pre-commit
### Summary

We use ruff for static checking. The rules to use are configured in `pyproject.toml`, we can run ruff either as a stand alone tool, or it will be run by pre-commit which is configured in `.pre-commit-config.yaml`.

One rule we use is `INP`, https://docs.astral.sh/ruff/rules/implicit-namespace-package/. This prevents bugs like #2442

But this is not needed for `docs`, `scripts` or other directories that are not intended to be packages. So we use the rule

```
[tool.ruff.lint.per-file-ignores]
"!mantidimaging/**.py" = ["INP"]
```
to ignore that rule, for any path that is not in `mantidimaging`

However this ignore rule seems be ignored in pre-commit. So when an edit is made to a file like `docs/conf.py` there is an `INP` error from pre-commit.

### Steps To Reproduce

To run with ruff
```
pixi shell -e lint
ruff check -v docs/conf.py
```

To run with pre-commit
```
pixi shell -e dev
pre-commit run --files docs/conf.py ruff
```
You can get a but more info by telling pre-commit to run ruff in verbose mode. In `.pre-commit-config.yaml` add an `args`
```
    hooks:
      - id: ruff
        args: ['-v']
```

### Current Behaviour

Fine in stand alone ruff. You can see a debug message `Adding per-file ignores for "/home/sam/git/mantidimaging/docs/conf.py" due to ...`

In pre-commit, the `per-file-ignores` does not seem to be considered.

### Notes

Ruff in pre-commit is reading `pyproject.toml`. If you add a rule there, e.g. `"T"`, then run
`pre-commit run --all-files` or `ruff check -v ` you will see both are using the new rule. So this only effects `per-file-ignores`

### Context

Current main.

samtygier-stfc:
Solves issue with ruff per-file-ignores setting not being used in pre-commit

### Description

In pixi.toml we are using ruff 0.3.7, in pre-commit we had 0.3.3.

We are using the `per-file-ignores` rule with a negative match, which was introduced in [ruff 0.3.6](https://github.com/astral-sh/ruff/releases/tag/v0.3.6)

### Developer Testing 

- Run `pre-commit run --files docs/conf.py ruff`

- [ ] `pre-commit run --files docs/conf.py ruff` does not give the `INP` error (you might need to be in pixi shell, see bug report)

Not needed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
         language: system
         types: [python]
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.3
+    rev: v0.3.7
     hooks:
       - id: ruff
   - repo: https://github.com/pre-commit/mirrors-prettier


ashleymeigh2 (approved):
Ran pre-commit run --files docs/conf.py ruff with no issue

PR #3140, opened 2026-04-30
Prevent IndexError in spectrum viewer when dataset contains no NaNs

### LINKED ISSUE #3127: Spectrum Viewer: Datasets with NaNs Can Trigger Index Error
### Summary

Loading a ToF dataset that contains NaNs or applying operations to a dataset to remove NaNs can sometimes produce an index error in the spectrum viewer. 

This could be possibly be caused when NaNs are removed ,`np.argwhere(np.isnan())` will return an empty or the array was empty and called before stack selection resulting in an index error on `nanInds[0,0]` within `spectrum_calculation_setup()`

I can't reproduce reliably as the error returns inconsistently for me on ADA.

### Current Behaviour

Index error produced on NaN removal to a stack. 

### Expected Behaviour

The Spectrum Viewer should be resilient to external changes while open. 

### Context

Discovered during release testing

### Failure Logs

```python
2026-04-23 21:12:42,939 [mantidimaging.gui.windows.main.view:L672] ERROR: Traceback (most recent call last):
  File "/home/ja1106423/.conda/envs/mantidimagingrelease/lib/python3.12/site-packages/mantidimaging/gui/windows/spectrum_viewer/presenter.py", line 327, in handle_roi_moved
    chunk_start, chunk_end = self.spectrum_calculation_setup()
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ja1106423/.conda/envs/mantidimagingrelease/lib/python3.12/site-packages/mantidimaging/gui/windows/spectrum_viewer/presenter.py", line 313, in spectrum_calculation_setup
    chunk_start = int(nanInds[0, 0])
                      ~~~~~~~^^^^^^
IndexError: index 0 is out of bounds for axis 0 with size 0

2026-04-23 21:12:42,939 [mantidimaging.gui.mvp_base.view:L46] ERROR: show_error_dialog(): Uncaught exception IndexError: index 0 is out of bounds for axis 0 with size 0

```

### Screenshot(s)

Image

ashleymeigh2:
### Description

Fixes an IndexError in the Spectrum Viewer caused by attempting to access elements of an empty array when no NaNs are present in the dataset.

Previously, np.argwhere(np.isnan(...)) could return an empty array, and the code would still attempt to access nanInds[0, 0], leading to a crash.

Changes

- Replaced np.argwhere with np.flatnonzero for simpler and safer 1D indexing
- Added a guard for empty or None datasets
- Explicitly handle the case where no NaNs are present by returning the full data range
- Fixed chunk boundary calculation to avoid off-by-one errors
- Removed reliance on ambiguous sentinel values (e.g. -1)
-

### DIFF
--- a/mantidimaging/gui/windows/spectrum_viewer/presenter.py
+++ b/mantidimaging/gui/windows/spectrum_viewer/presenter.py
@@ -323,21 +323,18 @@ class SpectrumViewerWindowPresenter(BasePresenter):
         queue = self.roi_to_process_queue
         mask = self.image_nan_mask_dict
         roi_name = list(queue.keys())[0]
+        data = mask[roi_name]
         chunk_size = 100
-        if chunk_size > 0:
-            nanInds = np.argwhere(np.isnan(mask[roi_name]))
-            if nanInds.size == 0:
-                # No NaNs found, return safe defaults
-                chunk_start, chunk_end = 0, -1
-            else:
-                chunk_start = int(nanInds[0, 0])
-                if len(nanInds) > chunk_size:
-                    chunk_end = int(nanInds[chunk_size, 0])
-                else:
-                    chunk_end = int(nanInds[-1, 0]) + 1
+        if data is None or data.size == 0:
+            return 0, 0
+        nan_positions = np.flatnonzero(np.isnan(data))
+        if nan_positions.size == 0:
+            return 0, data.shape[0]
+        chunk_start = int(nan_positions[0])
+        if nan_positions.size >= chunk_size:
+            chunk_end = int(nan_positions[chunk_size - 1]) + 1
         else:
-            chunk_start, chunk_end = (0, -1)
-
+            chunk_end = int(nan_positions[-1]) + 1
         return chunk_start, chunk_end
 
     def handle_roi_moved(self) -> None:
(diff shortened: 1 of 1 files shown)


RasmiaKulan:
This looks good - the added guard for empty NaN indices resolves the IndexError when no NaNs are present, and switching to flatnonzero simplifies the indexing.

A couple of non-blocking questions:

- Is data always 1D? Since flatnonzero flattens the array, behaviour could differ for multi-dimensional inputs.
- The chunk range is based on NaN positions rather than a fixed window size - is that intentional?

Otherwise, happy with this change 😊

ashleymeigh2:
> This looks good - the added guard for empty NaN indices resolves the IndexError when no NaNs are present, and switching to flatnonzero simplifies the indexing.
> 
> A couple of non-blocking questions:
> 
> * Is data always 1D? Since flatnonzero flattens the array, behaviour could differ for multi-dimensional inputs.
> * The chunk range is based on NaN positions rather than a fixed window size - is that intentional?
> 
> Otherwise, happy with this change 😊

the first point data is expected to be 1D in this context (it’s effectively a flattened spectrum along the ToF axis). so using flatnonzero shouldn’t change behaviour. If that assumption ever changes and multi-dimensional data is passed through then you’re right. it would be worth revisiting this to either enforce shape or handle it explicitly.

On the chunking behaviour, yes that’s intentional. The goal here is to process regions around NaNs rather than a fixed window since the NaNs define where recalculation is needed. this keeps the work focused on the affected areas instead of arbitrarily chunking the whole dataset.

Happy to add a shape assertion or comment if you think that would make the intent clearer.

RasmiaKulan (approved):

PR #3145, opened 2026-05-08
Operations: Fix Crop Coordinates ROI not updating when switching between stacks

### LINKED ISSUE #3115: Operations: Crop Coordinates ROI does Not Rescale Nor is AutoCrop Updated When Switching between Stacks
### Summary

If more than one dataset is loaded into Mantid imaging of different sizes (x, y) to highlight the issue where, switching from the larger dataset to a smaller one in the crop coordinates stack selector,  where the ROI is initially larger at first than the FoV of the smaller dataset (x, y) is not rescaled resulting in the ROI being larger than the sample FoV and causing an error.

Additionally, when switching between stacks, autocrop does not update per stack selection when opening the ROI placement editor. 

### Steps To Reproduce

**ROI Larger than FoV steps:**
* Load two datasets, one with a larger (x, y) axis than the other.
* Open the Crop coordinates operation and select the larger stack
* Resize the ROI to be larger than the smaller stack and click okay, but don't apply the change
* Using the stack selector, select the smaller sample
* Open up the ROI editor view
* The ROI should be larger than the FoV of the imagestack and an error should be returned

Auto-crop does not re-activate when switching between stacks when multiple datasets are loaded
* Load two datasets, one with a larger (x, y) axis than the other.
* Open the Crop coordinates operation and open ROI editor - autocrop will work as expected
* Using the stack selector, select the smaller sample
* Open up the ROI editor view and notice autocrop does not update correctly

### Current Behaviour

* The crop ROI can become larger than the sample FoV on sample stack change when the (x, y) dimensions differ across image stacks.
* Autocrop does not update when switching between different image stacks using the stack selector and editing ROI. 

### Expected Behaviour

* The ROI should be rescaled on stack selection to never be larger than the FoV without affecting Autocrop.
* Autocrop should update on sample stack selection change when opening the ROI crop area selector 

### Context

Discovered during manual testing #3079

RasmiaKulan:
### Description

- Added validation within `set_stacks` to refresh the ROI field using the current stack if the currently selected filter is Crop Coordinates, and the ROI field exists.
- Added a test for  `roi_field_called_with_auto_bounding_box_on_different_stacks`

### Developer Testing 

- I have verified the test `roi_field_called_with_auto_bounding_box_on_different_stacks` passes locally. Tested using the debugger as well to check if the variables were returning expected results.

#### ROI within FoV steps:
- [ ] Load two datasets, one with a larger (x, y) axis than the other.
- [ ] Open the Crop coordinates operation and select the larger stack
- [ ] Resize the ROI to be larger than the smaller stack and click okay, but don't apply the change
- [ ] Using the stack selector, select the smaller sample
- [ ] Open up the ROI editor view
- [ ] The ROI should be within the FoV of the imagestack 

#### Auto-crop is re-activated when switching between stacks when multiple datasets are loaded:
- [ ] Load two datasets, one with a larger (x, y) axis than the other.
- [ ] Open the Crop coordinates operation and open ROI editor - autocrop will work as expected
- [ ] Using the stack selector, select the smaller sample
- [ ] Open up the ROI editor view and confirm that the autocrop updates correctly

### DIFF
--- a/mantidimaging/gui/windows/operations/test/presenter_test.py
+++ b/mantidimaging/gui/windows/operations/test/presenter_test.py
@@ -462,6 +462,29 @@ class FiltersWindowPresenterTest(unittest.TestCase):
         self.presenter.init_roi_field(mock_roi_field)
         assert_called_once_with(mock_roi_field.setText, "0, 0, 100, 100")
 
+    @mock.patch("mantidimaging.gui.windows.operations.presenter.get_bounding_box")
+    def test_init_roi_field_called_with_auto_bounding_box_on_different_stacks(self, mock_bounding_box):
+        mock_roi_field = mock.Mock()
+
+        first_stack = mock.Mock()
+        first_stack.data = np.ones((2, 201, 201))
+
+        second_stack = mock.Mock()
+        second_stack.data = np.ones((2, 143, 140))
+
+        mock_bounding_box.side_effect = [SensibleROI(0, 0, 100, 100), SensibleROI(0, 0, 80, 70)]
+
+        self.presenter.stack = first_stack
+        self.presenter.init_roi_field(mock_roi_field)
+
+        self.presenter.stack = second_stack
+        self.presenter.init_roi_field(mock_roi_field)
+
+        expected_calls = [mock.call("0, 0, 100, 100"), mock.call("0, 0, 80, 70")]
+        mock_roi_field.setText.assert_has_calls(expected_calls)
+
+        assert mock_bounding_box.call_count == 2
+
     @mock.patch('mantidimaging.gui.windows.operations.presenter.FiltersWindowPresenter._do_apply_filter_sync')
     def test_negative_values_found_in_twelve_or_less_ranges(self, do_apply_filter_sync_mock):
         self.presenter.view.safeApply.isChecked.return_value = False
--- a/mantidimaging/gui/windows/operations/test/presenter_test.py
+++ b/mantidimaging/gui/windows/operations/test/presenter_test.py
@@ -17,8 +17,8 @@ from mantidimaging.core.operation_history.const import OPERATION_HISTORY, OPERAT
 from mantidimaging.core.utility.sensible_roi import SensibleROI
 from mantidimaging.gui.windows.main import MainWindowView
 from mantidimaging.gui.windows.operations import FiltersWindowPresenter
-from mantidimaging.gui.windows.operations.presenter import REPEAT_FLAT_FIELDING_MSG, FLAT_FIELDING, _find_nan_change, \
-    _group_consecutive_values, FLAT_FIELD_REGION
+from mantidimaging.gui.windows.operations.presenter import CROP_COORDINATES, REPEAT_FLAT_FIELDING_MSG, \
+    FLAT_FIELDING ,_find_nan_change, _group_consecutive_values, FLAT_FIELD_REGION
 from mantidimaging.test_helpers.unit_test_helper import assert_called_once_with, generate_images
 from mantidimaging.core.data import ImageStack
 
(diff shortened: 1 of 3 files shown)


mantidimaging/gui/windows/operations/test/presenter_test.py
```diff
@@ -462,6 +462,29 @@ def test_init_roi_field_called_with_auto_bounding_box(self, mock_bounding_box):
...
 
+    @mock.patch("mantidimaging.gui.windows.operations.presenter.get_bounding_box")
+    def test_init_roi_field_called_with_auto_bounding_box_on_different_stacks(self, mock_bounding_box):
+        mock_roi_field = mock.Mock()
+
+        first_stack = mock.Mock()
+        first_stack.data = np.ones((2, 201, 201))
+
+        second_stack = mock.Mock()
+        second_stack.data = np.ones((2, 143, 140))
```
JackEAllen: Are the stacks used in this test if we are mocking `get_bounding_box`?
 I think that when the `init_roi_field` runs, it does not inspect `stack.data`, it just uses the mocked return values via the `side_effect` so you may be abel to remove them so long as you keep the distinct stack assignment and init calls by having two:
```Python
self.presenter.stack = mock.Mock()
self.presenter.init_roi_field(mock_roi_field)
```

where assigning one stack object to the `presenter.stack` and calling `init_roi_field` performed twice may be enough to emulate a stack change.

JackEAllen (changes requested):
Thank you for adding a fix. Functionally this works well and resolves the issue 🥳  
I do have a small question about the test I'd like to get your opinion on, but otherwise happy with the change.

We can explore improving the extendability of updating fields on stack change for specific operations such is the case with this fix separately as part of #3152

mantidimaging/gui/windows/operations/test/presenter_test.py
```diff
@@ -462,6 +462,24 @@ def test_init_roi_field_called_with_auto_bounding_box(self, mock_bounding_box):
...
+
+        stack = mock.Mock(
+            name="test_stack",
+            shape=(2, 201, 201),
+            num_images=2,
+            num_sinograms=201,
+        )
+        with mock.patch("mantidimaging.gui.windows.operations.presenter.BlockQtSignals"):
+            self.presenter.set_stack(stack)
+
```
JackEAllen: The main assertion here looks good, but do we need the mock attributes here? 
They are set, but don't seem to be asserted anywhere after being passed to the mock unless i've missed something.
As `do_update_previews` is not mocked, I wonder if this could silently succeed or fail for unrelated reasons. i'd be tempted to add it (`    @mock.patch('mantidimaging.gui.windows.operations.presenter.FiltersWindowPresenter.do_update_previews')`) as an additional patch and we could then simplify to:
```python
        stack = mock.Mock(name="test_stack")
        self.presenter.set_stack(stack)
```

JackEAllen (changes requested):

mantidimaging/gui/windows/operations/test/presenter_test.py
RasmiaKulan: @JackEAllen I still needed to keep:

```python
with mock.patch("mantidimaging.gui.windows.operations.presenter.BlockQtSignals")
```

because `self.view` in the test is a `MagicMock` rather than a real Qt `QObject`. Without the patch, `BlockQtSignals` runs its real assertion:

```python
assert isinstance(obj, QObject)
```

which causes the test to fail before reaching the behaviour under test.

I also still needed to specify `num_images` on the stack mock because `set_stack()` accesses `self.max_preview_image_idx`, which evaluates:

```python
self.stack.num_images - 1
```

If `num_images` is left as the default auto-generated `Mock`, this results in:

```python
TypeError: unsupported operand type(s) for -: 'Mock' and 'int'
```

Since `do_update_previews` is now mocked, the additional attributes like `shape` and `num_sinograms` are no longer required, so the test can still be simplified to only the minimal attributes actually used by `set_stack()`.

JackEAllen: Sorry for being so slow to get back to you on this. I think we can possibly reduce the mock object attributes and have the test still pass from playing around locally to remove the unused attributes needed for `max_preview_image_idx` to compute `self.stack.num_images -1`, so:
```python
        stack = mock.Mock(
            name="test_stack",
            shape=(2, 201, 201),
            num_images=2,
            num_sinograms=201,
        )
```

Could become: 

```python
stack = mock.Mock(num_images=2)
```

JackEAllen (approved):
Looks Good! Happy to approve! This functionally resolves the issue of the crop-coordinate operation updating on stack change

PR #3156, opened 2026-05-18
Reconstruction Window: Remove maximum projection angle field

### LINKED ISSUE #1851: Reconstruction: Disable Maximum Projection Angle if Dataset Loaded with Angles
### Desired Behaviour

Disable/grey out the maximum Projection Angle in the reconstruction window panel if angle data is loaded from log files.

### Current Behaviour
Currently the maximum Projection Angle in the reconstruction window panel is enabled and can be changed if a dataset is loaded with log files containing angle data. This value is ignored if angle data is loaded and should be disabled/greyed out if angle data is available. 

Screenshot 2023-06-29 at 11 22 48

RasmiaKulan:
### Description

Removed the `Maximum projection angle` field from the reconstruction window along with any of its references. This is not required anymore as now there is no case where the value is used. If there is a geometry that is used instead of the value, if there is not a geometry then recon won't work.

### Developer Testing 

- I have manually tested with datasets with and without geometry to see if removing `Maximum projection angle` had an impact  on them in the reconstruction window.

- [ ] Open Mantid Imaging and navigate to the reconstruction window
- [ ] Confirm that the reconstruction window does not have the `Maximum projection angle` field.

### DIFF
--- a/mantidimaging/gui/windows/recon/view.py
+++ b/mantidimaging/gui/windows/recon/view.py
@@ -68,7 +68,7 @@ class ReconstructWindowView(BaseMainWindowView):
 
     reconTab: QWidget
 
-    maxProjAngleSpinBox: QDoubleSpinBox
+    #maxProjAngleSpinBox: QDoubleSpinBox
     algorithmNameComboBox: QComboBox
     filterNameComboBox: QComboBox
     numIterSpinBox: QSpinBox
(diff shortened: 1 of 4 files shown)


mantidimaging/gui/windows/recon/view.py
```diff
@@ -419,9 +417,9 @@ def slope(self) -> float:
     def slope(self, value: float) -> None:
         self.resultSlopeSpinBox.setValue(value)
 
-    @property
-    def max_proj_angle(self) -> float:
-        return self.maxProjAngleSpinBox.value()
+    # @property
```
ashleymeigh2: Can this not be removed?

ashleymeigh2 (changes requested):

mantidimaging/gui/windows/recon/view.py
RasmiaKulan: It can be removed, I mustve forgotten it while testing 😅

ashleymeigh2 (approved):
Looks good

PR #3160, opened 2026-05-21
3007 distance units for geometry

### LINKED ISSUE #3007: Geometry: Distance units for geometry
### Current Behaviour
Currently MI runs reconstructions using the pixel size as the unit. I.e. the reconstruction algorithm considers the pixel size to be 1x1. A scaling to the attenuation coefficient to account for the actual pixel size is applied after reconstruction.
When using cone beam geometry, the distances between the source, sample and detector are needed. CIL expects the same units for these distances and the pixel size. Therefor currently the distances must be specified in units of the pixel size. This is not user friendly.

### Desired Behaviour

The user should be able to specify distances and pixel size in SI units, i.e. m or mm.

Note: Using 1x1 pixels for iterative algorithms is working well. Some parameters may need to be scaled if we set pixel sizes in SI units. So if SI units were used in the geometry object, it may be good to scale them back to pixel units when calling the reconstruction algorithm.

If the distances are entered in SI, but still held in pixel units internally, then it might be needed to some back and forth conversions when the user changes the pixel size. E.g. suppose the use has a geometry with a pixel size of 50 μm and source-sample distance of 200 mm (4000 pixels). If the user changes the pixel size to 40 μm, the source-sample distance should still be 200 mm (now 5000 pixels).

### Importance of Feature

It is currently not clear for a user how to input distances for geometry

ashleymeigh2:
## Issue
Closes #3007

### Description

Adds SI unit support for geometry source and detector distances.

Changes include:
- Replacing pixel-based UI distances with `mm` and `m`
- Refactoring distance conversion into a single helper
- Updating geometry creation/update flows to use SI values
- Removing pixel units from geometry selectors
- Preventing recursive UI updates during unit switching

### Developer Testing

- Verified geometry distances update correctly when switching between `mm` and `m`
- Verified geometry creation and updates work correctly

- [ ] Verify geometry distances can be entered in `mm` and `m`
- [ ] Verify unit switching updates values correctly
- [ ] Verify geometry creation and updates still work correctly

- No documentation changes required

### DIFF
--- a/mantidimaging/core/data/geometry.py
+++ b/mantidimaging/core/data/geometry.py
@@ -46,6 +46,9 @@ class Geometry(AcquisitionGeometry):
         """
         super().__init__()
 
+        self._source_position_mm: float = 0.0
+        self._detector_position_mm: float = 0.0
+
         if type == GeometryType.PARALLEL3D:
             parallel_3d = self.create_Parallel3D(*args, units=units, **kwargs)
             self.config = parallel_3d.config
@@ -59,6 +62,7 @@ class Geometry(AcquisitionGeometry):
 
         self.set_panel(num_pixels=num_pixels, pixel_size=pixel_size)
         self.set_angles(angles=angles, angle_unit=angle_unit)
+        self.set_source_detector_positions(source_position[1], detector_position[1])
 
     def set_geometry_from_cor_tilt(self, cor: ScalarCoR, tilt: float) -> None:
         """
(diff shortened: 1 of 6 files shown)


RasmiaKulan:
The `_convert_distance` helper being inside the presenter is fine for the current scope since it’s only used within the geometry window. However, since this is a generic mm to m conversion, it would be slightly better long-term to place it in a shared utility module (e.g. `unit_conversion` or similar) if there’s any expectation that other parts of the application might need consistent unit handling in the future. Not a blocker for this PR though.

The current unit tests still assume a fixed "mm" unit configuration for all geometry values. Given that this PR introduces dynamic mm/m unit handling and conversion logic in both the presenter and view, there are currently no tests covering unit switching or verifying conversion correctness across units (e.g. mm to m round trips). It may be worth adding a test for unit transitions to ensure consistency between displayed values, internal mm storage, and conversion behaviour.

RasmiaKulan (approved):

PR #3164, opened 2026-06-03
Platform Specific Inline Code Block Custom Extension

JackEAllen:
### Description

Allow for inline code blocks to use platform specific commands. This is to work around that fact that Sphinx does not directly support this. It is only possible via code blocks with [group-tab](https://sphinx-design.readthedocs.io/en/latest/tabs.html) via [sphinx-design](https://sphinx-design.readthedocs.io/en/latest/index.html), but not inline code snippets.

Create a custom extension and supporting JS to use local platform OS to map RST to a CSS class to handle visibility of an inline code block which can be implemented as:

```rst
1. Open the Command Palette :IsMac:⌘+Shift+P :IsWindows:Ctrl+Shift+P :IsLinux:Ctrl+Shift+P and search for "Python: Select Interpreter"
```

Where whichever class is visible is shown while those that do not meet the platform criterion are hidden.

### Developer Testing 

- Checked no sphinx build warnings are occur as a result of this implementation by replacing an inline code block with OS specific code blocks based on the example above for a given page and running `pixi run build-docs`
- Tested each code block is shown depending on the OS used by testing on various OS's and by swapping over maps within `docs/ext/platform_detection_rst_to_css_mapping._OS_ROLES` by checking build docs hosted locally with `pixi run serve-docs`

- [ ] Check no sphinx build warnings are occur as a result of this implementation by replacing an inline code block with OS specific code blocks based on the example above for a given page.
- [ ] Test each code block is shown depending on the OS used by testing on various OS's and by swapping over maps within `docs/ext/platform_detection_rst_to_css_mapping._OS_ROLES` - it may be good to test different OS if possible and resort to swapping the mapping if this is not possible for you. 

- [ ] Sphinx documentation has been updated

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,2 +1,11 @@
 div.sd-card-body {font-size: 120%}
 div.sd-card-title {font-size: 300%}
+
+/* OS-specific keyboard shortcut visibility */
+.shortcut-mac     { display: none; }
+.shortcut-windows { display: none; }
+.shortcut-linux   { display: none; }
+
+.os-mac     .shortcut-mac     { display: inline; }
+.os-windows .shortcut-windows { display: inline; }
+.os-linux   .shortcut-linux   { display: inline; }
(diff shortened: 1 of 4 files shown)


RasmiaKulan:
I can approve that these changes work as expected, thank you Jack! 😊

RasmiaKulan (approved):

PR #3166, opened 2026-06-04
Fix headless system-test matplotlib backend

### LINKED ISSUE #2926: System tests: Problems with offscreen qt when running headlessly
### Summary

When running system tests we generally don't want things draw to the screen. When running tests on a headless system we can't draw things to screen.

We previously used xvfb (X virtual frame buffer) on Linux, but there is no equivalent on windows. 

https://github.com/mantidproject/mantidimaging/pull/2804
Added flags to set `QT_QPA_PLATFORM`
`--run-system-tests: os.environ["QT_QPA_PLATFORM"] = "minimal"`
`--run-system-tests-show: os.environ["QT_QPA_PLATFORM"] = ""`
`minimal` stops the GUI from being drawn on screen. This works on linux and windows

https://github.com/mantidproject/mantidimaging/pull/2887
Fixed an issue where the the new geometry window needs matplotlib to use a 
Checks QT_QPA_PLATFORM in geometry/view.py. if it is offscreen then
`matplotlib.use('Agg')`
else:
`matplotlib.use('Qt5Agg')`

These seem to not be fully interacting properly in some cases.

If I run 
`pixi run test-system-cpu`
from VSCode on Ada, e.g. on linux with out a screen, I get many errors:
```
  File "/mnt/ceph/home/t1110869/pixi_test/mantidimaging/mantidimaging/gui/windows/geometry/view.py", line 39, in <module>
    matplotlib.use('Qt5Agg')
  File "/mnt/ceph/home/t1110869/pixi_test/mantidimaging/.pixi/envs/dev/lib/python3.12/site-packages/matplotlib/__init__.py", line 1284, in use
    plt.switch_backend(name)
  File "/mnt/ceph/home/t1110869/pixi_test/mantidimaging/.pixi/envs/dev/lib/python3.12/site-packages/matplotlib/pyplot.py", line 432, in switch_backend
    raise ImportError(
ImportError: Cannot load backend 'Qt5Agg' which requires the 'qt' interactive framework, as 'headless' is currently running
```

I can also reproduce locally on Linux if I do
`unset DISPLAY`

I can trigger on conda or pixi environments. We maybe did not see it before because on linux we were still using `xvfb` so even on headless machines qt5agg could start.

### Steps To Reproduce

On a linux conda dev env
```
unset DISPLAY
python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true --run-system-tests mantidimaging/gui/test/
```
Fails quickly with `ImportError: Cannot load backend 'Qt5Agg' which requires the 'qt' interactive framework, as 'headless' is currently running`

```
unset DISPLAY
QT_QPA_PLATFORM="offscreen" python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true --run-system-tests mantidimaging/gui/test/
```
fails when it hits a test with multiprocessing.

### Current Behaviour

Crashes

### Expected Behaviour

No crashes

ashleymeigh2:
### Description

Fixes headless system-test crashes by setting matplotlib backend during pytest collection (after test flags are applied), not at geometry view import time.

### Developer Testing 

Agg for QT_QPA_PLATFORM = offscreen or minimal
Qt5Agg otherwise

- [ ]  With DISPLAY unset + --run-system-tests, tests do not fail at import with Qt5Agg headless error
- [ ] offscreen/minimal selects Agg backend
s baselines](https://eyes.applitools.com/app/baselines) for more info) -->

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import os
 from collections import Counter
 
+import matplotlib
 import pytest
 from PyQt5 import QtCore
 from PyQt5.QtGui import QFont
@@ -30,16 +31,24 @@ skipped_tests = []
 
 
 def pytest_collection_modifyitems(config, items):
+    qt_platform = None
     if config.getoption("--run-system-tests"):
         allowed_markers.append(pytest.mark.system.mark)
-        os.environ["QT_QPA_PLATFORM"] = "minimal"
+        qt_platform = "offscreen"
     if config.getoption("--run-system-tests-show"):
         allowed_markers.append(pytest.mark.system.mark)
-        os.environ["QT_QPA_PLATFORM"] = ""
+        qt_platform = ""
     if config.getoption("--run-eyes-tests"):
         allowed_markers.append(pytest.mark.eyes.mark)
     if config.getoption("--run-unit-tests") or len(allowed_markers) == 0:
         allowed_markers.append(pytest.mark.unit.mark)
+    if qt_platform is not None:
+        os.environ["QT_QPA_PLATFORM"] = qt_platform
+    if os.environ.get("QT_QPA_PLATFORM") in ["offscreen", "minimal"]:
+        matplotlib.use("Agg")
+    else:
+        matplotlib.use("Qt5Agg")
+
     for item in items:
         if "gui_system" in item.nodeid:
             item.add_marker(pytest.mark.system)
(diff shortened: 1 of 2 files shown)


RasmiaKulan (approved):

RasmiaKulan:
I have tested these changes on Ada and windows and running `pixi run test-system-cpu` does not show `ImportError: Cannot load backend 'Qt5Agg' which requires the 'qt' interactive framework, as 'headless' is currently running`. Thank you for these changes! 😊

PR #3167, opened 2026-06-09
update pyright

### LINKED ISSUE #3165: Package Management: Update Pyright (Round 2)
### Current Behaviour

#3162 updated pyright to `1.1.407` but issues states to update to latest version which was 1.1.409 also mentioned in the PR at the time of development in `meta.yml` and `environment-dev.yml`, but missed updating `pixi.toml`. Update all to `1.1.410` (the latest version available as of today) including `pixi.toml`

Files to be updated:
* `meta.yml`
* `pixi.toml`
* `environment-dev.yml`

ashleymeigh2:
### Description

Updated pyright to version 1.1.410 across project package management files:

pixi.toml
environment-dev.yml
meta.yaml

In addition to the version bump, I also worked through compatibility fallout from newer Pyright checks by updating type usage in the GUI layer, including pyqtgraph import/type annotations and stricter return/callback typing where needed. I then adjusted a few type hints to satisfy both Pyright and mypy without changing runtime behavior, and refined tree-item typing in main window helpers so static checks pass while existing mocked tests remain compatible.

### Developer Testing 

Verified dependency pins were updated to pyright 1.1.410 in all three target files.
I have not run the full unit test suite locally for this dependency-only update.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -37,6 +37,7 @@ requirements:
     - cil=25.0.*
     - ccpi-regulariser=25.0.*
     - jenkspy=0.4.*
+    - pyright=1.1.410
     - pyqt=5.15.*
     - pyqtgraph=0.13.7
     - qt-material=2.14
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -37,7 +37,6 @@ requirements:
     - cil=25.0.*
     - ccpi-regulariser=25.0.*
     - jenkspy=0.4.*
-    - pyright=1.1.410
     - pyqt=5.15.*
     - pyqtgraph=0.13.7
     - qt-material=2.14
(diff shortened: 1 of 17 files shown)


RasmiaKulan:
Once this PR is merged we should add `- pyright=1.1.410` to `recipe.yaml` as well. Ref: https://github.com/mantidproject/mantidimaging/pull/3171

JackEAllen:
Apologies, I should have updated the issue descriptio to nlt state we need this in the `meta.yml`. I don't think we need pyright as a dependency in the `meta.yml` as pyright is a developer only dependency so we should only need to include where a developer environment is required such as `environment-dev.yml` and `pixi.toml`

JackEAllen:
@ashleymeigh2 Could you add a description to commits: `1545f78e4e0a9548d6aac4c07bcaa35e2c4bcefc` and `5959e69b55687288e673ca76438685f19f9dd251` to briefly add a little context to the changes made here.

PR #3171, opened 2026-06-10
Migrate to rattler-build

### LINKED ISSUE #3136: Package Management: Boa is Superseded, Replace with Rattler-Build
### Current Behaviour

boa is still used within Mantid Imaging for actions despite the project being [archived](https://github.com/mamba-org/boa), superseded by rattler-build.
Migrate to use [rattler-build](https://github.com/prefix-dev/rattler-build) which is improved performance and is actively maintained.

RasmiaKulan:
### Description

- Added a `recipe.yaml` package recipe for rattler-build.
- Replaced Boa/mambabuild with rattler-build for Conda package builds.
- Added a `recipe_config.toml` configuration file to define build channels.
- Updated the package publishing GitHub Action to:
    - create a rattler-build environment instead of a Boa environment.
    - build packages using rattler-build.
    - add an optional debug_build input for inspecting build artifacts.
    - use `--force` during upload to avoid failures when rerunning workflows.

Updated the Makefile to build Conda packages using rattler-build.
Updated Pixi tasks to use rattler-build and remove legacy channel configuration commands that were only required for mambabuild.

### Developer Testing 

- Built the package locally using: `rattler-build build -r conda/recipe.yaml --experimental --output-dir output/unstable  --config-file conda/recipe_config.toml`
- Verified package upload succeeds in CI.
- Verified rerunning the publish workflow no longer fails due to existing package artifacts.
- Verified changes made after the last release were visible on Ada justifying successful package build.

- [ ] Validate package builds successfully using rattler-build.
- [ ] Validate package is uploaded to the unstable Anaconda label.
- [ ] Validate changes made after the last release were visible on Ada justifying successful package build.

### DIFF
--- a/.github/actions/publish-package/action.yml
+++ b/.github/actions/publish-package/action.yml
@@ -15,7 +15,7 @@ inputs:
 
   debug_build:
     description: "List build output contents after build (for debugging)"
-    default: "true"
+    default: "false"
 
 description: Build conda package
 runs:
@@ -42,6 +42,6 @@ runs:
         # if the upload silently fails - check the token expiration. Conda can fail silently!
         rattler-build build --recipe $GITHUB_WORKSPACE/conda --experimental --config-file $GITHUB_WORKSPACE/conda/recipe_config.toml
         if [[ "${{ inputs.debug_build }}" == "true" ]]; then ls -R output; fi
-        anaconda -t ${{ inputs.token }} upload --user ${{ inputs.repository }} --label ${{ inputs.label }} output/**/*.conda |& tee upload.log
+        anaconda -t ${{ inputs.token }} upload --force --user ${{ inputs.repository }} --label ${{ inputs.label }} output/**/*.conda |& tee upload.log
         # Check that upload completed
         grep "Upload complete" upload.log
--- a/.github/actions/publish-package/action.yml
+++ b/.github/actions/publish-package/action.yml
@@ -13,6 +13,10 @@ inputs:
     description: "Anaconda API Token"
     required: true
 
+  debug_build:
+    description: "List build output contents after build (for debugging)"
+    default: "true"
+
 description: Build conda package
 runs:
   using: "composite"
@@ -24,7 +28,7 @@ runs:
         conda config --set always_yes yes --set changeps1 no
         # Install build requirements
         # We can't use the makefile target for this because the CONDA_ACTIVATE command is incompatible with GitHub Actions Windows runners
-        conda create -n build-env --yes boa anaconda-client
+        conda create -n build-env --yes rattler-build anaconda-client
         conda activate build-env
         # Configure the conda channels
         conda config --env $(cat environment.yml | sed -ne '/channels:/,/dependencies:/{//!p}' | grep '^  -' | sed 's/ - / --append channels /g' | tr -d '\n')
@@ -36,7 +40,8 @@ runs:
         conda activate build-env
         conda config --set anaconda_upload no
         # if the upload silently fails - check the token expiration. Conda can fail silently!
-        conda mambabuild $GITHUB_WORKSPACE/conda
-        anaconda -t ${{ inputs.token }} upload --user ${{ inputs.repository }} --label ${{ inputs.label }} ${CONDA_PREFIX}/conda-bld/*/mantidimaging*.tar.bz2 |& tee upload.log
+        rattler-build build --recipe $GITHUB_WORKSPACE/conda --experimental --config-file $GITHUB_WORKSPACE/conda/recipe_config.toml
+        if [[ "${{ inputs.debug_build }}" == "true" ]]; then ls -R output; fi
+        anaconda -t ${{ inputs.token }} upload --user ${{ inputs.repository }} --label ${{ inputs.label }} output/**/*.conda |& tee upload.log
         # Check that upload completed
         grep "Upload complete" upload.log
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ install-build-requirements:
 	conda run -n build-env conda config --env $(CHANNELS)
 
 build-conda-package: install-build-requirements
-	conda run -n build-env conda mambabuild conda --label unstable
+	rattler-build build -r conda/recipe.yaml --experimental --output-dir output/unstable
 
 build-conda-package-nightly: .remind-for-user .remind-for-anaconda-api install-build-requirements
 	conda run -n build-env conda-build conda $(AUTHENTICATION_PARAMS) --label nightly
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ install-build-requirements:
 	conda run -n build-env conda config --env $(CHANNELS)
 
 build-conda-package: install-build-requirements
-	rattler-build build -r conda/recipe.yaml --experimental --output-dir output/unstable
+	rattler-build build -r conda/recipe.yaml --experimental --output-dir output/unstable --config-file conda/recipe_config.toml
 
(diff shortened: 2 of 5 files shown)


.github/actions/publish-package/action.yml
```diff
@@ -23,8 +23,7 @@ runs:
       run: |
         conda config --set always_yes yes --set changeps1 no
         # Install build requirements
-        # We can't use the makefile target for this because the CONDA_ACTIVATE command is incompatible with GitHub Actions Windows runners
```
JackEAllen, line 26: We should keep this comment in as i believe this is still an issue

.github/actions/publish-package/action.yml
```diff
@@ -36,7 +35,8 @@ runs:
         conda activate build-env
         conda config --set anaconda_upload no
         # if the upload silently fails - check the token expiration. Conda can fail silently!
-        conda mambabuild $GITHUB_WORKSPACE/conda
-        anaconda -t ${{ inputs.token }} upload --user ${{ inputs.repository }} --label ${{ inputs.label }} ${CONDA_PREFIX}/conda-bld/*/mantidimaging*.tar.bz2 |& tee upload.log
+        rattler-build build --recipe $GITHUB_WORKSPACE/conda --experimental --channel conda-forge --channel ccpi --channel mantidimaging/label/unstable --channel algotom --channel https://software.repos.intel.com/python/conda/
```
JackEAllen: if we are going to rely on `--experimental` features to get `git.latest_tag` / `git.latest_tag_distance` from `recipe.yml`, we should keep a very close eye on it's stability and when this moves out of `--experimental`.  Would you be okay to keep a close eye on this?
The only way we'd be able to possibly get around this is to pre-compute them which I think you've already tried right?

.github/actions/publish-package/action.yml
```diff
@@ -36,7 +35,8 @@ runs:
...
+        ls -R output || true
```
JackEAllen: I don't think we will want to debug the output by default so I wonder if we should consider replacing with  something like: `if [[ "${{ inputs.debug_build }}" == "true" ]]; then ls -R output; fi` and adding a small block in `action.yml` like:
```yml
debug_build:
  description: "List build output contents after build (for debugging)"
  default: "false"
``` 
which we can set to True when we do want to debug the output. 

What are your thoughts on this? 

We'd also want to update the developer debugging documentation if we choose to do this.

.github/actions/publish-package/action.yml
RasmiaKulan: Yes I did try to pre-compute it and `--experimental` is needed to run `git.latest_tag / git.latest_tag_distance`. I will keep a close eye on this.

conda/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+      then: qt-gtk-platformtheme
+
+tests:
+  - python:
+      imports:
+        - mantidimaging
+      pip_check: false
+
+about:
+  license: GPL-3.0-only
```
samtygier-stfc: Should be `GPL-3.0-or-later`

conda/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+requirements:
+  build:
+    - python 3.12.*
+    - setuptools 72.*
+
+  host:
+    - python 3.12.*
+    - setuptools 72.*
+    - pip
```
samtygier-stfc, line 26: I don't think a `host` section is needed as MI does not do cross compiling.

Makefile
```diff
@@ -24,7 +24,7 @@ install-build-requirements:
 	conda run -n build-env conda config --env $(CHANNELS)
```
samtygier-stfc:
```
conda create -n build-env --yes rattler-build anaconda-client
```
the channels line (and the variable on line 8) can probably be remove if that's done through a config file now.

conda/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+  run:
+    - python 3.12.*
+    - pip
+    - astropy 7.1.*
+    - scipy 1.14.*
```
samtygier-stfc, line 32: For consistency, can the dependencies all have a `=`, `>=` or which ever they need. Then this section can be copy and pasted between the meta.yaml and here.

samtygier-stfc (changes requested):
Jack asked if I could have a quick look at this.

Makefile
```diff
@@ -24,7 +24,7 @@ install-build-requirements:
...
 
 build-conda-package: install-build-requirements
-	conda run -n build-env conda mambabuild conda --label unstable
+	rattler-build build -r conda/recipe.yaml --experimental --output-dir output/unstable --config-file conda/recipe_config.toml
```
samtygier-stfc: This needs to be
```
conda run -n build-env rattler-build build -r conda/recipe.yaml --experimental --output-dir output/unstable --config-file conda/recipe_config.toml
```

samtygier-stfc (changes requested):

samtygier-stfc:
With the requested makefile changes I can build and install locally on linux

conda/recipe.yaml
RasmiaKulan: Rattler throws incompatibility errors when versions are specified with operators, for instance:
```
Error:   × Failed to parse recipe
  ╰─▶ invalid value for 'item': '*' is incompatible with '=' operator'
    ╭─[C:\Programming\mantidimaging\conda\recipe.yaml:28:7]
 27 │   run:
 28 │     - python=3.12.*
    ·       ──────┬──────
    ·             ╰── invalid value for 'item': '*' is incompatible with '=' operator'
 29 │     - pip
    ╰────
```
After testing a non-existent python version `python=3.17.*` and receiving the following confirms that it takes the version as defined.
```
Error:   × Failed to resolve dependencies
  ╰─▶ Cannot solve the request because of: No candidates were found for python 3.17.*.
```

conda/recipe.yaml
RasmiaKulan: I tried removing `host`, but the build fails. Without a `host` environment, pip falls back to the system Python instead of an isolated build environment, resulting in the `FileNotFoundError`:
```
│ │ FileNotFoundError: [WinError 2] The system cannot find the file specified: 'C:\\Python310\\Scripts\\mantidimaging.exe' -> 'C:\\Python310\\Scripts\\mantidimaging.exe.deleteme'
 │ │ × error Script failed with status 1
 │ │ × error
 │ │ × error Script execution failed.
 │ │ × error
 │ │ × error   Work directory: C:\Programming\mantidimaging\output\unstable\bld\rattler-build_mantidimaging_1783350669\work
 │ │ × error   Prefix: C:\Programming\mantidimaging\output\unstable\bld\rattler-build_mantidimaging_1783350669\h_env
 │ │ × error   Build prefix: C:\Programming\mantidimaging\output\unstable\bld\rattler-build_mantidimaging_1783350669\build_env
 │ │ × error
 │ │ × error To run the script manually, use the following command:
 │ │ × error   cd "C:\\Programming\\mantidimaging\\output\\unstable\\bld\\rattler-build_mantidimaging_1783350669\\work" && ./conda_build.bat
 │ │ × error
 │ │ × error To run commands interactively in the build environment:
 │ │ × error   cd "C:\\Programming\\mantidimaging\\output\\unstable\\bld\\rattler-build_mantidimaging_1783350669\\work" && call build_env.bat
 │ │
 │ ╰─────────────────── (took 2 minutes)
 │
 ╰─────────────────── (took 3 minutes)
Error:   × Script failed to execute
```

samtygier-stfc (approved):
This looks good. A local build using the makefile on Linux works, and I can install and run the output package.

Github actions part looks good. I think it is easiest to commit it, and see if it works.

PR #3172, opened 2026-06-16
Fix docs publish empty commit on release branches

### LINKED ISSUE #2757: Github actions: doc build failure on release branch
### Summary

On https://github.com/mantidproject/mantidimaging/actions/runs/16467470532/job/46548277644 (from https://github.com/mantidproject/mantidimaging/pull/2735), the docs action workflow failed.

```
nothing to commit, working tree clean
Error: Process completed with exit code 1.
```

I think this is because `git rm -r -q -- ${dir_name}/[!0-9]* || true` line in `git rm -r -q -- ${dir_name}/[!0-9]* || true` is not doing quite the right thing on the release branch, so `if [ "$(git status --porcelain | wc -l)" -gt "0" ]` sees changes outside the `${dir_name}` directory, but the add/commit of `${dir_name}` has no changes.

In the more common case were there are changes to the docs, everything works fine. So this is not directly causing problems, other than having the workflow show as failed.

ashleymeigh2:
### Description

Fix docs publish empty commit on release branches
Stage the target docs directory before checking for changes in the publish-docs action, and only commit when that staged directory has a diff. This avoids release branch docs jobs failing when the gh-pages worktree has unrelated changes but the versioned docs output is unchanged.

### DIFF
--- a/.github/actions/publish-docs/action.yml
+++ b/.github/actions/publish-docs/action.yml
@@ -29,8 +29,8 @@ runs:
         mkdir -p ${dir_name}
         cp -a ../docs/build/html/* ${dir_name}
         md5sum ${dir_name}/index.html # if no index then there must have been a problem
-        if [ "$(git status --porcelain | wc -l)" -gt "0" ] ; then
           git add ${dir_name}
+          if ! git diff --cached --quiet -- ${dir_name} ; then
           git commit -m "Update documentation"
           git push
         else
--- a/.github/actions/publish-docs/action.yml
+++ b/.github/actions/publish-docs/action.yml
@@ -29,8 +29,8 @@ runs:
         mkdir -p ${dir_name}
         cp -a ../docs/build/html/* ${dir_name}
         md5sum ${dir_name}/index.html # if no index then there must have been a problem
-          git add ${dir_name}
-          if ! git diff --cached --quiet -- ${dir_name} ; then
+        git add ${dir_name}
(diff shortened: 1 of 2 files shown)


.github/actions/publish-docs/action.yml
```diff
@@ -29,8 +29,8 @@ runs:
         mkdir -p ${dir_name}
         cp -a ../docs/build/html/* ${dir_name}
         md5sum ${dir_name}/index.html # if no index then there must have been a problem
-        if [ "$(git status --porcelain | wc -l)" -gt "0" ] ; then
           git add ${dir_name}
+          if ! git diff --cached --quiet -- ${dir_name} ; then
           git commit -m "Update documentation"
```
samtygier-stfc, line 34: Could you fix the indentation here
```bash
        md5sum ${dir_name}/index.html # if no index then there must have been a problem
        git add ${dir_name}
        if ! git diff --cached --quiet -- ${dir_name} ; then
          git commit -m "Update documentation"
          git push
        else
          echo "No changes to publish"
        fi
```

samtygier-stfc (changes requested):
Changes is good. Hard to test, but we'll find out soon with the release coming.

Just a small tweak to the indentation needed.

samtygier-stfc (dismissed):
Thanks. Looks good now.

JackEAllen (approved):

PR #3177, opened 2026-06-17
3158 Resolve CoR Refinement Inflated Values and Plot CoR Values Over Sample Image

### LINKED ISSUE #3158: Reconstruction: COR Refinement with Extreme Angles Produce's Poor Results
### Summary

When manually refining the COR in the presence of non-zero tilt with extreme angles (larger than 1 degree), the refine window returns incorrect, inflated COR values. The tilt offset appears to be applied possibly twice - once when reading from the COR table, and again internally via `get_cor_at_slice_index()`, causing a possible issue of double-counting that worsens reconstruction quality.

### Steps To Reproduce

- Load the Tilt4 dataset (or any dataset with a significant title angle).
- Apply outlier removal and flat fielding.
- Create a parallel geometry.
- Run Correlate Projections to auto-find COR values
- Open the refine window and manually refine the COR for a Slice (e.g., 655)
- Observe the refined COR value and compare the reconstruction preview to the auto found result.

### Current Behaviour

Manual refinement produces a significantly higher COR than the auto-correlation result, and the reconstruction preview visibly degrades. For example, at slice 655 for the "Tile4" dataset, the auto-correlation finds a COR of 1288 (better), while manual refinement shifts it to 1346 (worse). Setting the tilt back to zero causes the refine window to return a correct value suggesting the bug is tilt-dependent.

it is likely that the root cause is in `_recon_cor_preview` where `center_value` is set to the COR at `self.slice_idx` (read from the COR table), but is then assigned to `geometry.cor` which expects the COR at slice 0. When `get_cor_at_slice_index(slice_idx)` is then called, the tilt offset is added a second time, double-counting it.

### Expected Behaviour

The refine window should determine the COR as if not tilt were present i.e. it should probably back-calculate and return the COR at slice 0, independent of tilt. Manual refinement should produce values consistent with, or an improvement on the auto-correlation result.

A possible fix: back-calculate the COR at slice 0 before assigning to `geometry.cor` in `_recon_cor_preview`.

### Context

- The bug is most apparent with extreme tilt angles. A typical dataset will have <1deg title which may have previously masked the issue. The Tilt4 dataset is a useful case for reproducing and verifying a fix.
- There is an underlying state management concern here: COR information exists in two places: the image stack geometry and the COR table, which makes synchronisation fragile and may contribute to this sort of bug. 
- Possible Feature to incorporate in bug resolution: Overlaying the table's COR positions as crosshairs on the reconstruction preview image would make it immediately visible whether the COR/tilt line correctly passes through the found points. This would aid debugging and hel verify any fix.

JackEAllen:
### Description

* Resolve CoR refinement inflated values by preventing double calculating of tilt to be applied and create appropriate tests with the intention of avoiding CoR inflation going forward.
* Allow manual CoR refinement dialog window to be moveable and resizeable.
* Plot CoR values over sample image to improve CoR value visibility.

image

### Developer Testing 

- Manually tested by comparing with `main` that CoR refinement is not inflated by applying the same steps to the same dataset and comparing the output and validating the failure of `test_recon_cor_preview_does_not_inflate_cor_with_tilt` if introduce in `main` branch.
- Checked CoR spot selection updates table of CoR row selection and vice versa.

- [ ] CoR values are no longer inflated by the double calculation of tilt when compared to `main` when performing manual refinement. You can check this via visual comparison with a large tilt, via debug statements and / or by adding `mantidimaging/gui/dialogs/cor_inspection/test/model_test.test_recon_cor_preview_does_not_inflate_cor_with_tilt()` to the main branch and validating a failure. 
- [ ] Check CoR spot selection updates table of CoR row selection and vice versa.

- [ ] Screenshot tests have been updated
  - [ ] **Before merge for developer:** Resolve the change on applitools by, creating a new baseline for test and verify that the tests pass.
  - [ ] **After merge for reviewer:** Go to [Applitools compare and merge page](https://eyes.applitools.com/app/merge/), select all changes and merge dev branch to default branch
  - [ ] **After merge for reviewer:** Verify that other PR's screenshot tests do not start to fail (See [Applitools baselines](https://eyes.applitools.com/app/baselines) for more info)

### DIFF
--- a/mantidimaging/gui/dialogs/cor_inspection/model.py
+++ b/mantidimaging/gui/dialogs/cor_inspection/model.py
@@ -84,7 +84,7 @@ class CORInspectionDialogModel:
     def _recon_cor_preview(self, image: ImageType) -> np.ndarray:
         assert (self.images.geometry is not None)
         assert self.proj_angles is not None
-        self.images.geometry.cor = ScalarCoR(self.cor(image))
+        self.images.geometry.set_cor_at_slice_index(self.slice_idx, ScalarCoR(self.cor(image)))
         return self.reconstructor.single_sino(self.images, self.slice_idx, self.recon_params)
 
     def _recon_iters_preview(self, image: ImageType) -> np.ndarray:
--- a/mantidimaging/gui/dialogs/cor_inspection/model.py
+++ b/mantidimaging/gui/dialogs/cor_inspection/model.py
@@ -82,10 +82,20 @@ class CORInspectionDialogModel:
             return self.centre_value + self.step
 
     def _recon_cor_preview(self, image: ImageType) -> np.ndarray:
-        assert (self.images.geometry is not None)
+        """
+        Reconstructs a preview of the rotation centre for the given image
+        Ensures geometry is preserved on exit, even if the reconstruction fails
+        """
+        assert self.images.geometry is not None
         assert self.proj_angles is not None
+
+        original = self.images.geometry.get_cor_at_slice_index(self.slice_idx)
         self.images.geometry.set_cor_at_slice_index(self.slice_idx, ScalarCoR(self.cor(image)))
-        return self.reconstructor.single_sino(self.images, self.slice_idx, self.recon_params)
+
+        try:
+            return self.reconstructor.single_sino(self.images, self.slice_idx, self.recon_params)
+        finally:
+            self.images.geometry.set_cor_at_slice_index(self.slice_idx, original)
 
     def _recon_iters_preview(self, image: ImageType) -> np.ndarray:
         assert self.proj_angles is not None
--- a/mantidimaging/gui/windows/recon/image_view.py
+++ b/mantidimaging/gui/windows/recon/image_view.py
@@ -20,7 +20,7 @@ class ReconImagesView(GraphicsLayoutWidget):
     def __init__(self, parent):
         super().__init__(parent)
         self.parent = parent
-        self._selected_cor_row = -1
+        self._selected_cor_row: int = -1  # Default to no row selected
         self._scatter_rows: list[int] = []
         self._scatter_cors: list[float] = []
         self._scatter_slice_indices: list[int] = []
@@ -147,24 +147,30 @@ class ReconImagesView(GraphicsLayoutWidget):
         self._scatter_rows = list(range(len(slice_indices)))
         self._refresh_cor_scatter()
 
-    def _make_spot(self, cor: float, slice_idx: int) -> dict:
+    def set_selected_cor_row(self, row: int) -> None:
+        self._selected_cor_row = row
+        self._refresh_cor_scatter()
+
+    def _make_spot(self, cor: float, slice_idx: int, row: int) -> dict:
         """
-        Create CoR spot
+        Create CoR spot and update its appearance based on whether it is selected or not
 
         :param cor: CoR value
         :param slice_idx: slice index
+        :param row: row index
         :return: dict of spot parameters
         """
+        selected = row == self._selected_cor_row
         return {
             'pos': (cor, slice_idx),
-            'pen': self._normal_pen,
-            'brush': self._normal_brush,
+            'pen': self._selected_pen if selected else self._normal_pen,
+            'brush': self._selected_brush if selected else self._normal_brush,
         }
 
     def _refresh_cor_scatter(self) -> None:
-        """Update scatter plot with current CoR"""
-        cor_table_data = zip(self._scatter_cors, self._scatter_slice_indices, strict=True)
-        spots = [self._make_spot(cor, slice_idx) for cor, slice_idx in cor_table_data]
+        """Update scatter plot with current CoR by creating a list of spots and setting them to the scatter plot item"""
+        cor_table_data = zip(self._scatter_cors, self._scatter_slice_indices, self._scatter_rows, strict=True)
+        spots = [self._make_spot(cor, slice_idx, row) for cor, slice_idx, row in cor_table_data]
         self.cor_scatter.setData(spots=spots, symbol='x', size=14, tip=None)
 
     def show_cor_line(self, tilt: Degrees, pos: float) -> None:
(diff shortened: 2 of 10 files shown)


JackEAllen:
Note for reviewer: Resolved Applitools changes introduced by #3156 so we will need to merge in the applitools branch once this PR is merged into `main`

mantidimaging/gui/dialogs/cor_inspection/model.py
```diff
@@ -84,7 +84,7 @@ def iterations(self, image: ImageType) -> float:
     def _recon_cor_preview(self, image: ImageType) -> np.ndarray:
         assert (self.images.geometry is not None)
         assert self.proj_angles is not None
-        self.images.geometry.cor = ScalarCoR(self.cor(image))
+        self.images.geometry.set_cor_at_slice_index(self.slice_idx, ScalarCoR(self.cor(image)))
         return self.reconstructor.single_sino(self.images, self.slice_idx, self.recon_params)
```
ashleymeigh2: set_cor_at_slice_index() mutates the Geometry object used by the preview reconstruction. I don't see an obvious restoration of the original geometry after rendering or when the dialog is cancelled. Can you confirm whether the geometry is copied elsewhere (e.g. inside single_sino() or the presenter), or whether we should restore the original CoR after generating the preview to avoid leaking preview state into the main reconstruction model?

mantidimaging/gui/windows/recon/view.py
```diff
@@ -334,6 +346,16 @@ def update_projection(self, image_data, preview_slice_index: int) -> None:
     def update_sinogram(self, image_data) -> None:
         self.image_view.update_sinogram(image_data)
 
+    def _update_cor_scatter(self) -> None:
+        """Updates the COR scatter points on the projection preview"""
+        points = self.cor_table_model._points
```
ashleymeigh2: The view is reaching into cor_table_model._points directly. Would it be better to expose these through a public accessor on the model to reduce coupling?

mantidimaging/gui/windows/recon/image_view.py
```diff
@@ -122,6 +141,56 @@ def hide_cor_line(self) -> None:
...
+        Updates the CoR scatter points on the projection preview and the tilt line if there is a tilt
+
+        :param slice_indices: list of slice indices for the CoR points
+        :param cors: list of CoR values for the CoR points
+        :param tilt: tilt value for the tilt line, if there is a tilt
+        """
+        self._tilt_value = tilt
+        self._scatter_cors = cors
+        self._scatter_slice_indices = slice_indices
+        self._scatter_rows = list(range(len(slice_indices)))
```
ashleymeigh2, line 155: Scatter points use their current table index as the identifier? Just wondering whether sorting/filtering the table in future could break the mapping.

mantidimaging/gui/windows/recon/view.py
```diff
@@ -386,6 +408,7 @@ def add_cor_table_row(self, row: int, slice_index: int, cor: float) -> None:
         """
         self.cor_table_model.appendNewRow(row, slice_index, cor)
         self.tableView.selectRow(row)
+        self._update_cor_scatter()
```
ashleymeigh2: appendNewRow() emits rowsInserted, which is already connected to _update_cor_scatter(). Is this explicit refresh still needed, or could we rely on the model signal to avoid an extra redraw?

ashleymeigh2 (changes requested):

mantidimaging/gui/dialogs/cor_inspection/model.py
JackEAllen: This is a good point. well catched! You are correct we don't actually revert geometry back after we set here which is a problem. I did wonder if we could easily alter the geometry in-place only on completion of cor inspection to avoid needing to implement logic to store the original core value and revert back to it, however this would not be trivial. Reason being is that every `BaseRecon.single_sino()` call reads from `images.geometry.get_cor_at_slice_index(slice_idx)` so if not set in-place, the recon previews will use the old COR value. because of this, we have two options, setting the cor value for the preview recon and immediately restoring or implementing a new attribute in `CORInspectionDialogModel` to hold the original value for a while to restore too if a user cancel of closes the CoR inspection dialog. I think option 1 is probably the most minimal implementation. I'll also add a test to check we correctly revert geometry on cancellation.

mantidimaging/gui/windows/recon/view.py
JackEAllen: As it stands, `self._update_cor_scatter()` is needed to ensure the table or CoRs selection remains in sync with the spot selection. The redraw in itself fortunately isn't expensive so we're okay to redraw here without worrying about it affecting responsiveness. 

`appendNewRow` internally calls `sort_points()` which re-orders the points and emits `layoutChanged`, but since `layoutChanged` doesn't connect to `updated_cor_scatter`, the scatter plot was never redrawn after sort including when we initially populate the table or CoRs. The explicit call to `self._update_cor_scatter` in `add_cor_table_row` essentially solved this by redrawing after `selectRow`, but thinking about this more, it makes much more sense to instead just add a signal directly to `layoutChanged` to call `self. _update_cor_scatter` so that whenever the signal fires such as when we `sort_points`, including inside `appendNewRow`, the scatter plot points should remain in sync with the table of CoRs row selection. 

Noticing that we still call `_update_cor_scatter` through `rowsInserted`, which from looking at the execution, fires before we sort, we can probably safely remove this and rely solely on a connection to `layoutChanged`.

mantidimaging/gui/windows/recon/image_view.py
JackEAllen: Valid concern, but should be fine if i'm moving `_update_cor_scatter` to be called on `layoutChanged`.  Would likely only be of concern if we wanted to apply filtering, but this isn't something we plan on implementing as of right now 🙂

ashleymeigh2:
Can you squish your commits, please?

JackEAllen:
> Can you squish your commits, please?

@ashleymeigh2  Which commits do you think need condensing? I think they are split well to describe the context of the changes made

ashleymeigh2 (approved):
Looks good

PR #3181, opened 2026-07-01
Operations:  Add Generic hook with polymorphic dispatch for on stack change

### LINKED ISSUE #3152: Operations: Generic hook with polymorphic dispatch for on stack change
### Current Behaviour

When switching stack in the Operation Window, operations such as Crop Coordinates do not update fields such the ROI and text input fields (temporary fix implemented as part of #3145) to match the newly selected stacks bounding box. This may be an issue for other operations. 

Currently, on stack changes such as the one resolved by #3145 are handled with filter-specific logic inside the presenter. This works, but causes scope creep as the presenter should not contain filter specific logic. This breaks MVP more than the below proposed solution would taking into consideration our existing architecture, and it not very extendable should we wish to do the same for future operations other than crop coordinates.

Create a central base hook which can be overwritten and explore which other operations may benefit from updating on-stack change, writing up a separate parent issue and sub issues for proposed refactors (if any).

### Desired Behaviour

Create a generic "stack changed" hook on filters in `base_filter.py`, and move operation specific behaviour such as the logic added as part of #3145 into their specific filters by adding a `on_stack_changed(...)` method to overwrite the `base_filter` hook. 

This will allow the presenter to call the generic hook with something like: 
```Python
self.model.selected_filter.on_stack_changed(self.model.filter_widget_kwargs, self.stack)
```

Add tests for new architecture in `presenter_test.py` to check `set_stack` calls the selected filter on stack change and separately add tests in the specific filters to test the hook updates any relevant fields or when no operation or widget is present, nothing happens.

Example base filter hook concept:
```Python
"staticmethod
def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
    """
    Optional hook for filters that need to refresh UI state when the selected stack changes.

    :param filter_widget_kwargs: The widgets registered for the active filter.
    :param stack: The newly selected stack, or None when selection is cleared.
    """
    return
```

### Importance of Feature

Creating a central hook to be polymorphically overwritten provides better seperation of concerns whenever an operation need to update on-stack changed as the presenter handles orchestration while filter-specific behaviour lives in each filter. The proposed architecture also provides improved extendability as new filters can react to stack changes by overriding one method and there is no need to touch the presenter each time making maintenance easier.  

### Additional Context

The proposed crop coordinates `on_stack_changed(filter_widget_kwargs, stack: ImageStack | None)` filter will need to import `get_bounding_box` to compute on stack change.

RasmiaKulan:
### Description

- Added a `on_stack_changed` hook for filters that require refreshing UI states when the selected stack changes.
- Refactored `presenter.py` to make use of the hook through `self.model.selected_filter.on_stack_changed(self.model.filter_widget_kwargs, self.stack)` instead of checking against `CROP_COORDINATES` filter and removed its declaration.
- Added `test_set_stack_calls_selected_filter_on_stack_changed` which replaces the previous one by moving the assertion from checking that the presenter directly initialises the ROI UI to verifying that it correctly delegates stack changes to the active filter through the `on_stack_changed` hook.
- Added `test_on_stack_changed_updates_roi_field` to verify `on_stack_changed` gets the bounding box from the stack and updates the ROI field text accordingly. (NOTE: The `crop_coords.crop_coords` path is intentional because the package and module have identical names, so Python resolves it by first importing the package and then accessing the module inside it.)
- Added `test_on_stack_changed_with_no_stack_or_no_widget_does_nothing` to verify the case when there is no stack or no widget present `on_stack_changed` doesn't do anything.

### Developer Testing 

- I added logs `within `set_stack`, `do_register_active_filter` and `_post_filter` during development to verify the functions were called as expected when crop coordinates operation was applied on different datasets.
- I have verified the tests I have added pass locally with:
   - `pytest mantidimaging/gui/windows/operations/test/presenter_test.py::FiltersWindowPresenterTest::test_set_stack_calls_selected_filter_on_stack_changed`
   - `pytest mantidimaging/core/operations/crop_coords/test/crop_coords_test.py::CropCoordsTest::test_on_stack_changed_updates_roi_field` 
   - `pytest mantidimaging/core/operations/crop_coords/test/crop_coords_test.py::CropCoordsTest::test_on_stack_changed_with_no_stack_or_no_widget_does_nothing`
   - Tested against the initial bug fix [3145](https://github.com/mantidproject/mantidimaging/pull/3145) and I can verify that the crop coordinates operation worked as expected.
  

- [ ] Newly added tests pass locally: 
   - [ ] `pytest mantidimaging/gui/windows/operations/test/presenter_test.py::FiltersWindowPresenterTest::test_set_stack_calls_selected_filter_on_stack_changed`
   - [ ] `pytest mantidimaging/core/operations/crop_coords/test/crop_coords_test.py::CropCoordsTest::test_on_stack_changed_updates_roi_field` 
   - [ ] `pytest mantidimaging/core/operations/crop_coords/test/crop_coords_test.py::CropCoordsTest::test_on_stack_changed_with_no_stack_or_no_widget_does_nothing`
- Following cases give expected results
 #### ROI within FoV steps:
- [ ] Load two datasets, one with a larger (x, y) axis than the other.
- [ ] Open the Crop coordinates operation and select the larger stack
- [ ] Resize the ROI to be larger than the smaller stack and click okay, but don't apply the change
- [ ] Using the stack selector, select the smaller sample
- [ ] Open up the ROI editor view
- [ ] The ROI should be within the FoV of the imagestack 

#### Auto-crop is re-activated when switching between stacks when multiple datasets are loaded:
- [ ] Load two datasets, one with a larger (x, y) axis than the other.
- [ ] Open the Crop coordinates operation and open ROI editor - autocrop will work as expected
- [ ] Using the stack selector, select the smaller sample
- [ ] Open up the ROI editor view and confirm that the autocrop updates correctly

### DIFF
--- a/docs/release_notes/next/dev-3152-polymorphic_stack_hook
+++ b/docs/release_notes/next/dev-3152-polymorphic_stack_hook
@@ -1,0 +1,1 @@
+#3152: Add a polymorphic on_stack_changed hook and refactor code accordingly
--- a/docs/release_notes/next/dev-3152-polymorphic_stack_hook
+++ b/docs/release_notes/next/dev-3152-polymorphic_stack_hook
@@ -1,1 +1,1 @@
-#3152: Add a polymorphic on_stack_changed hook and refactor code accordingly
+#3152: Adds support for filters to respond to stack changes, allowing filter settings to update automatically based on the selected data.
--- a/mantidimaging/core/operations/base_filter.py
+++ b/mantidimaging/core/operations/base_filter.py
@@ -114,6 +114,20 @@ class BaseFilter:
             raise ValueError(f"Selected stack for {msg} does not exist") from exc
         return stack
 
+    @staticmethod
+    def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
+        """
+        Optional hook for filters that need to refresh UI state when the selected stack changes.
+
+        :param filter_widget_kwargs: The widgets registered for the active filter.
+        :param stack: The newly selected stack, or None when selection is cleared.
+        """
+        widget = filter_widget_kwargs.get("stack_selector")
+        if widget is not None:
+            widget.set_current_stack(stack)
+
+        return
+
 
 def raise_not_implemented(function_name):
     raise NotImplementedError(f"Required method '{function_name}' not implemented for filter")
--- a/mantidimaging/core/operations/base_filter.py
+++ b/mantidimaging/core/operations/base_filter.py
@@ -115,17 +115,13 @@ class BaseFilter:
         return stack
 
     @staticmethod
-    def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
+    def on_stack_changed(filter_widget_kwargs: dict[str, QWidget], stack: ImageStack | None) -> None:
         """
         Optional hook for filters that need to refresh UI state when the selected stack changes.
 
         :param filter_widget_kwargs: The widgets registered for the active filter.
         :param stack: The newly selected stack, or None when selection is cleared.
         """
-        widget = filter_widget_kwargs.get("stack_selector")
-        if widget is not None:
-            widget.set_current_stack(stack)
-
         return
 
 
--- a/mantidimaging/core/operations/crop_coords/crop_coords.py
+++ b/mantidimaging/core/operations/crop_coords/crop_coords.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from functools import partial
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from mantidimaging import helper as h
 from mantidimaging.core.operations.base_filter import BaseFilter, FilterGroup
@@ -11,6 +11,7 @@ from mantidimaging.core.parallel import utility as pu
 from mantidimaging.core.utility.progress_reporting import Progress
 from mantidimaging.core.utility.sensible_roi import SensibleROI
 from mantidimaging.gui.utility.qt_helpers import Type
+from mantidimaging.core.fitting.bounding_box import get_bounding_box
 
 if TYPE_CHECKING:
     from mantidimaging.core.data import ImageStack
@@ -93,6 +94,19 @@ class CropCoordinatesFilter(BaseFilter):
             return partial(CropCoordinatesFilter.filter_func, region_of_interest=roi)
         except Exception as exc:
             raise ValueError(f"The provided ROI string is invalid! Error: {exc}") from exc
+
+    @staticmethod
+    def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
+
+        if stack is None:
+            return
+
+        roi_field = filter_widget_kwargs.get("roi_field")
+        if roi_field is None:
+            return
+
+        crop_roi = get_bounding_box(stack)
+        roi_field.setText(crop_roi.to_list_string())
 
     @staticmethod
     def group_name() -> FilterGroup:
--- a/mantidimaging/core/operations/crop_coords/crop_coords.py
+++ b/mantidimaging/core/operations/crop_coords/crop_coords.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from functools import partial
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
 from mantidimaging import helper as h
 from mantidimaging.core.operations.base_filter import BaseFilter, FilterGroup
@@ -15,7 +15,7 @@ from mantidimaging.core.fitting.bounding_box import get_bounding_box
 
 if TYPE_CHECKING:
     from mantidimaging.core.data import ImageStack
-    from PyQt5.QtWidgets import QLineEdit
+    from PyQt5.QtWidgets import QLineEdit, QWidget
 
 
 class CropCoordinatesFilter(BaseFilter):
@@ -96,7 +96,10 @@ class CropCoordinatesFilter(BaseFilter):
             raise ValueError(f"The provided ROI string is invalid! Error: {exc}") from exc
 
     @staticmethod
-    def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
+    def on_stack_changed(filter_widget_kwargs: dict[str, QWidget], stack: ImageStack | None) -> None:
+        """
+        Update the ROI field with the bounding box of the newly selected stack.
+        """
 
         if stack is None:
             return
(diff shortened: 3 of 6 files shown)


mantidimaging/core/operations/base_filter.py
```diff
@@ -114,6 +114,20 @@ def get_images_from_stack(widget: DatasetSelectorWidgetView, msg: str) -> ImageS
             raise ValueError(f"Selected stack for {msg} does not exist") from exc
         return stack
 
+    @staticmethod
+    def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
```
JackEAllen: For the base hook to be truly polymorphic, it's best not to have logic which assumes components exist such as the presence of a `stack_selector` and instead should just be a no-op pass like the example in #3152 where instead, any widget that has a `stack_selector` that wishes to apply logic `on_stack_changed` overwrites `on_stack_changed` in their specific operation. 
I'm not sure if any of the filters would actually return `"stack_selector"` from `register_gui ` so the logic within on_stack_changed may not actually be getting hit where the overide in `CropCoordinatesFilter` isn't calling super (I could be wrong here). 

So next steps here would just be refactoring on_stack_changed to remove logic and just be:
```
@staticmethod
def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
      """
      Optional hook for filters that need to refresh UI state when the selected stack changes.

      :param filter_widget_kwargs: The widgets registered for the active filter.
      :param stack: The newly selected stack, or None when selection is cleared.
      """
    return
```

and remove DatasetSelectorWidgetView import.

mantidimaging/core/operations/crop_coords/crop_coords.py
```diff
@@ -94,6 +95,19 @@ def execute_wrapper(roi_field: QLineEdit) -> partial:
         except Exception as exc:
             raise ValueError(f"The provided ROI string is invalid! Error: {exc}") from exc
 
+    @staticmethod
+    def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
+
```
JackEAllen, line 103: NITPICK: Maybe just a small docstring to describe that this overwrites the base filter so anyone new to the codebase not familiar with are architecture know's how this works if wishing to apply the same approach to other operations

mantidimaging/core/operations/crop_coords/test/crop_coords_test.py
```diff
@@ -53,6 +53,42 @@ def test_execute_wrapper_bad_roi_raises_valueerror(self):
...
+            mock_roi = mock.Mock()
+            mock_roi.to_list_string.return_value = "[1, 2, 3, 4]"
+            get_bounding_box_mock.return_value = mock_roi
+
+            CropCoordinatesFilter.on_stack_changed(filter_widget_kwargs, stack)
+
+            get_bounding_box_mock.assert_called_once_with(stack)
+            roi_mock.setText.assert_called_once_with("[1, 2, 3, 4]")
+
+    def test_on_stack_changed_with_no_stack_or_no_widget_does_nothing(self):
```
JackEAllen: I think this could be simplified to split the two different assertions going on here into separate tests: one to pass `stack=None` and assert `setText` is never called and another that patches `get_bounding_box_mock` to assert it's never reached.  

Maybe refactoring into two tests like this if you agree?
```Python
def test_on_stack_changed_with_no_stack_does_not_update_roi_field(self):
  roi_mock = mock.Mock()
  filter_widget_kwargs = {"roi_field": roi_mock}
  
  CropCoordinatesFilter.on_stack_changed(filter_widget_kwargs, None)
  
  roi_mock.setText.assert_not_called()

def test_on_stack_changed_with_no_roi_field_does_not_call_get_bounding_box(self):
  with mock.patch("mantidimaging.core.operations.crop_coords.crop_coords.get_bounding_box") as get_bounding_box_mock:
      CropCoordinatesFilter.on_stack_changed({}, mock.Mock())
      get_bounding_box_mock.assert_not_called()
```

mantidimaging/core/operations/crop_coords/crop_coords.py
```diff
@@ -94,6 +95,19 @@ def execute_wrapper(roi_field: QLineEdit) -> partial:
         except Exception as exc:
             raise ValueError(f"The provided ROI string is invalid! Error: {exc}") from exc
 
+    @staticmethod
+    def on_stack_changed(filter_widget_kwargs: dict[str, Any], stack: ImageStack | None) -> None:
```
JackEAllen: NITPICK: This is my bad for suggesting it in the original example, but i'm thinking that maybe using `Any` may be a bit too loose and probably better to be something closer to what we expect, maybe in `CropCoordinatesFilter`'s case, a `QWidget`  and also in the base filter and loosen again if we need to later on. if you agree?

docs/release_notes/next/dev-3152-polymorphic_stack_hook
```diff
@@ -0,0 +1 @@
+#3152: Add a polymorphic on_stack_changed hook and refactor code accordingly
```
JackEAllen: NITPICK: Maybe worth assuming no context here for users and describing very briefly why we are doing this to save someone needing to go the the issue to understand

JackEAllen (changes requested):
Thanks for implementing this! Definitely feels more robust and extendable for if we choose to later extend to other operations. Just a few minor suggestions relating to the polymophic nature of `on_stack_changed` not needing logic in the base filter and splitting up `test_on_stack_changed_with_no_stack_or_no_widget_does_nothing` to individual test cases then happy to merge in if you agree with my suggestions 🙂

mantidimaging/core/operations/crop_coords/test/crop_coords_test.py
```diff
@@ -53,6 +53,47 @@ def test_execute_wrapper_bad_roi_raises_valueerror(self):
...
+        """
+        roi_mock = mock.Mock()
+
+        filter_widget_kwargs = {"roi_field": roi_mock}
+        stack = mock.Mock()
+
+        with mock.patch(
+                "mantidimaging.core.operations.crop_coords.crop_coords.get_bounding_box") as get_bounding_box_mock:
+            mock_roi = mock.Mock()
+            mock_roi.to_list_string.return_value = "[1, 2, 3, 4]"
```
JackEAllen: NITPICK: from having a quick look at `SensibleROI.to_list_string`, I think `to_list_string` would return `"1, 2, 3, 4"` instead of `"[1, 2, 3, 4]"` so by wrapping in brackets, this may add a bit of confusion in that it's inconsistent  with what is actually going on, despite the test being technically correct as they are mocked.

JackEAllen (approved):
Functionally looks good from testing and code changes. Definitely cleans up this area in that we can now override stack change operations going forward. Happy to approve 👍 

I did notice on testing that we also have the same issue with the ROI Normalisation operation when selecting an air region. With this PR merged in, it should be an straightforward fix. I've written an issue for this here: #3191

PR #3184, opened 2026-07-10
[CI/CD] Developer Tools: Enable pre-commit hooks to use pixi environments & add feature tasks for checking license and release notes format

### LINKED ISSUE #2967: [CI/CD] Developer Tools: Enable pre-commit hooks to use pixi environments
### Current Behaviour

The `.pre-commit-config.yaml` file defines several hooks that depend on Python packages:
- yapf - uses remote repo `https://github.com/google/yapf`
- mypy - uses remote repo `hhttps://github.com/pre-commit/mirrors-mypy` with `additional_dependencies`
- ruff - uses remote repo `https://github.com/astral-sh/ruff-pre-commit`
- Local hooks - use `language: python` or `language: system`

These hooks currently:
1. Create isolated virtual environments managed by pre-commit
2. Don't use the pixi-managed environments that developers already have installed.
3. Can have version mismatches with the pixi environment (e.g. yapf, mypy, ruff versions)
4. Require downloading and installing packages separately form Pixi's dependency manangment 

### Desired Behaviour

Pre-commit hooks should use the pixi `lint` environment for all linting tools:
- Configure pre-commit to use pixi's environments instead of creating isolated ones.
- Ensure version consistency between pre-commit checks and CI/local development.
- Leverage pixi's existing package cache to avoid duplicate downloads.
- Make pre-commit hook execution faster by using already installed dependencies.
- Document Pre-commit setup with pixi.

- **File passing:** Some pre-commit hooks pass changed filenames to the tool. Pixi tasks currently run on all files (`$SOURCE_DIRS`). New tasks may be needed or exiting ones modified to accept file arguments or `pass_filenames: false` should be used and all files could be checked on each commit (the former is likely the better option)
- **Performance:** Running linters on all files instead just changed files will be slower. Consider pixi tasks that accept file arguments for pre-commit use.
- **Pixi Availability:** Pre-commit hooks reuire that pixi is installed and available in PATH. New developers will need to ensure they have installed pixi to ensure they don't experience errors.
- **CI consistency:** GitHub Actions workflows should use the same pixi tasks to ensure consistency between pre-commit and CI
- **Hook versioning:** The `.pre-commit-config.yaml` file won't have version pins for tools anymore (those are in `pixi.toml`) instead of being duplicated in `.pre-commit-config.yaml`

### Importance of Feature

- **Consistency:** Pre-commit checks will use exactly the same tool versions as CI and local development (defined in pixi.toml), eliminating "works on my machine" issues.
- **Performance:** Reusing pixi environments avoids duplicate package installations. Pre-commit currently downloads and installs yapf, mypy, ruff separately.
- **Maintenance:** Tool versions are managed in one place (pixi.toml) instead of being duplicated in .pre-commit-config.yaml.
- Developer Experience: Developers already use `pixi install` to setup their environment. Pre-commit will just work without additional setup (after initial `pre-commit install`
- **Alignment with migration goals:** This continues the shift from conda/mamba to pixi as the single source of truth for development environments.

### Additional Context

- Pre-commit is already installed in the `dev` environment: `pre-commit = "3.5.*"` in pixi.toml
- All required linting tools are in the `lint` environment: yapf, ruff, mypy, pyright
- The `lint` environment includes `lint-base` tasks that can be used directly
- Current pre-commit versions in .pre-commit-config.yaml:
    - yapf: `v0.43.0` (matches pixi: `yapf = "0.43.*"`)
    - mypy: `v1.17.1` (close to pixi: `mypy = "==1.17"`)
    - ruff: `v0.3.3` (close to pixi: `ruff = "0.3.7.*"`)
- The `SOURCE_DIRS` environment variable is defined in both `feature.lint.activation.env` and `feature.dev.activation.env`
- Consider creating file-specific pixi tasks for better pre-commit performance as part of this Issue

#### Possible Implementation Steps

Update .pre-commit-config.yaml to use language: system and pixi commands

1. Update `.pre-commit-config.yaml` to use `language: system` and pixi commands
2. Test pre-commit hooks locally with various scenarios:
    - Clean commit (should pass)
    - Commit with formatting issues (should fail/fix)
    - Commit with type errors (should fail)
3. Update documentation (README, contributing guide) to mention pixi requirement
4. Consider adding a helper script or make target: pixi run setup-pre-commit that runs pre-commit install
5. Test in CI to ensure consistency between pre-commit and GitHub Actions

RasmiaKulan:
### Description

- Updated `.pre-commit-config.yaml` with pre-commits using pixi's environment instead of creating their own isolated ones.
- Updated `pixi.toml` with the following tasks:
  - `check-license` to run `python -m CheckLicensesInFiles`.
  - `check-release-notes-format` to run `python -m CheckReleaseNoteFormat`.
  - `setup-pre-commit` to run `pre-commit install`.
- Updated documentation to:
  - run `pixi run -e dev setup-pre-commit` for enabling pre-commit hooks instead of `pre-commit install`.
  - run `pixi run test-base` for running unit tests instead of `make pytest`.
  - add tool description for `pyright`.
  - specify running the tools collectively with `pixi run lint-all` rather than `make check`.
  - specify pixi command for running the tools individually.

### Developer Testing 

- I have verified unit tests pass locally: `python -m pytest -vs`.
- I have verified clean commits passing.
- I have verified committing with formatting issues fails.

- [ ] Unit tests pass locally: `python -m pytest -vs`.
- [ ] Validate clean commits pass as expected.
- [ ] Validate commits with formatting issues fail as expected.
- [ ] Validate the documentation is updated correctly.

There was no task added for `end-of-file-fixer` and `prettier` as they are self-contained hooks. Unlike Python tooling they dont duplicate configuration from `pixi.toml` so moving them to Pixi wouldn't really simplify maintenance.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,3 +56,10 @@ repos:
         types: [python]
         files: ^mantidimaging/
         pass_filenames: false
+
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v3.0.0
+    hooks:
+      - id: prettier
+        files: \.(json|ya?ml)$
+        exclude: ^conda/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,6 @@ repos:
         entry: pixi run -e dev check-licenses
         language: system
         types: [python]
-        files: ^mantidimaging/
 
       - id: check-release-notes-format
         name: Check Release Notes Format
@@ -24,38 +23,37 @@ repos:
       - id: yapf
         name: yapf
         description: Checks that all files pass yapf checks
-        entry: pixi run -e dev yapf
+        entry: pixi run -e lint yapf-staged-base
         language: system
         types: [python]
         files: ^mantidimaging/
-        pass_filenames: false
+        pass_filenames: true
 
       - id: mypy
         name: mypy
         description: Checks that all files pass mypy checks
-        entry: pixi run -e dev mypy
+        entry: pixi run -e dev mypy-staged-base
         language: system
         types: [python]
         files: ^mantidimaging/
-        pass_filenames: false
+        pass_filenames: true
 
       - id: pyright
         name: pyright
         description: Checks that all files pass pyright checks
-        entry: pixi run -e dev pyright
+        entry: pixi run -e dev pyright-staged-base
         language: system
         types: [python]
-        files: ^mantidimaging/
-        pass_filenames: false
+        pass_filenames: true
 
       - id: ruff
         name: ruff
         description: Checks that all files pass ruff checks
-        entry: pixi run -e dev ruff
+        entry: pixi run -e lint ruff-staged-base
         language: system
         types: [python]
         files: ^mantidimaging/
-        pass_filenames: false
+        pass_filenames: true
 
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v3.0.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,49 +4,55 @@ repos:
     hooks:
       - id: end-of-file-fixer
         types: [python]
-  - repo: https://github.com/google/yapf
-    rev: v0.43.0
-    hooks:
-      - id: yapf
   - repo: local
     hooks:
       - id: check-licenses
         name: Check Licenses
         description: Checks that all files have correct copyright licenses
-        entry: python -m CheckLicensesInFiles
-        language: python
+        entry: pixi run -e dev check-licenses
+        language: system
         types: [python]
-  - repo: local
-    hooks:
+        files: ^mantidimaging/
+
       - id: check-release-notes-format
         name: Check Release Notes Format
         description: Validate release note filename and contents format before commit
-        entry: python  -m CheckReleaseNoteFormat
-        language: python
+        entry: pixi run -e dev check-release-notes-format
+        language: system
         files: ^docs/release_notes/next/
-        additional_dependencies: [requests]
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: "v1.17.1"
-    hooks:
-      - id: mypy
+
+      - id: yapf
+        name: yapf
+        description: Checks that all files pass yapf checks
+        entry: pixi run -e dev yapf
+        language: system
+        types: [python]
         files: ^mantidimaging/
-        args: [--ignore-missing-imports]
-        additional_dependencies: [types-docutils, types-PyYAML, types-requests]
-  - repo: local
-    hooks:
+        pass_filenames: false
+
+      - id: mypy
+        name: mypy
+        description: Checks that all files pass mypy checks
+        entry: pixi run -e dev mypy
+        language: system
+        types: [python]
+        files: ^mantidimaging/
+        pass_filenames: false
+
       - id: pyright
         name: pyright
         description: Checks that all files pass pyright checks
-        entry: pyright
+        entry: pixi run -e dev pyright
         language: system
         types: [python]
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.7
-    hooks:
+        files: ^mantidimaging/
+        pass_filenames: false
+
       - id: ruff
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v3.0.0
-    hooks:
-      - id: prettier
-        files: \.(json|ya?ml)$
-        exclude: ^conda/
+        name: ruff
+        description: Checks that all files pass ruff checks
+        entry: pixi run -e dev ruff
+        language: system
+        types: [python]
+        files: ^mantidimaging/
+        pass_filenames: false
--- a/docs/developer_docs/how-to-guides/testing.rst
+++ b/docs/developer_docs/how-to-guides/testing.rst
@@ -63,7 +63,7 @@ Mantid Imaging uses `pre-commit <https://pre-commit.com/>`_ to run pre-commit ho
 
 To enable pre-commit hooks, you must manually install them by running::
 
-    pre-commit install
+    pixi run -e dev setup-pre-commit
 
 Once installed, the hooks will execute automatically with every commit.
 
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -4,46 +4,59 @@ repos:
...
         files: ^docs/release_notes/next/
-        additional_dependencies: [requests]
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: "v1.17.1"
-    hooks:
+
+      - id: yapf
+        name: yapf
+        description: Checks that all files pass yapf checks
+        entry: pixi run -e dev yapf
```
JackEAllen: We can rely on the task invoked environment here to shorten this to: `pixi run yapf`

.pre-commit-config.yaml
```diff
@@ -4,46 +4,59 @@ repos:
...
+        language: system
+        types: [python]
+        files: ^mantidimaging/
+        pass_filenames: false
+
       - id: mypy
+        name: mypy
+        description: Checks that all files pass mypy checks
+        entry: pixi run -e dev mypy
```
JackEAllen:
```suggestion
        entry: pixi run mypy
```

.pre-commit-config.yaml
```diff
@@ -4,46 +4,59 @@ repos:
...
-        additional_dependencies: [types-docutils, types-PyYAML, types-requests]
-  - repo: local
-    hooks:
+        pass_filenames: false
+
       - id: pyright
         name: pyright
         description: Checks that all files pass pyright checks
-        entry: pyright
+        entry: pixi run -e dev pyright
```
JackEAllen:
```suggestion
        entry: pixi run pyright
```

.pre-commit-config.yaml
```diff
@@ -4,46 +4,59 @@ repos:
...
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.7
-    hooks:
+        files: ^mantidimaging/
+        pass_filenames: false
+
       - id: ruff
+        name: ruff
+        description: Checks that all files pass ruff checks
+        entry: pixi run -e dev ruff
```
JackEAllen:
```suggestion
        entry: pixi run ruff
```

.pre-commit-config.yaml
```diff
@@ -4,46 +4,59 @@ repos:
...
         name: Check Licenses
         description: Checks that all files have correct copyright licenses
-        entry: python -m CheckLicensesInFiles
-        language: python
+        entry: pixi run -e dev check-licenses
+        language: system
         types: [python]
-  - repo: local
-    hooks:
+        files: ^mantidimaging/
```
JackEAllen: We may want to continue to leave out files here as this will tighten the scope to just inside mantid imaging and miss out the utility scripts we have etc.

.pre-commit-config.yaml
```diff
@@ -4,46 +4,59 @@ repos:
...
         name: pyright
         description: Checks that all files pass pyright checks
-        entry: pyright
+        entry: pixi run -e dev pyright
         language: system
         types: [python]
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.7
-    hooks:
+        files: ^mantidimaging/
```
JackEAllen: I think we'll want to leave out the files here again for the same reason of checking the utility scripts as well

.pre-commit-config.yaml
```diff
@@ -4,46 +4,59 @@ repos:
...
-    hooks:
+
+      - id: yapf
+        name: yapf
+        description: Checks that all files pass yapf checks
+        entry: pixi run -e dev yapf
+        language: system
+        types: [python]
+        files: ^mantidimaging/
+        pass_filenames: false
```
JackEAllen: To keep this running quickly, i think we can change all `pass_filenames` to `true` so we only check staged files instead of the whole source directly as we only want to block files in the commit. If anything else comes up for the whole source directly, we can catch this in CI.

One thing to note is that for this to work however, you'll also need to update `pixi.toml` with tasks to allow for this as the base tasks being used here will check the source directly either way. so you'll need to do something like:

```toml
[feature.lint.tasks]
yapf-staged-base = { cmd = "python -m yapf --parallel --diff", description = "Check formatting of staged files with yapf (for pre-commit)" }
ruff-staged-base = { cmd = "ruff check", description = "Run ruff on staged files (for pre-commit)" }
 
...

[feature.dev.tasks]
mypy-staged-base = { cmd = "python -m mypy --ignore-missing-imports", description = "Run mypy on staged files (for pre-commit)" }
pyright-staged-base = { cmd = "python -m pyright", description = "Run pyright on staged files (for pre-commit)" }
```

to be called in the pre-commit as `pixi run yapf-staged`, `pixi run mypy-staged`, `pixi run pyright-staged` and `pixi run ruff-staged`

JackEAllen (dismissed):
It's nice to see Pixi finally being used for pre-commit. This will improve our version consistency where instead of relying on pinned versions in the pre-commit, versions will now match whatever is used in our `pixi.toml` file. 

We also get the added benefit of a simplified pre-commit in that we can use familiar Pixi tasks for commands and there's no longer a need to include `additional_dependencies` like we needed to with the previous implementation for `check-release-notes-format` with `additional_dependencies: [requests]` or `additional_dependencies: [types-docutils, types-PyYAML, types-requests]` for MyPy due to no longer needing `language: python` where pre-commit needed to manage virtualenvs that needed those injected into the tasks.

I have a few suggestions, but overall it's looking really good. Thank you for tackling the migration 👍

JackEAllen:
Handing over this review to Ashley to continue on with.

.pre-commit-config.yaml
RasmiaKulan: @JackEAllen The environment definitions in pixi.toml look like:

```
[environments]
lint = ["lint", "runtime", "runtime-cpu"]
dev = ["runtime", "runtime-cpu", "dev", "test", "lint"]
dev-gpu = ["runtime", "runtime-gpu", "gpu-verify", "dev", "test", "lint"]
build-cpu = ["runtime", "runtime-cpu", "build"]
build-gpu = ["runtime", "runtime-gpu", "gpu-verify", "build"]
```
and because lint, dev and dev-gpu include the lint feature, they also inherit `yapf-staged-base`, `ruff-staged-base`, `pyright-staged-base`, `mypy-staged-base`  which then means that all these could be in either of those environments therefore needing to specify which one to use. So in our case it would be `-e lint` for `yapf-staged-base`, `ruff-staged-base`and `-e dev` for  `pyright-staged-base`, `mypy-staged-base`.

ashleymeigh2 (approved):
The suggested chnages have been made. The PR looks good.

PR #3188, opened 2026-07-14
feat(recon): add SIRT non-negativity constraint

### LINKED ISSUE #3144: Reconstruction: Add ability to apply SIRT Non-negativity constraint
### Current Behaviour
Currently there is no option to apply a non-negativity constraint with SIRT:

Image

### Desired Behaviour
It would be great to add a checkbox for application of a non-negativity constraint to SIRT, similar to how there is one for PDHG-TV:

Image

### Importance of Feature
The non-negativity constraint can improve the reconstruction result. This is because it prevents negative values in the solution - as it is impossible to have negative attenuation we know that all the values should be positive. This provides reduction of noise in regions of low attenuation.
An example of an image of a rod with and without the non-negativity (NN) applied. We see a change in what the background region looks like:

Image

### Additional Context
The SIRT currently used is from ASTRA.
The `option.MinConstraint` could be set to 0 to enforce non-negativity
https://astra-toolbox.com/docs/algs/SIRT.html

ashleymeigh2:
### Description

This change adds support for a non-negativity constraint in SIRT reconstructions using ASTRA. When the existing reconstruction “Non-negative” option is enabled for SIRT, the backend now forwards ASTRA’s MinConstraint = 0 setting so reconstructed values are clamped at zero.

The reconstruction window also now exposes the non-negative control for SIRT, matching the behaviour already available for other algorithms that support the option.

- [x]  Confirm the SIRT reconstruction UI shows the non-negative checkbox.
- [x]  Confirm enabling non-negative reconstruction maps to ASTRA MinConstraint = 0.
- [x]  Run the focused tests for the recon window and ASTRA recon slice.

### DIFF
--- a/docs/release_notes/next/feature-3144-add-sirt-non-negativity-constraint
+++ b/docs/release_notes/next/feature-3144-add-sirt-non-negativity-constraint
@@ -1,0 +1,1 @@
+#3144: Add ability to apply a non-negativity constraint to SIRT reconstruction.
--- a/mantidimaging/core/reconstruct/astra_recon.py
+++ b/mantidimaging/core/reconstruct/astra_recon.py
@@ -146,6 +146,8 @@ class AstraRecon(BaseRecon):
             proj_geom = astra.create_proj_geom('parallel_vec', image_width, vectors)
             cfg = astra.astra_dict(recon_params.algorithm)
             cfg['FilterType'] = recon_params.filter_name
+            if recon_params.non_negative:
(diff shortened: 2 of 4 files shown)


RasmiaKulan:
I can verify that the changes work as expected where the NN constraint appears for SIRT and doesnt for FBP and the range shortens when the NN constraint is applied. The test cases cover the feature implemented correctly. Thank you so much for this @ashleymeigh2!

RasmiaKulan (approved):

PR #3192, opened 2026-07-21
[CI/CD]: Migrate all checks from conda/mamba to pixi in GitHub Actions

### LINKED ISSUE #2958: [CI/CD]: Migrate linting checks from conda/mamba to pixi in GitHub Actions (STEP 1)
### Current Behaviour

The `.github/workflows/conda.yml` workflow uses `conda-incubator/setup-miniconda@v3` to set up a conda environment and runs linting checks (yapf, ruff, mypy, pyright) using the conda environment created by `setup.py create_dev_env`.

This requires:
- Manual environment creation steps
- Custom caching configuration for conda packages
- Maintenance of both `setup.py` environment creation and `pixi.toml` configuration

### Desired Behaviour

The linting workflow should use Pixi's native environment management:
- Use `prefix-dev/setup-pixi@v0.9.3` GitHub Action (https://github.com/marketplace/actions/setup-pixi)
- Run linting tasks using the `lint` environment defined in `pixi.toml`
- Commands should be: `pixi run -e lint yapf`, `pixi run -e lint ruff`, `pixi run -e lint mypy`, `pixi run -e lint pyright`
- Leverage Pixi's automatic dependency caching

This keeps the workflow aligned with local development environments and removes the need for `setup.py create_dev_env` in CI.

- Ensure the `lint` environment in `pixi.toml` includes all necessary dependencies (mypy, pyright, yapf, ruff, and their type stubs)
- Verify that Pixi's caching mechanism works correctly in GitHub Actions
- Keep the conda-based workflow steps commented out initially so we can easily revert if issues arise
- Test that linting task exit codes are properly propagated to fail the workflow on errors

### Importance of Feature

This is the first step in migrating from conda/mamba to pixi in CI/CD. Starting with linting checks is low-risk because:
- Linting tasks are fast and independent
- They don't require test data, GUI dependencies, or complex setup
- Failures are easy to debug and won't block development
- Success here validates the pixi migration approach for more complex workflows

This migration will also:
- Reduce CI maintenance burden by using a single source of truth (`pixi.toml`)
- Speed up CI through Pixi's improved caching and lock files
- Align CI environment with developer local environments

### Additional Context

- Pixi documentation: https://pixi.sh/dev/
- The `lint` environment is already defined in `pixi.toml` with all required dependencies
- The existing linting tasks in `pixi.toml` can be used directly: `yapf`, `ruff`, `mypy`, `pyright`
- This is part of a phased migration from conda/mamba to pixi across the entire project

RasmiaKulan:
### Description

1. `conda.yml`:
   - Added `Setup Pixi` environment with `prefix-dev/setup-pixi@v0.9.3` and added `pixi.lock` to the repo so the setup can compute cache keys and restore the Pixi environment in CI.
   - Updated `List versions` to run with Pixi.
   - Updated lint checks including `Yapf, `Ruff`, `mypy` and `pyright` to run with Pixi.
   - Updated `Pytest` to initiate with Pixi using xvfb-run with the full pytest flags directly.This is because`test-base` runs tests without `--ignore=mantidimaging/eyes_tests`.
   - Updated `GUI Tests System` and `GUI Tests Screenshots Applitools` to run with Pixi's `test-system-cpu` and `test-screenshots-cpu`
   - Updated `Coveralls` to with Pixi.
2. Updated `.gitignore` to remove `pixi.lock` as it is required for caching in `Setup Pixi`.
3. Updated `coverage` version range in `pixi.toml`. This was done because the version conda used is `7.4.4` and the one pixi prefered was `7.5.4` resulting in `Pytest` timing out. NOTE: This is done temporarily, I will be creating another issue to update the coverage version for both Pixi and Conda.

### Developer Testing 

- I have verified all these changes result in conda workflow passing in github.

- Validate the following implementations coincide with `pixi.toml`:
- [ ] Pixi environment version
- [ ] Lint checks commands
- [ ] System and Screenshot tests commands

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           # Needed to get tags so that git describe works during package build
           fetch-depth: "0"
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -31,40 +31,49 @@ jobs:
           echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
         shell: bash
 
-      - name: Cache Miniforge and Pip packages
-        uses: actions/cache@v4
-        env:
-          CACHE_NUMBER: 0
-        with:
-          path: |
-            ~/conda_pkgs_dir
-            ~/.cache/pip
-          key: ${{runner.os}}-condapkg-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml')}}
+      # - name: Cache Miniforge and Pip packages
+      #   uses: actions/cache@v4
+      #   env:
+      #     CACHE_NUMBER: 0
+      #   with:
+      #     path: |
+      #       ~/conda_pkgs_dir
+      #       ~/.cache/pip
+      #     key: ${{runner.os}}-condapkg-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml')}}
 
-      - name: Setup Miniforge
-        uses: conda-incubator/setup-miniconda@v4
-        with:
-          miniforge-version: latest
-          activate-environment: mantidimaging-dev
-          auto-activate: false
-          use-mamba: true
-          conda-remove-defaults: true
+      # - name: Setup Miniforge
+      #   uses: conda-incubator/setup-miniconda@v4
+      #   with:
+      #     miniforge-version: latest
+      #     activate-environment: mantidimaging-dev
+      #     auto-activate: false
+      #     use-mamba: true
+      #     conda-remove-defaults: true
 
-      - name: Cache Miniforge environment
-        uses: actions/cache@v4
-        id: cache-miniforge-environment
-        env:
-          CACHE_NUMBER: 0
+      - name: Setup Pixi
+        uses: prefix-dev/setup-pixi@v0.9.3
         with:
-          path: ~/miniconda3/envs/mantidimaging-dev
-          key: ${{runner.os}}-condaenv-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml', 'setup.py')}}
+          pixi-version: v0.59.0
 
-      - name: Mantid Imaging developer dependencies
-        if: steps.cache-miniforge-environment.outputs.cache-hit != 'true'
-        shell: bash -l {0}
-        run: |
-          conda deactivate
-          python3 ./setup.py create_dev_env
+          cache: true
+          auth-host: prefix.dev
+          auth-token: ${{ secrets.PREFIX_DEV_TOKEN }}
+
+      # - name: Cache Miniforge environment
+      #   uses: actions/cache@v4
+      #   id: cache-miniforge-environment
+      #   env:
+      #     CACHE_NUMBER: 0
+      #   with:
+      #     path: ~/miniconda3/envs/mantidimaging-dev
+      #     key: ${{runner.os}}-condaenv-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml', 'setup.py')}}
+
+      # - name: Mantid Imaging developer dependencies
+      #   if: steps.cache-miniforge-environment.outputs.cache-hit != 'true'
+      #   shell: bash -l {0}
+      #   run: |
+      #     conda deactivate
+      #     python3 ./setup.py create_dev_env
 
       - name: Other dependencies
         shell: bash -l {0}
@@ -72,31 +81,31 @@ jobs:
           sudo apt update -y
           sudo apt-get install -y xvfb libxcb* libxkb*
 
-      - name: List versions
-        shell: bash -l {0}
-        run: |
-          mamba env list
-          python --version; mamba list ; pip list
+      # - name: List versions
+      #   shell: bash -l {0}
+      #   run: |
+      #     mamba env list
+      #     python --version; mamba list ; pip list
 
       - name: Yapf
         shell: bash -l {0}
         run: |
-          yapf --parallel --diff --recursive .
+          pixi run -e lint yapf
 
       - name: Ruff
         shell: bash -l {0}
         run: |
-          ruff check .
+          pixi run -e lint ruff
 
       - name: mypy
         shell: bash -l {0}
         run: |
-          mypy --ignore-missing-imports mantidimaging
+          pixi run -e lint mypy
 
       - name: pyright
         shell: bash -l {0}
         run: |
-          pyright mantidimaging
+          pixi run -e lint pyright
 
       - name: pytest
         timeout-minutes: 10
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -54,6 +54,9 @@ jobs:
         uses: prefix-dev/setup-pixi@v0.9.3
         with:
           pixi-version: v0.59.0
+          environments: >-
+            dev
+            lint
           cache: true
           cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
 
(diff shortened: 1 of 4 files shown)


.github/workflows/conda.yml
```diff
@@ -75,34 +82,34 @@ jobs:
       - name: List versions
         shell: bash -l {0}
         run: |
-          mamba env list
-          python --version; mamba list ; pip list
+          pixi run -e dev python --version
+          pixi run -e dev python -m pip list
```
JackEAllen: I think we can change this to look at all Pixi packages grabbed from Conda and PyPi using rather than grabbing only the PyPi ones which will give us an incomplete picture of our packages:

```suggestion
          pixi list -e dev
```

.github/workflows/conda.yml
```diff
@@ -75,34 +82,34 @@ jobs:
...
         shell: bash -l {0}
         run: |
-          yapf --parallel --diff --recursive .
+          pixi run -e lint yapf
 
       - name: Ruff
         shell: bash -l {0}
         run: |
-          ruff check .
+          pixi run -e lint ruff
```
JackEAllen, line 59: I think for Yapf and Ruff steps, we can possibly take advantage of the base tasks to set the lint environment so we may be able to simplify to `pixi run yapf` and `pixi run ruff`

.github/workflows/conda.yml
```diff
@@ -75,34 +82,34 @@ jobs:
...
         shell: bash -l {0}
         run: |
-          mypy --ignore-missing-imports mantidimaging
+          pixi run -e lint mypy
 
       - name: pyright
         shell: bash -l {0}
         run: |
-          pyright mantidimaging
+          pixi run -e lint pyright
```
JackEAllen: For MyPy and Pyright, I'm not sure we want to use the lint environment for these as I think they'll need  access to the full runtime packages to type-check correctly. Do they work if we rely on the base tasks to select the correct environment by changing to `pixi run mypy` and `pixi run pyright`?

.github/workflows/conda.yml
```diff
@@ -135,7 +138,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          coveralls --service=github
+          pixi run -e dev coveralls --service=github
```
JackEAllen: I wonder if we still need `--service=github`?
The environment variable `GITHUB_ACTION` is [set by default for all GitHub actions](https://docs.github.com/en/actions/reference/workflows-and-actions/variables) to `true` which I think coveralls own docs also mention [HERE](https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support) so I'd be tempted to try and just remove `--service=github` and see what happens. 

It would also seem with newer version of coveralls, `--service` has been deprecated in favour of `--service-name`  so removing also saves us having issues with this for newer versions of coveralls.

.github/workflows/conda.yml
```diff
@@ -31,40 +31,47 @@ jobs:
...
+      # - name: Setup Miniforge
+      #   uses: conda-incubator/setup-miniconda@v4
+      #   with:
+      #     miniforge-version: latest
+      #     activate-environment: mantidimaging-dev
+      #     auto-activate: false
+      #     use-mamba: true
+      #     conda-remove-defaults: true
+
+      - name: Setup Pixi
```
JackEAllen, line 28: Out of interest, in the pixi setup task, have you tried specifying the environments needed in the workflow to cache them so we don't need to install them lazily on first use? So something like:
```yml
with:
    pixi-version: v0.59.0
    environmnets: >-
        dev
        lint
    cache: true
    cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
```

.github/workflows/conda.yml
```diff
@@ -115,18 +122,14 @@ jobs:
       - name: GUI Tests System
         shell: bash -l {0}
         run: |
-          xvfb-run --auto-servernum /bin/time -v python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true mantidimaging/gui/test/ --run-system-tests --durations=10
+          pixi run -e dev test-system-cpu
```
JackEAllen: I think we may be able to simplify this to `pixi run test-system-cpu` as the task itself defines the environment `dev`

.github/workflows/conda.yml
```diff
@@ -115,18 +122,14 @@ jobs:
...
       - name: GUI Tests Screenshots Applitools
         if: github.event_name == 'pull_request'
         shell: bash -l {0}
-        env:
-          APPLITOOLS_API_KEY: ${{ secrets.APPLITOOLS_API_KEY }}
-          APPLITOOLS_BATCH_ID: ${{ github.sha }}
-          GITHUB_BRANCH_NAME: ${{ github.head_ref }}
         run: |
-          xvfb-run --auto-servernum /bin/time -v python -m pytest -vs -rs -p no:xdist -p no:randomly -p no:repeat -p no:cov -o log_cli=true mantidimaging/eyes_tests --run-eyes-tests --durations=10
+          pixi run -e dev test-screenshots-cpu
```
JackEAllen: I think we may be able to simplify this to `pixi run test-screenshots-cpu`.

JackEAllen (changes requested):

JackEAllen:
The GUI Tests Screenshots Applitools don't quite look right. I think you need to add the environment variabled back within `env:` to get the GitHub secrets. 

I think you may also need to rework `pixi.toml` so we can have a seperate task for running this locally and one for CI as the current task sets `APPLITOOLS_API_KEY = "local"` which we wont want for CI as we want to use the GitHub secrets API key.

JackEAllen:
With the coveralls syntax warnings we are experiencing, i suspect this may be solved by updating to `"4.1.*"` where the newer versions replace [`docopt`](https://github.com/TheKevJames/coveralls-python/blob/4.0.0/coveralls/cli.py) with [`typer`](https://github.com/TheKevJames/coveralls-python/blob/4.1.0/coveralls/cli.py) in [4.1.0](https://github.com/TheKevJames/coveralls-python/blob/master/CHANGELOG.md) which i think has fixes for the incompatibility issues with Python 3.12 where it uses [unescapable backslashes in regex strings that 3.12 warns about](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes).

JackEAllen:
**Note**: The following warning raised in `GUI Tests Screenshots Applitools` step will be handled by a separate Issue and subsequently a different PR as it is outside the scope of work encapsulated within this PR:
```bash
mantidimaging/eyes_tests/compare_images_window_test.py::CompareImagesWindowTest::test_compare_images_window_opens (node:14937) ExperimentalWarning: Module._stat is an experimental feature and might change at any time
(Use `core --trace-warnings ...` to show where the warning was created)
(node:14937) ExperimentalWarning: Module._readPackage is an experimental feature and might change at any time
You are using "5.23.12" and there is a newer version "6.4.22" of the "eyes-images" SDK.
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
PASSED--- Test passed. 
```

.github/workflows/conda.yml
```diff
@@ -31,40 +31,15 @@ jobs:
           echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
         shell: bash
```
JackEAllen: I believe the "Get Date" step may no longer be needed now we are using Pixi as i think it was previously used when building the Conda cache key. Now we are get the cache key from the hash of the `pixi.lock`

JackEAllen (commented):
Overall, I think this is a good first step to migrating our workflows over to Pixi. There will still be a few thing's to do after this is merged in and we have also updated the version of Pixi we are using. Because of this, there is no strong argument to consider creating new tasks in our `pixi.toml` as part of this PR for steps such as coveralls and pytest used in the `conda.yml` workflow as we can reconsider this once we are using a newer version of Pixi (`>=0.63.0`) that allows for `default-environment` which will simplify tasks further anyway 🙂 

The only thing left here is just to remove the get data task and condense and/or add a little more clarity to some of the commits to add context to the changes and we are good to go!

JackEAllen (dismissed):

JackEAllen:
Handing over the final approval review to @ashleymeigh2

ashleymeigh2 (approved):
Looks good. Suggested change was made

PR #3194, opened 2026-07-27
Default image export stack selection to the currently selected stack …

### LINKED ISSUE #3180: Image Save Dialog: Auto-select Stack Selected in Main Window TreeDatasetWidget
### Current Behaviour

IMAT have voiced that they often find themselves accidentally saving a different stack to the one they intended to export and close their instance of Mantid Imaging before they have realised, resulting in the stack that was intended to be exported being unrecoverable.

When a stack is selected within the main window, users intuitively assume that the selected stack will be the default selection when exporting a stack which is not currently the case.

* Refactor the tree dataset widget and save dialogs to set the selected stack from the main window (when a stack is selected) as the default selection when exporting the image stack in the dialog window.

### Desired Behaviour

If a stack is selected in the main window, this should be the default selection in the save dialog when a user wants to export their data

When no selection is made of a stack in the main window, the stack selection should match current behaviour. 

### Importance of Feature

* Improve user friendliness of Mantid Imaging
* Reduces the risk of closing Mantid Imaging before exporting modified data.

ashleymeigh2:
…in the main window tree when available

### Description

This change improves export safety and usability by defaulting the Save Images dialog to the stack currently selected in the main window tree, when that selection is a valid stack.

- [ ]  Unit tests pass locally: python -m pytest -vs
- [ ]  Selecting a stack in the tree makes it the default in Save Images (including Recon/Spectrum workflows)
- [ ]  With no valid stack selected, fallback default remains Recon/Tomo preference

### DIFF
--- a/mantidimaging/gui/windows/main/presenter.py
+++ b/mantidimaging/gui/windows/main/presenter.py
@@ -545,11 +545,11 @@ class MainWindowPresenter(BasePresenter):
 
     def _select_tree_widget_item(self, tree_widget_item: QTreeWidgetItem) -> None:
         """
-        Clears the existing selection on the dataset tree view and selects a given item.
+        Clears the existing selection on the dataset tree view and sets a given current item.
         :param tree_widget_item: The item to select.
         """
         self.view.dataset_tree_widget.clearSelection()
-        tree_widget_item.setSelected(True)
+        self.view.dataset_tree_widget.setCurrentItem(tree_widget_item)
 
     @staticmethod
     def _remove_recon_item_from_tree_view(recon_group, uuid_remove: uuid.UUID) -> bool:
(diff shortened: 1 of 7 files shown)


samtygier-stfc (changes requested):
It would be good if this also work when you select a stack by clicking its tab. Clicking a tab does highlight the stack in the tree view, but maybe something extra is needed so that it "selects" it.

mantidimaging/gui/windows/main/presenter.py
```diff
@@ -549,6 +549,7 @@ def _select_tree_widget_item(self, tree_widget_item: QTreeWidgetItem) -> None:
         :param tree_widget_item: The item to select.
         """
         self.view.dataset_tree_widget.clearSelection()
+        self.view.dataset_tree_widget.setCurrentItem(tree_widget_item)
         tree_widget_item.setSelected(True)
```
samtygier-stfc: I think that now that it has `self.view.dataset_tree_widget.setCurrentItem(tree_widget_item)` it does not need `tree_widget_item.setSelected(True)`

samtygier-stfc (commented):
Working well. I think the old `setSelected()` line can be removed now (and maybe some tests updated from that).

samtygier-stfc (approved):
Working well for me.

PR #3200, opened 2026-07-29
Resolve Version Schema for built packages

JackEAllen:
Resolve schema for build version from V.V.V<N> to V.V.V<.post<N>> to match the version schema used previously in conda/meta.yaml

### Description

The [version numbers in Anaconda](https://anaconda.org/channels/mantidimaging/packages/mantidimaging/files) don't quite match what we used to have with `conda/meta.yml` prior to the rattler-build migration. Our build version seems to be `3.2.0a172` (i.e. V.V.Va1<N>` when it should be `V.V.Va1<.postN>` if we want to continue to match the version schema we have used previously to avoid possible confusion. where `.postN` will be added when the version tag is non-zero.

This PR modified the `recipe.yaml` to update the version with an if statement to check the tag version and dynamically format with `.postN`

### Developer Testing 

- Check Anaconda to validate version schema is correctly set when publish package is ran (i'v

- [ ] See comment below: https://github.com/mantidproject/mantidimaging/pull/3200#issuecomment-5132466411 and validate package visible on Anaconda

### DIFF
--- a/conda/recipe.yaml
+++ b/conda/recipe.yaml
@@ -1,9 +1,9 @@
 schema_version: 1
 
 context:
-  version_tag: ${{ git.latest_tag(".") | default("0.0.0")}} # ratter-build --experimental
-  commit_number: ${{ git.latest_tag_distance(".")  | default("0")}} # ratter-build>=0.65 --experimental
-  version: ${{ version_tag + commit_number }}
+  version_tag: ${{ git.latest_tag(".") | default("0.0.0")}} # rattler-build --experimental
+  commit_number: ${{ git.latest_tag_distance(".") | default("0")}} # ratter-build>=0.65 --experimental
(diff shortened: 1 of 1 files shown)


JackEAllen:
See confirmation of correct schema being applied to Linux workflow build versions.
image

https://anaconda.org/channels/mantidimaging/packages/mantidimaging/files

JackEAllen:
**Note**: we may need to remove all previous builds in Anaconda with the incorrect schema (`V.V.Va1<N>`) depending on how things are sorted. This may also result in it taking a couple of data for Ada to start grabbing the correct version again with the `V.V.Va1<.post<N>` schema again.

RasmiaKulan:
I can confirm that I see the correct schema in https://anaconda.org/channels/mantidimaging/packages/mantidimaging/files Thank you for correcting this! 😊

RasmiaKulan (approved):

Issue #2962, opened 2025-12-02
[CI/CD]: Migrate documentation building from conda/mamba to pixi (STEP 5)


JackEAllen:
### Current Behaviour

The `.github/workflows/docs.yml` workflow builds Sphinx documentation using:
- Conda environment created by `setup-miniconda@v3`
- Manual environment creation via `setup.py create_dev_env`
- Sphinx build process that requires the full application dependencies
- Documentation deployment to GitHub Pages

The workflow currently uses conda for dependency management.

### Desired Behaviour

Documentation building should use pixi:
- Use `prefix-dev/setup-pixi@v0.9.3` with `environments: dev`
- Build docs using: `pixi run -e dev build-docs`
- Leverage the existing `build-docs` task defined in `pixi.toml`
- Maintain the same documentation output format and location
- Keep GitHub Pages deployment steps unchanged

### Potential Pitfalls or Risks to Avoid

- Verify that Sphinx and all extensions are in the `dev` environment (sphinx, pydata-sphinx-theme, sphinx-copybutton, sphinx-design)
- Ensure that the application imports work during doc building (autodoc requires importable modules)
- Confirm that static assets (images, CSS) are collected correctly
- Test that cross-references and API documentation generate properly
- Verify that the built HTML output is in the expected location for deployment
- Ensure documentation warnings are still visible in the workflow logs

### Importance of Feature

Documentation is critical for users and developers:
- User-facing documentation helps users understand the application
- API documentation helps developers contribute to the project
- Release notes are generated from the documentation system
- Documentation is deployed to https://mantidproject.github.io/mantidimaging/

Benefits of pixi migration:
- Faster documentation builds through better caching
- Consistent documentation environment between local builds and CI
- Reduced risk of documentation build failures due to environment mismatches
- Alignment with other CI workflows

### Additional Context

- The `build-docs` task is defined in `pixi.toml`
- Sphinx configuration is in `docs/conf.py`
- Documentation dependencies are in the `dev` feature of `pixi.toml`: sphinx, pydata-sphinx-theme, sphinx-copybutton, sphinx-design
- Documentation is built from the `docs/` directory
- The workflow triggers on pushes to `main` branch and creates GitHub Pages deployment
- This migration can be done independently of other workflow migrations
- Documentation build is less critical than tests, so this is a good candidate for later migration

PR #3227, opened 2026-08-04
[CI/CD]: Migrate Windows testing workflow from conda/mamba to pixi

### LINKED ISSUE #2961: [CI/CD]: Migrate Windows testing workflow from conda/mamba to pixi (STEP 4)
### Current Behaviour

The `.github/workflows/windows.yml` workflow uses conda to:
- Set up a Windows development environment via `setup-miniconda@v3`
- Run pytest unit tests
- Run system tests with Windows-specific display configuration (1920x1080 resolution)
- Handle Windows-specific path and environment variable quirks

The workflow uses `setup.py create_dev_env` and manual conda activation.

### Desired Behaviour

The Windows workflow should use pixi's native Windows support:
- Use `prefix-dev/setup-pixi@v0.9.3` GitHub Action
- Run tests using: `pixi run -e dev test-cpu` and `pixi run -e dev test-system-cpu`
- Maintain Windows-specific display configuration steps
- Use pixi's automatic caching for faster Windows builds
- Keep the same test output and artifact collection

- Verify pixi works correctly on Windows GitHub Actions runners (windows-latest)
- Ensure Windows path separators are handled correctly by pixi tasks
- Test that PyQt5 and VTK Windows binaries work in pixi environment
- Confirm that Windows display resolution configuration still works with pixi processes
- Watch for Windows-specific environment variable issues (PATH, TEMP, etc.)
- Verify that pytest output capturing works correctly on Windows with pixi
- Ensure test artifacts (coverage reports, screenshots) are created in expected locations

### Importance of Feature

Windows support is critical because:
- Many users run Mantid Imaging on Windows such as IMAT
- Windows has different binary dependencies and behaviour than Linux
- Windows-specific bugs are common in GUI applications
- This validates that pixi can fully replace conda on all supported platforms

Benefits:
- Consistent environment between Windows developers and CI
- Faster Windows CI through improved caching
- Reduced maintenance of platform-specific conda configurations
- Alignment with Linux workflows (same pixi commands work on both platforms)

### Additional Context

- The `dev` environment in `pixi.toml` uses `platforms = ["linux-64", "win-64"]`
- Windows runners use `windows-latest` (currently Windows Server 2022)
- Display resolution is set to 1920x1080 before tests run
- May need to add Windows-specific environment variables in pixi tasks if needed
- Test locally on Windows before submitting if possible

RasmiaKulan:
### Description
1. `windows.yml`:
   - Added `Setup Pixi` environment with `prefix-dev/setup-pixi@v0.9.3`.
   - Updated `List versions` to run with Pixi.
   - Updated lint checks including `Yapf`,`Ruff`,`mypy` and `pyright` to run with Pixi.
   - Updated `Pytest` to initiate with Pixi.
   - Updated `GUI Tests System` and `GUI Tests Screenshots Applitools` to run with Pixi's `test-system-cpu` and `test-screenshots-cpu`
2. `version_check_test.py` : Updated `test_conda_update_message` to mock the Conda executable lookup. The test previously relied on conda or mamba being installed and available on PATH, which caused a `FileNotFoundError` in the Pixi environment. Mocking `find_conda_executable()` to return `"conda"` isolates the test from the environment and allows it to focus on verifying the generated update message.
**This was done as a workaround to avoid the error. Once #2962 has been merged, the checks can then be changed accordingly.**
3. `pixi.toml`: Updated test-system-base to specify screenshot directory.

**NOTE**: This PR mimics changes made to `pixi.toml` in #3237 to make sure the windows migration is compatible with the versions implemented as part of #3237. Once that PR is merged, the related commit ([Update pixi to match updated versions](https://github.com/mantidproject/mantidimaging/pull/3227/commits/e60e975ff2bff7c4985ff884f53611349a1dd84a))  can be reverted in this PR and the branch can be rebased.

### Developer Testing 

- I have verified all these changes result in windows workflow passing in github.

- Validate the following implementations coincide with `pixi.toml`:
- [ ] Pixi environment version
- [ ] Lint checks commands
- [ ] System and Screenshot tests commands

- [ ] ...

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -22,80 +22,90 @@ jobs:
       CONDA_OVERRIDE_CUDA: 11.8
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           # Needed to get tags so that git describe works during package build
           fetch-depth: "0"
 
-      - name: Get Date
-        id: get-date
-        run: |
-          echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
-        shell: bash
+      # - name: Get Date
+      #   id: get-date
+      #   run: |
+      #     echo "date=$(/bin/date -u "+%Y%V")" >> $GITHUB_OUTPUT
+      #   shell: bash
 
-      - name: Cache Miniforge and Pip packages
-        uses: actions/cache@v5
-        env:
-          CACHE_NUMBER: 0
+      # - name: Cache Miniforge and Pip packages
+      #   uses: actions/cache@v5
+      #   env:
+      #     CACHE_NUMBER: 0
+      #   with:
+      #     path: |
+      #       ~/conda_pkgs_dir
+      #       ~/.cache/pip
+      #     key: ${{runner.os}}-condapkg-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml')}}
+
+      # - name: Setup Miniforge
+      #   uses: conda-incubator/setup-miniconda@v4
+      #   with:
+      #     miniforge-version: latest
+      #     activate-environment: mantidimaging-dev
+      #     auto-activate: false
+      #     use-mamba: true
+      #     conda-remove-defaults: true
+
+      # - name: Cache Miniforge environment
+      #   uses: actions/cache@v5
+      #   id: cache-miniforge-environment
+      #   env:
+      #     CACHE_NUMBER: 0
+      #   with:
+      #     path: ${{env.CONDA}}/envs/mantidimaging-dev
+      #     key: ${{runner.os}}-condaenv-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml')}}
+
+      - name: Setup Pixi
+        uses: prefix-dev/setup-pixi@v0.9.3
         with:
-          path: |
-            ~/conda_pkgs_dir
-            ~/.cache/pip
-          key: ${{runner.os}}-condapkg-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml')}}
+          pixi-version: v0.59.0
+          environments: >-
+            dev
+            lint
+          cache: false #enable this after linux testing workflow migration to pixi is merged
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
 
-      - name: Setup Miniforge
-        uses: conda-incubator/setup-miniconda@v4
-        with:
-          miniforge-version: latest
-          activate-environment: mantidimaging-dev
-          auto-activate: false
-          use-mamba: true
-          conda-remove-defaults: true
+      # - name: Install setuptools for system Python
+      #   run: python -m pip install --upgrade pip setuptools
 
-      - name: Cache Miniforge environment
-        uses: actions/cache@v5
-        id: cache-miniforge-environment
-        env:
-          CACHE_NUMBER: 0
-        with:
-          path: ${{env.CONDA}}/envs/mantidimaging-dev
-          key: ${{runner.os}}-condaenv-${{env.CACHE_NUMBER}}-${{steps.get-date.outputs.date}}-${{hashFiles('environment-dev.yml','conda/meta.yaml')}}
-
-      - name: Install setuptools for system Python
-        run: python -m pip install --upgrade pip setuptools
-
-      - name: Mantid Imaging developer dependencies
-        if: steps.cache-miniforge-environment.outputs.cache-hit != 'true'
-        shell: bash -l {0}
-        run: |
-          conda deactivate
-          python ./setup.py create_dev_env
+      # - name: Mantid Imaging developer dependencies
+      #   if: steps.cache-miniforge-environment.outputs.cache-hit != 'true'
+      #   shell: bash -l {0}
+      #   run: |
+      #     conda deactivate
+      #     python ./setup.py create_dev_env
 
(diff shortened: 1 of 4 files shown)


pixi.toml
```diff
@@ -225,7 +226,7 @@ description = "Run system tests (Windows)"
 cmd = [
     "cmd",
     "/c",
-    "mkdir screenshot_tests_output 2>nul & set APPLITOOLS_API_KEY=local & set APPLITOOLS_IMAGE_DIR=%CD%\\screenshot_tests_output & pytest -p no:xdist -p no:randomly -p no:cov mantidimaging/eyes_tests/ -vs --run-eyes-tests",
+    "mkdir screenshot_tests_output 2>nul & set APPLITOOLS_API_KEY=local & set APPLITOOLS_IMAGE_DIR=screenshot_tests_output pytest -p no:xdist -p no:randomly -p no:cov mantidimaging/eyes_tests/ -vs --run-eyes-tests",
```
JackEAllen, line 229: I think we may be missing an `&` between `screenshot_tests_output` and `pytests` here meaning everything after pytest is treated as a value of `APPLITOOLS_IMAGE_DIR` and never executed. The linux task has the correct pattern which should help as a reference.

.github/workflows/windows.yml
```diff
@@ -22,86 +22,52 @@ jobs:
       CONDA_OVERRIDE_CUDA: 11.8
```
JackEAllen, line 22: I think we can remove `CONDA_OVERRIDE_CUDA: 11.8` with the switch to Pixi

pixi.toml
JackEAllen, line 149: I'd test locally on windows, but we "may" be able to remove `CONDA_DLL_SEARCH_MODIFICATION_ENABLE = "1"`

.github/workflows/windows.yml
```diff
@@ -22,86 +22,52 @@ jobs:
...
-          python ./setup.py create_dev_env
+          pixi-version: v0.59.0
+          environments: >-
+            dev
+            lint
+          cache: true
+          cache-write: ${{ github.event_name == 'push' && github.ref_name == 'main' }}
 
       - name: List versions
         shell: bash -l {0}
```
JackEAllen, line 41: I think we can possibly remove all the `shell: bash -l {0}` now as pixi should handle the activation unlike conda which required it to source it's `~/.bash_profile` activation.

JackEAllen (changes requested):
Generally looks good! 
I've noticed a few additional things which we may be able to remove now we are using Pixi which i'd recommend having a play with. 
I can see the screenshot tests aren't running and I think it's a similar issue to what we previously had with the equivalent linux task where we are just missing and `&` before running pytest.

pixi.toml
```diff
@@ -227,7 +226,13 @@ description = "Run system tests (Windows)"
...
-    "mkdir screenshot_tests_output 2>nul & set APPLITOOLS_API_KEY=local & set APPLITOOLS_IMAGE_DIR=%CD%\\screenshot_tests_output & pytest -p no:xdist -p no:randomly -p no:cov mantidimaging/eyes_tests/ -vs --run-eyes-tests",
+    """
+    mkdir screenshot_tests_output 2>nul
+    echo Screenshots directory: screenshot_tests_output
+    set APPLITOOLS_API_KEY=local
+    set APPLITOOLS_IMAGE_DIR=screenshot_tests_output
+    pytest -p no:xdist -p no:randomly -p no:cov mantidimaging/eyes_tests/ -vs --run-eyes-tests
+    """,
 ]
 description = "Run screenshot tests (Windows)"
```
JackEAllen, line 228: If we want to make this a multiline task to help with readability, I think we will need  to use `bash -c` instead of `cmd /c` for this task as `cmd` may be struggling here with the newlines acting as separators which is why the screenshot workflow isn't working. We could also consider moving environment variables to a table like we do in the windows `test-system-base` task
```toml
[target.win-64.tasks.test-screenshots-base]
cmd = [
    "bash", "-c",
    """
    mkdir screenshot_tests_output
    echo "Screenshots directory: screenshot_tests_output"
    pytest -p no:xdist -p no:randomly -p no:cov mantidimaging/eyes_tests/ -vs --run-eyes-tests
    """,
]
env = { APPLITOOLS_API_KEY = "local", APPLITOOLS_IMAGE_DIR = "screenshot_tests_output" }
description = "Run screenshot tests (Windows)"
```

One thing we should be aware of with this change is that we would be adding a dependency on Git Bash on the `PATH` which is true for GitHubs hosted runners, but not necessarily true for all Windows instances given windows doesn't ship with bash. It would mean anyone wanting to use the task natively on windows would require git for windows, WSL or similar to have a bash installation.

Alternatively we could use powershell, but I'm not very familiar with powershell, or we could continue with `cmd`, but we'll need to revert to be one line again which does affect readability, not to by too much in my opinion if we still condense by using an env table in the same way we could with bash:

```toml
[target.win-64.tasks.test-screenshots-base]
cmd = ["cmd", "/c", "mkdir screenshot_tests_output 2>nul & pytest -p no:xdist -p no:randomly -p no:cov mantidimaging/eyes_tests/ -vs --run-eyes-tests"]
env = { APPLITOOLS_API_KEY = "local", APPLITOOLS_IMAGE_DIR = "screenshot_tests_output" }
description = "Run screenshot tests (Windows)"
```

pixi.toml
JackEAllen: Did Windows tasks still work locally after removing this?

JackEAllen (changes requested):
Thanks for the additional changes. The [screenshot tests for windows](https://github.com/mantidproject/mantidimaging/actions/runs/35632930087/job/106443252952?pr=3227) don't seem to be running quite right. I believe this is to do with how cmd doesn't handle new lines as separators.

JackEAllen (approved):
Changes look good and the each step in the windows workflow is working as expected when compared to a windows workflow using Conda so happy to merge in 👍

JackEAllen:
Thanks for updating some of our older release notes in this PR to be dev changes instead of features.

PR #3237, opened 2026-08-19
Pixi Config: Fix version discrapencies in Pixi

### LINKED ISSUE #3228: Pixi Config: Fix version discrepancies in Pixi
### Current Behaviour
While working on [#3192](https://github.com/mantidproject/mantidimaging/pull/3192), there were a couple of issues found within `pixi.toml` related to version differences. These include:

1. Looking at the CI pipeline for [#3192](https://github.com/mantidproject/mantidimaging/pull/3192) under the "List Versions" step, we noticed a discrepancy with the Pixi file itself. iIf we take a look at the version of ccpi-regulariser, the dev build which should be cpu only is using `cuda13_h519cf14_200` but it should be using `cpu_py312hdfdb54b_3`
This requires updating pixi.toml to something like:

   - ccpi-regulariser = { version = "25.0.*", build = "cuda*" } within [feature.runtime-gpu.dependencies]
   - ccpi-regulariser = { version = "25.0.*", build = "cpu*" } within [feature.runtime-cpu.dependencies]
   - Removing ccpi-regulariser = "25.0.*" from [feature.runtime.dependencies]

When adding `ccpi-regulariser = { version = "25.0.*", build = "cuda*" }`  and  `ccpi-regulariser = { version = "25.0.*", build = "cpu*" }`, it requires updating `astra-toolbox` and `libastra` versions . Also `cudatoolkit` definition has been replaces with `cuda-version` and that requires updating as well.

2. #3192 required shortening the range of coverage from `coverage = ">=7.0,<7.6"` to `coverage = ">=7.0,<7.5"` because the version conda used is `7.4.4` and the one pixi prefers is `7.5.4` and using `7.5.4` was resulting in Pytests timing out. The latest version of coverage is `7.15.2` and therefore to avoid downgrading the version in Pixi, we could explore upgrading the version in Conda and Pixi if we need to change the version anyway.

3. Investigate why we get the following warnings when running `GUI Tests Screenshots Applitools`. An inital investigation showed that removing the ci environment from #3192 resulted in them appearing.
```
mantidimaging/eyes_tests/compare_images_window_test.py::CompareImagesWindowTest::test_compare_images_window_opens (node:14937) ExperimentalWarning: Module._stat is an experimental feature and might change at any time
(Use `core --trace-warnings ...` to show where the warning was created)
(node:14937) ExperimentalWarning: Module._readPackage is an experimental feature and might change at any time
You are using "5.23.12" and there is a newer version "6.4.22" of the "eyes-images" SDK.
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
This plugin does not support propagateSizeHints()
PASSED--- Test passed. 
```

### Desired Behaviour
1. Avoid having discrepencies in pixi setup
2. Have updated coverage version that avoid Pytest timing out
3. Avoid warnings when running `GUI Tests Screenshots Applitools`

RasmiaKulan:
### Description

Updated `pixi.toml` with:

- ~~Updated `coverage` version of `">=7.14,<7.16"`.~~ Unable to find a coverage version that avoids Pytest timing out requiring further investigation.
- Replaced `cudatoolkit` with `cuda-version` and updated the version as required by the `ccpi-regulariser`
- Added `ccpi-regulariser = { version = "25.0.*", build = "cuda*" }` to all gpu OS build dependencies and updated `astra-toolbox` and `libastra` to acceptable versions accordingly.
- Added `ccpi-regulariser = { version = "25.0.*", build = "cpu*" }` to all cpu OS build dependencies and  `astra-toolbox` and `libastra`  to acceptable versions accordingly.
- Updated `[feature.gpu-verify.system-requirements]` and `[feature.gpu-verify.system-requirements]` to the updated cuda version.
- Removed `QT_QPA_PLATFORM=offscreen` so that Qt can automatically use the virtual display provided by `xvfb-run`, avoiding the `propagateSizeHints()` warning produced by the offscreen plugin.

### Developer Testing 

- I have verified the compatibility of the dependencies with one another by running `pixi install`, `pixi install -e dev` and `pixi install -e dev-gpu`.

- [ ] ...


RasmiaKulan:
Pytest timing out update: The CI timeout is isolated to the Coverage.py upgrade. `Coverage 7.4.4` passes with the existing test configuration, while `Coverage 7.5.4` and `7.15.4` cause the test suite to time out. Changing pytest-cov and pytest-xdist versions did not resolve the issue, and the timeout also occurs with the original `pytest/pytest-cov/pytest-xdist` versions when only Coverage is upgraded.

JackEAllen (approved):
Changes look good to me. Happy to merge in. We can work on updating conda to match later on.

PR #3241, opened 2026-09-14
CI: Add ruff print check to docker-ubuntu workflow

### LINKED ISSUE #2826: CI: Enable ruff print checks
### Current Behaviour

We occasionally have print statements left over from developer debugging make it into PRs. There is a ruff rule (https://docs.astral.sh/ruff/rules/#flake8-print-t20) that could prevent this.

### Desired Behaviour

Remove some unnecessary print statements and turn others into logs.

There might be a few that really should be print, these can have an ignore rule

Add a step into the docker-ubuntu workflow to run ruff with the print rule
`ruff check --select T20 .`
So that these get caught before merging

This should not be in pre-commit because it is sometimes useful to have them in the code while working on a branch. We just want them removed from the final PR.

Do we want them removed from all the individual commits? Should we encourage people to remove them with an interactive rebase?

There are some prints in scripts and setup.py. Should we allow these?

### Importance of Feature

Nice to have.

RasmiaKulan:
### Developer Testing 

- ...

- [ ] ...


RasmiaKulan:
The print check and updating the prints to logs where appropriate will be added as part of [Migrating Docker workflows from conda/mamba to pixi](#3243)

